PR #11, opened 2021-10-15
Remove welcome notification after installing/activating subscriptions

### LINKED ISSUE #8: Move Welcome message after installing/activating from Core to Subscriptions pro
Remove this message from core and port it to Subscriptions pro

135557328-536501a7-73b2-4b1c-8d51-2ebaed860d3b

This is displayed when WCPay is active and WCPay Subscriptions is enabled (via wp option `_wcpay_feature_subscriptions`).

Should be worker alongside with https://github.com/Automattic/woocommerce-payments/issues/3024  and https://github.com/woocommerce/woocommerce-subscriptions/issues/4226

### To Reproduce
1. Install WC Pay, get subscriptions active. 
2. Activate WC Pay subscriptions using WC Pay DEV
3. See admin notice in plugins screen and possibly elsewhere in admin.

### Expected behavior
Each plugin extending Subscriptions core should implement it's own way of showing a welcome message. The core should not have a default message.

The existing message should be ported to Subscriptions pro through [#4226](https://github.com/woocommerce/woocommerce-subscriptions/issues/4226)

leonardola:
Fixes #8

Completely remove the message that shows up right after installing/activating the subscription plugin 

#### Testing instructions
- Install a fresh Wordpress install with WC Payments plugin
- Install subscriptions-core
- Activate subscriptions using WC Pay Dev tools
- This message should **not** show

135557328-536501a7-73b2-4b1c-8d51-2ebaed860d3b

After merging this we will need to bump the Subscriptions core version at https://github.com/Automattic/woocommerce-payments/issues/3024

### DIFF
--- a/includes/class-wc-subscriptions-core-plugin.php
+++ b/includes/class-wc-subscriptions-core-plugin.php
@@ -287,6 +287,15 @@ class WC_Subscriptions_Core_Plugin {
 	}
 
 	/**
+	 * Gets the activation transient name.
+	 *
+	 * @return string The transient name used to record when the plugin was activated.
+	 */
+	public function get_activation_transient() {
+		return 'woocommerce_subscriptions_activated';
+	}
+
+	/**
 	 * Gets the core Payment Gateways handler class
 	 *
 	 * @since 4.0.0
@@ -468,6 +477,8 @@ class WC_Subscriptions_Core_Plugin {
 			}
 
 			update_option( WC_Subscriptions_Admin::$option_prefix . '_is_active', true );
+
+			set_transient( $this->get_activation_transient(), true, 60 * 60 );
 
 			flush_rewrite_rules();
 
--- a/includes/class-wc-subscriptions-core-plugin.php
+++ b/includes/class-wc-subscriptions-core-plugin.php
@@ -17,7 +17,7 @@ class WC_Subscriptions_Core_Plugin {
 	 *
 	 * @var string
 	 */
-	protected $plugin_version = '4.0.0';
+	protected $plugin_version = '3.1.6';
 
 	/**
 	 * The subscription scheduler instance.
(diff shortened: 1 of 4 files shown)


includes/class-wc-subscriptions-core-plugin.php
```diff
@@ -286,15 +286,6 @@ public function get_product_type_name() {
 		return 'subscription';
 	}
 
-	/**
-	 * Gets the activation transient name.
-	 *
-	 * @return string The transient name used to record when the plugin was activated.
-	 */
-	public function get_activation_transient() {
```
mattallan: Let's keep this `get_activation_transient()` functionality in WCS Core. While we're not using in core at the moment, this function could be helpful down the track so it's best to avoid moving it now and then having to move it back or duplicate this function.

Same goes for setting the transient below:

```
set_transient( $this->get_activation_transient(), true, 60 * 60 );
```

mattallan (commented):
Changes here look good @leonardola, I've just requested a small change :)
Once that's been addressed this is good to go

includes/class-wc-subscriptions-core-plugin.php
haszari: I'd lean the other way @mattallan  - to avoid coupling between subs-core and the pro extension. The activation of an extension seems entirely the extension's business, so I'd be happy for this to live in pro only, assuming we're not using this for WC Pay at all.

haszari: (I'll leave the decision in your hands @mattallan - wanted to share my rationale for the alternative.)

mattallan:
Thanks @leonardola I've just rebased this on the latest trunk and added a changelog entry. This is looking good! 💯 Merging :)

PR #15, opened 2021-10-21
Hide Paypal config notice when WC Subscriptions pro welcome is being shown

leonardola:
related to #8 

On [this comment](https://github.com/woocommerce/woocommerce-subscriptions/pull/4228#issuecomment-948083392) @mattallan suggested that we should include a validation to avoid showing Paypal Config notices when WC Subscriptions pro welcome is being shown.


haszari:
Thanks @leonardola ! I've set the milestone for this to `3.2 WCPay Subscriptions` - @mattallan can you confirm if we need this change in 3.2 and get it wrapped up - thanks 👍

mattallan:
Thanks @leonardola, this change looks good to me!
I just pushed up bdb2335 to fix a few missing whitespaces but apart from that this is good 💯 

Merging this

PR #16, opened 2021-10-22
Don't redirect customers to a Welcome to Subscriptions 2.1 about us page... ever

mattallan:
#### Description

If a store is using a really old version of Subscriptions (i.e. 1.5.0), or once had a very old version of subscriptions and then stopped using our extension, when they update to WC Payments 3.2 (which will come with Subscriptions 3.1.6) they'll see a "Welcome to Subscriptions 2.1" page 🙈 

image

Now that we're in version 3.1.6, we never want to show this page to customers so in this PR I've change the upgrade process to redirect customers to the **WooCommerce > Home** page.

#### Testing Instructions

1. Make sure WC Subscriptions extension is deactivated and WC Payments is activated
2. With `woocommerce-subscriptions-core` plugin activated, checkout this branch
3. Set `woocommerce_subscriptions_previous_version` to `1.1.0` and `woocommerce_subscriptions_active_version` to `1.5.0` to trigger all the upgrade processes from running
4. Try to visit any WP Admin page and you'll see an upgrade database prompt
5. Continue through the upgrade process
6. Once complete, click `Continue`
7. You should be redirected to the WooCommerce Home page
8. On `trunk` you'll see some fatal errors (fixed in this PR) but once those are addressed you'll be redirected to the Subscriptions 2.1 landing page

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -5,6 +5,8 @@
 * Fix: Correctly show the available payment methods when paying for a subscription renewal order. PR#9
 * Fix: Don't show the WC Subscriptions extension welcome/installation message after installing WCS Core. PR#11
 * Fix: Remove the "Welcome to Subscriptions" notice that is displayed upon upgrading from previous minor versions. PR#14
+* Fix: Don't display a "Welcome to Subscriptions 2.1" for stores that have upgraded from really old version of Subscriptions. PR#16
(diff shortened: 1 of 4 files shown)


james-allan (approved):
LGTM. 👍

PR #19, opened 2021-10-25
Fix payment method not updated in subscriptions via "Default payment method"

a-danae:
Fixes WCPay's [#3171](https://github.com/Automattic/woocommerce-payments/issues/3171)

When updating the payment method of all subscriptions as a shopper via My Account -> Payment methods -> Make default:

* Add the selected token to the token list of the subscription
* If both the old and new token belong to the same gateway, update the token in the subscription
* If the old and new token belongs to different gateways, update the subscription's payment method to the new one

#### Testing instructions

1. As a shopper, purchase a subscription with a test payment method
2. Purchase another subscription with a different payment method
3. Go to My Account -> Payment Methods and select "Make default" on a payment method
4. Click "Yes" in the "Would you like to update your subscriptions to use this new payment method - xxxx ending in xxxx (expires xx/xx)?"
5. Go to My Account -> Subscriptions
6. Opposite to step 6 on the related issue, notice that payment methods associated with the subscriptions are updated

To test that the payment for the subscription does use the updated payment method:
1. As a shopper, purchase a subscription with a test payment method
2. As an admin, enable Billing clocks in WCPay Dev Tools if not enabled already
3. As an admin, go to the newly added subscription (WooCommerce -> Subscription -> "Edit" on the latest subscription)  and enable Dev Testing Mode
4. Log in as incognito with the newly created user for the subscription to proceed as a shopper
5. Follow the testing steps mentioned before, starting from step 2
6. As an admin, go edit the subscription. Advance the billing clocks until you trigger the `invoice.paid` event
7. View the subscription in Stripe, and under "Invoices" go to the latest invoice (it should show up as "paid")
8. In the invoice's details, under "Payment method", check that the details match the payment method you selected in the store as a shopper
image

### DIFF
--- a/includes/class-wcs-payment-tokens.php
+++ b/includes/class-wcs-payment-tokens.php
@@ -47,6 +47,21 @@ class WCS_Payment_Tokens extends WC_Payment_Tokens {
 		$subscription->update_meta_data( $token_meta_key, $new_token->get_token() );
 		$subscription->save();
 
+		$available_gateways        = WC()->payment_gateways->get_available_payment_gateways();
+		$new_token_payment_gateway = $new_token->get_gateway_id();
+
+		// Add the new token to the subscription.
+		if ( isset( $available_gateways[ $new_token_payment_gateway ] ) ) {
+			$gateway_instance = $available_gateways[ $new_token_payment_gateway ];
+			$gateway_instance->add_token_to_order( $subscription, $new_token );
+
+			$payment_meta_table = self::get_subscription_payment_meta( $subscription, $new_token_payment_gateway );
+
+			if ( is_array( $payment_meta_table ) ) {
+				WC_Subscriptions_Change_Payment_Gateway::update_payment_method( $subscription, $new_token->get_gateway_id(), $payment_meta_table );
+			}
+		}
+
 		// Copy the new token to the last renewal order if it needs payment so the retry system will pick up the new method.
 		$last_renewal_order = $subscription->get_last_order( 'all', 'renewal' );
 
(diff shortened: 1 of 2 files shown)


leonardola:
This will fix part of [this issue](https://github.com/Automattic/woocommerce-payments/issues/3209) too

includes/class-wcs-payment-tokens.php
```diff
@@ -47,6 +47,27 @@ public static function update_subscription_token( $subscription, $new_token, $ol
 		$subscription->update_meta_data( $token_meta_key, $new_token->get_token() );
 		$subscription->save();
 
+		$available_gateways        = WC()->payment_gateways->get_available_payment_gateways();
+		$new_token_payment_gateway = $new_token->get_gateway_id();
+
+		// Add the new token to the subscription.
+		if ( isset( $available_gateways[ $new_token_payment_gateway ] ) ) {
+			$gateway_instance = $available_gateways[ $new_token_payment_gateway ];
+			$gateway_instance->add_token_to_order( $subscription, $new_token );
```
mattallan, line 56: The `$gateway->add_token_to_order()` is only found on the WCPay gateway (not on Stripe or Square for example), so I don't think we can use this here inside WC Subscriptions Core.

mattallan (commented):
Hey @a-danae 👋 Sorry this PR hasn't received much attention!

I still need to put more thought into these changes but I'm wondering how these changes will affect the WC Subscriptions extension 🤔 Does this bug exist when WC Subscriptions is active or is it specific to the WCPay Subscriptions feature.?

Anyway, I've left a comment about using a WCPay Gateway function which won't work in WC Subscriptions when other gateways like Stripe and Square are being used :(

Thinking about these changes in the context of WC Subscriptions and WC Payments might complicate the fix a little

a-danae:
No worries, thanks for jumping in! 🙂  And apologies for my delayed reply

Ah, of course. I mistakenly assumed this method was shared between gateways.

> Does this bug exist when WC Subscriptions is active or is it specific to the WCPay Subscriptions feature.?

Testing it now, this works fine with WC Subscriptions and Stripe. So it looks like things are okay in the core and this would be specific to WCPay. Closing this PR and checking out a different approach.

Thanks for the guidance @mattallan!

PR #23, opened 2021-10-28
Remove trailing comma in function arguments.

shendy-a8c:
Related to https://github.com/Automattic/woocommerce-payments/issues/3244

Remove trailing comma in function arguments to avoid PHP fatal error for sites with by PHP 7.2.

Testing instructions.

I can't think of other easy and sure way to confirm that I get all the trailing commas than this way:
1. Load this PR's subscription-core change from [woocommerce-payments](https://github.com/Automattic/woocommerce-payments).
Change this line in [woocommerce-payments' composer.json](https://github.com/Automattic/woocommerce-payments/blob/develop/composer.json):
```
"woocommerce/subscriptions-core": "1.0.1"
```
to
```
"woocommerce/subscriptions-core": "dev-fix/fatal-error-trailing-comma-function-args"
```
Then, run `composer update`.
Confirm that the changes from this PR present in the `vendor/woocommerce/subscriptions-core`, for example `vendor/woocommerce/subscriptions-core/includes/class-wc-subscriptions-cart.php` on line 1022 should not have trailing comma.

2. Change [woocommerce-payments' phpcs.xml.dist](https://github.com/Automattic/woocommerce-payments/blob/develop/phpcs.xml.dist) locally.
Remove
```
<exclude-pattern>./vendor/*</exclude-pattern>
```

3. Run `phpcs` on `vendor/woocommerce/subscriptions-core` files:
```
./vendor/bin/phpcs --standard=phpcs.xml.dist $(find ./vendor/woocommerce/subscriptions-core -name '*.php')
```
There will be a lot of errors and warnings. That's OK.

4. With base branch, expect there are 4 `PHPCompatibility.Syntax.NewFunctionCallTrailingComma.FoundInFunctionCall` errors.
With head branch, there shouldn't be that error at all.

To load subscriptions-core's trunk, change this line in [woocommerce-payments' composer.json](https://github.com/Automattic/woocommerce-payments/blob/develop/composer.json):
```
"woocommerce/subscriptions-core": "1.0.1"
```
to
```
"woocommerce/subscriptions-core": "dev-trunk"
```
to load this PR's head branch, change that line to:
```
"woocommerce/subscriptions-core": "dev-fix/fatal-error-trailing-comma-function-args"
```
Then, run `composer update`.

### DIFF
--- a/includes/class-wc-subscriptions-cart.php
+++ b/includes/class-wc-subscriptions-cart.php
@@ -1019,7 +1019,7 @@ class WC_Subscriptions_Cart {
 						'carts_with_multiple_payments' => $carts_with_multiple_payments,
 					),
 					'',
-					WC_Subscriptions_Core_Plugin::instance()->get_subscriptions_core_directory( 'templates/' ),
+					WC_Subscriptions_Core_Plugin::instance()->get_subscriptions_core_directory( 'templates/' )
 				);
 			}
 
--- a/includes/class-wc-subscriptions-email.php
+++ b/includes/class-wc-subscriptions-email.php
@@ -288,7 +288,7 @@ class WC_Subscriptions_Email {
 				'order_items_table_args' => $order_items_table_args,
 			),
 			'',
-			WC_Subscriptions_Core_Plugin::instance()->get_subscriptions_core_directory( 'templates/' ),
+			WC_Subscriptions_Core_Plugin::instance()->get_subscriptions_core_directory( 'templates/' )
 		);
 	}
 
--- a/includes/class-wc-subscriptions-order.php
+++ b/includes/class-wc-subscriptions-order.php
@@ -852,7 +852,7 @@ class WC_Subscriptions_Order {
 					'subscriptions' => $subscriptions,
 				),
 				'',
-				WC_Subscriptions_Core_Plugin::instance()->get_subscriptions_core_directory( 'templates/' ),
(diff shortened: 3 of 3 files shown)


RadoslavGeorgiev:
Went though the testing instructions, and there are no more `Trailing comma's are not allowed in function calls in PHP 7.2 or earlier 
(PHPCompatibility.Syntax.NewFunctionCallTrailingComma.FoundInFunctionCall)` errors with the contents of this PR 🎉 

There are a couple of PHP **warnings** like this: `Since PHP 7.0, functions inspecting arguments, like debug_backtrace(), no longer report the original value as passed to a parameter, but will instead provide the current value. The parameter "$new_callback_args" was used, and possibly changed (by reference), on line 86. (PHPCompatibility.FunctionUse.ArgumentFunctionsReportCurrentValue.NeedsInspection)`

We're only supporting 7.0+ though, and this is only a warning, so it should be fine.

Proceeding to create a test site...

haszari (approved):
Approving based on @RadoslavGeorgiev phpcs testing and confirm this catches the cases mentioned in the [WPORG thread](https://wordpress.org/support/topic/fatal-php-errors-in-version-3-2-0-php-7-2-34/).

Merging so we can tag a 1.0.2 subs core and include in WCPay patch release.

chickenn00dle:
Just noting that I ran PHPCS on Subscriptions Core codebase using a filtered down ruleset from WCPay that didn't include things like comment, short array syntax, etc., and didn't find anything else that seemed like it would break the plugin.

PR #30, opened 2021-10-30
Limit period intervals when creating a product to "1" when the period is set to "year" 

a-danae:
Fixes WCPay's [#3255](https://github.com/Automattic/woocommerce-payments/issues/3255)

On client-side, when changing the product's period to "year":
* Change the selected period interval value to "1"
* Disable the select for the period interval

On server-side:
* Force the period interval to "1" when the period is set to "year" when saving

These, for both simple and variable products.

Not adding the testing instructions since we won't go this way.

### DIFF
--- a/assets/js/admin/admin.js
+++ b/assets/js/admin/admin.js
@@ -373,6 +373,18 @@ jQuery( function( $ ) {
 
 			$( '#_subscription_one_time_shipping' ).prop( 'disabled', is_synced_or_has_trial );
 		},
+		// Sets the interval to 1 and disables its select when the selected period is yearly.
+		// Stripe doesn't allow greater intervals for yearly subscriptions.
+		limitYearlyPeriodInterval: function( $periodSelect ) {
+			const $intervalSelect = $periodSelect.closest( '._subscription_price_field').find( '.wc_input_subscription_period_interval' );
+
+			if ( 'year' === $periodSelect.val() ) {
+				$intervalSelect.val( 1 );
+				$intervalSelect.prop( 'disabled', true );
+			} else {
+				$intervalSelect.prop( 'disabled', false );
+			}
+		},
 		showHideSubscriptionsPanels: function() {
 			var tab = $( 'div.panel-wrap' ).find( 'ul.wc-tabs li' ).eq( 0 ).find( 'a' );
 			var panel = tab.attr( 'href' );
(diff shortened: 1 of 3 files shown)


mattallan:
Hey @a-danae, thanks for looking into this one!
Because these changes are being made to WCS Core, they will also affect our WC Subscriptions premium extension so we'll have to apply the opposite of these changes to re-enable longer than 1 year subscription cycles.

Something else I was curious about is whether Stripe is okay with Subscriptions that are longer than a year (but with a shorter period), i.e. renewing every 53 weeks, or 15 months 🤔 

Because this is a limitation which is specific to WC Payments integration with Stripe Billing, I wonder if we should add some sort of validation to WC Payments code when saving the subscription product, instead of adding these changes to WCS Core.

I haven't looked into this too deeply but one approach that might work:
 - Inside WC Payments Product service class, hook onto `save_post` with priority 10 (must be before Subscriptions saves the interval/period postmeta)
 - Validate the period/interval
 - If it's invalid, unset the `$_POST` so that Subscriptions doesn't try to save it
 - Add some admin notice notifying the store manager that this combination is invalid.

It would be nice if we could remove the "every x years" option when creating the Subscription product but I think this is too fiddly because we have to make sure it only does this for the "free" version of subscriptions in WC Payments and not for stores that have purchased the extension.

Keen to hear your thoughts on this 😃

a-danae:
Hi there @mattallan!

Gotcha. I missed this limitation wouldn't be needed for WC Subscriptions premium, makes sense. Thanks for pointing this out and for the feedback 😄 

> I haven't looked into this too deeply but one approach that might work:

Closing this PR and going that way.

> Something else I was curious about is whether Stripe is okay with Subscriptions that are longer than a year (but with a shorter period), i.e. renewing every 53 weeks, or 15 months

I just gave this a try, and nope. Stripe doesn't allow those either. We get:
`ERROR The interval_count for monthly prices can't be greater than 12 (invalid_request_error)`
`ERROR The interval_count for daily prices can't be greater than 365 (invalid_request_error)`
`ERROR The interval_count for weekly prices can't be greater than 52 (invalid_request_error)`
But the options for intervals we give are to "6". 6 days, 6 weeks, 6 months, so I guess we don't have to worry about this, right? I tried it by editing the code.

> It would be nice if we could remove the "every x years" option when creating the Subscription product but I think this is too fiddly
Yeah, probably. I'll go that way and see if there's a simple approach that would affect WCPay's subscriptions alone. In this approach, I've just disabled that select. Let's see how to handle it.

I'm adding the description and closing this PR.

PR #40, opened 2021-11-09
Prevent fatal errors caused by orders without a created date.

shendy-a8c:
Fixes 3939-gh-woocommerce/woocommerce-subscriptions

Copied from 4132-gh-woocommerce/woocommerce-subscriptions:

### Description

This PR fixes a fairly obscure issue that can occur when an order doesn't have a date created. Because we got the created date's timestamp on a single line with `wcs_get_objects_property( $order, 'date_created' )->getTimestamp();`, if the date didn't exist, a fatal error would occur (see below). This PR fixes that error.

```
An error of type E_ERROR was caused in line 30 of the file html-related-orders-row.php. 
Error message: Uncaught Error: Call to a member function getTimestamp() on null in .
```

While editing the file I also removed all uses of `wcs_get_objects_property()` in the file.

### Steps to test:

I wasn't able to replicate a scenario when the order has no created date. This PR is just rearranging things, to be more defensive so test and review the changes. 

1. View the edit admin screen for a subscription or a subscription-related order.
1. In the related orders tables, the content should remain unchanged.

-------------------

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Subscriptions Core Changelog ***
 
+2021-x-x - version 1.0.4
+* Fix: Fatal error due to order with no created date in order row template. PR#40
+
 2021-10-29 - version 1.0.3
 * Fix: Errors when attempting to get the plugin version during PayPal requests. PR#27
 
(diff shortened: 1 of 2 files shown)


brucealdridge (commented):
Have tested and works as expected.

brucealdridge (approved):
Forgot to approve, 👍

PR #41, opened 2021-11-09
Fix: Misleading order note on payment method change

brucealdridge:
Fixes 3268-gh-woocommerce/woocommerce-subscriptions 

Order note when payment method is changed can be misleading and confusing.

Previously the order note used to read `Payment method changed from “[OLD_METHOD]” to “[NEW_METHOD]” by the subscriber from their account page` however the change didn't always take place on the account page. This message was also triggered when the user updated the payment method when the payment failed.

This fix updates the order note to only mention the new payment method: `Payment method changed from "[OLD_METHOD] to “[NEW_METHOD]” by the subscriber.` removing any confusion.

**Before**
image

**After**
image

This change will affect both woocommerce-subscriptions and woocommerce-payments flows.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -4,6 +4,7 @@
 * Fix: Add consistent margins to the recurring taxes totals row on the Checkout and Cart block. PR#39
 * Fix: Fatal error due to order with no created date in order row template. PR#40
 * Fix: Fatal error on the customer payment page for renewal orders with deleted products. PR#42
+* Fix: Misleading order note on payment method change. #PR41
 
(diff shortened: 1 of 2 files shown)


mattallan (approved):
This is a nice easy one to review 😄 

Would you mind updating this branch to the latest `trunk` and adding a changelog entry for this [here](https://github.com/Automattic/woocommerce-subscriptions-core/blob/915df670266d4ef2bb50bcf702e94dc9224f501b/changelog.txt#L3) please? 🙏 
Apart from that, this is good to merge!

PR #43, opened 2021-11-12
v1.1.0 Release

aprea:
```
2021-11-12 - version 1.1.0
* Fix: Add consistent margins to the recurring taxes totals row on the Checkout and Cart block. PR#39
* Fix: Fatal error due to order with no created date in order row template. PR#40
* Fix: Fatal error on the customer payment page for renewal orders with deleted products. PR#42
* Fix: Misleading order note on payment method change. PR#41
```


mattallan (approved):
🚀 

Thanks for including the built JS files as this release includes some Checkout block updates :+1:
This is good to merge!

PR #44, opened 2021-11-14
adjust readme to clarify that this is a package

haszari:
This PR updates the readme to further clarify that this is a component package of other products. 

The goal of this PR is to reduce confusion that this is a product or can be used standalone. 

Changes:

- Explain upfront that this is a library / package used to build other products.
- Link out to those products up front.
- Remove products that depend on this from `Dependencies` list.
- Use lowercase name to reinforce that this is not a product.
- Shift content about running as a plugin to `Development` section to focus on normal, supported use as part of another plugin.

#### To test the changes in this PR
- Review the readme in rich view. 
- Is it correct?
- Is anything unclear or misleading?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,17 +1,19 @@
-# WooCommerce Subscriptions Core
+# WooCommerce subscriptions-core
 
-This package adds core subscriptions functionality to your WooCommerce store.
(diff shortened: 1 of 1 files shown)


brucealdridge (approved):
Much clearer.

PR #49, opened 2021-11-19
Deprecate the admin error notice when downgrading subscriptions to a previous minor version

### LINKED ISSUE #12: Investigate whether we still need to show a downgrade notice for every major version
Related WC Subscriptions issue: https://github.com/woocommerce/woocommerce-subscriptions/issues/4271

---

In WCS Core, we display a warning to store managers when they downgrade their Subscriptions version to a previous major version: i.e. 
4.0.0 -> 3.1.6

As mentioned on https://github.com/Automattic/woocommerce-subscriptions-core/issues/7#issuecomment-943799044, this notice is somewhat not relevant these days and was originally added to Subscriptions to notify store managers that have downgraded from 2.0 back to 1.5.

This is because in 2.0, Subscriptions changed significantly from being array stored on an order, to a custom order type, so downgrading back to 1.5 was really bad.

These days downgrading from Subscriptions 4.0 to 3.0 isn't a big deal at all so this issue is to look at updating when we show this notice:

Some options:
 - downgraded 2 major version
 - downgraded back to 1.5.x

mattallan:
Fixes #12

 - Dont' show a downgrade notice whenever subscriptions has been downgraded by a minor version

This admin notice was previously used to make store managers aware that downgrading Subscriptions to a previous version could have consequences however, this was really only the case for stores that downgraded to much older version of Subscriptions (like 1.5.0).

Because the subscriptions core library is being used in two extensions, there's now a much higher chance that stores will downgrade their subscriptions version but it's not a big deal anymore.

In this PR, I've deprecated this downgrade notice so that it's no longer shown.

#### Testing instructions

1. Clone this repo into `wp-content/plugins`
1. Activate the WooCommerce Subscriptions Core plugin so that these changes are loaded by WC Payments and WC Subscriptions
1. Activate the Woocommerce Subscriptions extension (trunk version)
1. While on `trunk`of this repo, go into `woocommerce-subscriptions.php` and change the `$version` static variable to something like: `'2.9.0'`
1. Load any admin page and you should see a "downgraded notice"
1. Checkout this branch 
1. No admin notice

-------------------

- [ ] Covered with tests (or have a good reason not to test in description ☝️)
- [ ] Tested on mobile (or does not apply)

**Post merge**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -2,6 +2,7 @@
 
 2021-xx-xx - version 1.x.x
 * Fix: Update tooltip wording when deleting product variation. PR#46
+* Fix: Don't show an admin error notice when a store downgrades to a previous minor version of Subscriptions. WCS#4271
 
 2021-11-12 - version 1.1.0
 * Fix: Add consistent margins to the recurring taxes totals row on the Checkout and Cart block. PR#39
--- a/includes/upgrades/class-wc-subscriptions-upgrader.php
+++ b/includes/upgrades/class-wc-subscriptions-upgrader.php
@@ -101,7 +101,6 @@ class WC_Subscriptions_Upgrader {
 		// When WC is updated from a version prior to 3.0 to a version after 3.0, add subscription address indexes. Must be hooked on before WC runs its updates, which occur on priority 5.
 		add_action( 'init', array( __CLASS__, 'maybe_add_subscription_address_indexes' ), 2 );
 
-		add_action( 'admin_notices', array( __CLASS__, 'maybe_add_downgrade_notice' ) );
 		add_action( 'admin_notices', array( __CLASS__, 'maybe_display_external_object_cache_warning' ) );
 
 		add_action( 'init', array( __CLASS__, 'initialise_background_updaters' ), 0 );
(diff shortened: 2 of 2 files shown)


shendy-a8c:
Looks good, @mattallan!

I wonder why you moved the `maybe_add_downgrade_notice()` code block, just so deprecated function is at the bottom of the file?

Also, there is linting problem that needs fixing. Other than that, everything looks good.

shendy-a8c (approved):

mattallan:
Thanks @shendy-a8c for the quick review!

> why you moved the `maybe_add_downgrade_notice()` code block, just so deprecated function is at the bottom of the file?

Oh yeah this might be a thing we only did in Subscriptions 😅 We usually move deprecated functions down to the bottom of the page (underneath the `/* Deprecated Functions */` line).

I noticed some a new function was added in between deprecated functions so I've moved that up.

> Also, there is linting problem that needs fixing. Other than that, everything looks good.

Thanks! I've pushed up some PHPCS fixes and it looks like all the checks are passing now 👯 
Merging this now

shendy-a8c:
@mattallan, this is a really good call: https://github.com/Automattic/woocommerce-subscriptions-core/pull/49/commits/fcc1c1966d72eb00f2a54638c32c62bb92860905

PR #50, opened 2021-11-22
v1.2.0 release.

shendy-a8c:

- Increase version from `1.1.0` to `1.2.0` in `package.json` and `woocommerce-subscriptions-core.php`.
- Ran `composer update` since there is a change (version increase) in `composer.json`, resulting a change in `composer.lock`.
- Ran `npm install` resulting a change in `package-lock.json`. I needed `wp-scripts` on my local machine so I can run `npm run build:js` to make sure Javascript files are up-to-date.
- Set date and version in changelog.

-------------------

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,6 @@
 *** WooCommerce Subscriptions Core Changelog ***
 
-2021-xx-xx - version 1.x.x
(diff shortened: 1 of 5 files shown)


mattallan (approved):
Thanks, Shendy! 🎉  LGTM

PR #54, opened 2021-11-30
Update deprecated WooCommerce Blocks hook

opr:
## Description

This change is needed because recently WooCommerce Blocks graduated some of its experimental hooks (https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5017 and https://github.com/woocommerce/woocommerce-gutenberg-products-block/pull/5014) to stable, so the `__experimental` prefix was dropped. This PR updates the use of those hooks to the stable version.

Without this PR, deprecation warnings are shown in the Checkout Block.

This code can break if the user does not update their version of WooCommerce Blocks to the latest version that includes the graduated hooks.

Tips/tricks on how to mitigate this would be much appreciated, because I'm really unsure on how we can cater for this situation.

## How to test this PR

Place an order that includes subscription items and ensure the Checkout page does not show any warnings. Ensure the order in WP Admin is correct and has all the expected information.

## Product impact

- [x] Added changelog entry (~~or does not apply~~)
- [x] Will this PR affect WooCommerce Subscriptions? yes/~~no/tbc, add issue ref~~
^ I'm not 100% confident on the above, whether WC Payments would call these methods, I think it won't but could someone who knows more 👍🏼  this one?


includes/class-wc-subscriptions-checkout.php
```diff
@@ -24,7 +24,7 @@ public static function init() {
 		add_action( 'woocommerce_checkout_order_processed', array( __CLASS__, 'process_checkout' ), 100, 2 );
 
 		// Same as above, but this is for the Checkout block.
-		add_action( '__experimental_woocommerce_blocks_checkout_order_processed', array( __CLASS__, 'process_checkout' ), 100, 1 );
+		add_action( 'woocommerce_blocks_checkout_order_processed', array( __CLASS__, 'process_checkout' ), 100, 1 );
```
mattallan: @opr to maintain backwards compatibility, what are your thoughts on surrounding these with version checks (see suggestion below)?

```suggestion
		if ( class_exists( 'Automattic\WooCommerce\Blocks\Package' ) && version_compare( \Automattic\WooCommerce\Blocks\Package::get_version(), '6.3.0', '>=' ) ) {
			add_action( 'woocommerce_blocks_checkout_order_processed', array( __CLASS__, 'process_checkout' ), 100, 1 );
		} else {
			add_action( '__experimental_woocommerce_blocks_checkout_order_processed', array( __CLASS__, 'process_checkout' ), 100, 1 );
		}
```

mattallan (commented):
Thanks @opr for working on this for us! 🙏 

> Tips/tricks on how to mitigate this would be much appreciated, because I'm really unsure on how we can cater for this situation.

I've left a comment with a suggestion on how we could work with old and new version of blocks. Let me know what your thoughts are on this approach!

Thanks!

includes/class-wc-subscriptions-checkout.php
opr: Thanks for that suggestion, seems super obvious now 🤦🏼  I also wrapped it in a `is_experimental_build` so we don't call the experimental hooks if we're in a dev version of the plugin. It seems `version_compare` doesn't like the `-dev` suffix we use.

mattallan (approved):
Thanks @opr!

I've tested these changes on WooCommerce Blocks 6.4 (latest) and 6.2 and everything appears to be working in order :+1:

Marking that as good to merge!

PR #68, opened 2021-12-09
Migrates the WCS unit test helper classes/functions to subscriptions core

mattallan:
Fixes #60

## Description

This PR migrates the `woocommerce-subscriptions/tests/framework/helpers` over to the subscriptions core codebase so that we can use them in our new unit tests.

During the move, I changed a few minor things with these files:
 - Removed WC pre 3.x compatibility code
 - Updated arrays to be written as `[]` instead of `array()`
 - Cleaned up some of the doc blocks and whitespace
 - Changed the path of the helpers to be consistent with WC Payments (`tests/unit/helpers`)
 - `WCS_Helper_Coupon` no longer extends `WC_Helper_Coupon`

## How to test this PR

To test this I just tried using some of our helpers in the `Simple_Pass_Test->test_always_pass()` to check they're working.

PR #69, opened 2021-12-09
Migrate global function unit tests

### LINKED ISSUE #61: Migrating Unit Tests: wcs global functions unit tests
Working on this relies on the unit test helpers being migrated: https://github.com/Automattic/woocommerce-subscriptions-core/issues/60
Related to https://github.com/woocommerce/woocommerce-subscriptions/issues/4195

---

## Description

This issue will tackle migrating all of the unit tests for our `wcs-x-functions.php` files from `woocommerce-subscriptions/tests/unit/` over to the the subscriptions core repository.

The unit tests that need to be moved include:
 - [x] `wcs_test_wcs_helper_functions.php`
 - [x] `wcs_test_wcs_functions.php`
 - [x] `wcs_test_wcs_compatibility_functions.php`
 - [x] `wcs_test_wcs_cart_functions.php`
 - [x] `wcs_test_user_functions.php`
 - [x] `wcs_test_time_functions.php`
 - [x] `wcs_test_order_functions.php` (move the one unit test from `orders/functions.php` to this file as well)
 - [x] `wcs_test_deprecated_functions.php`

As part of doing this move we should also:
 - [x]	rename the files to `test-{file-name}.php` instead of `wcs_test_{file_name}.php`
 - [x] each test class should extend `WP_UnitTestCase` instead of `WCS_Unit_Test_Class` (I don't think we'll need this custom unit test class anymore)

brucealdridge:
Fixes #61

## Description

Migrated global function unit tests from main subscription repository to subscriptions-core so we can have the benefit of tests on the core library.

As part of this PR, new unit tests should run w/ github actions on each PR. 

## How to test this PR

1. Run `composer test` at the cli.
2. Verify that all tests pass

## Product impact

- [x] ~Added changelog entry~ - **does not apply**


brucealdridge:
@mattallan I'm having issues with one test. Can you see what the problem may be this this test? 

```
There was 1 failure:

1) Test_Functions::test_wcs_create_subscription_no_order with data set #0 (array('month', 3, 1), array('GBP', 'month', 3, 1, '5.9.0', 'yes', 'yes', 'wc-pending', 0, '', ''))
Failed asserting that two strings are equal.
--- Expected
+++ Actual
@@ @@
-'GBP'
+'USD'
```

build/index.asset.php
```diff
@@ -1 +1,4 @@
-<?php return array('dependencies' => array('wc-blocks-checkout', 'wc-price-format', 'wc-settings', 'wp-element', 'wp-i18n', 'wp-plugins'), 'version' => '74dee4a9fac3d7adb358afc27508dd11');
\ No newline at end of file
+<?php return array(
```
mattallan: This file is a built asset file so we don't need to fix any whitespace in it 😅

tests/unit/bootstrap.php
```diff
@@ -24,7 +24,48 @@
...
+tests_add_filter( 'muplugins_loaded', '_manually_load_plugin' );
+
+// Need those polyfills to run tests in CI.
+require_once dirname( __FILE__ ) . '/../../vendor/yoast/phpunit-polyfills/phpunitpolyfills-autoload.php';
+
+// Start up the WP testing environment.
+require $_tests_dir . '/includes/bootstrap.php';
+require dirname( __FILE__ ) . '/helpers/class-wcs-helper-coupon.php';
+require dirname( __FILE__ ) . '/helpers/class-wcs-helper-product.php';
+require dirname( __FILE__ ) . '/helpers/class-wcs-helper-subscription.php';
```
mattallan: Just making a note that I noticed in my own testing that I didn't need to manually require these helper classes. Well, some of them I do but some of them, like `class-wcs-helper-subscription.php` I didn't need to load.

Is there an autoloader issue we need to address so we don't have to load all of these?
Because it was working for some classes and not others I tried running `composer dump-autoload` but that didn't seem to fix it.

Curious if you have any ideas :)

tests/unit/test-autoloader.php
```diff
@@ -0,0 +1,397 @@
...
+	/**
+	 * Test the should_autoload() method.
+	 *
+	 * @param string $class_name
+	 * @param string $expected
+	 *
+	 * @dataProvider should_autoload_provider
+	 */
+	public function test_should_autoload( $class_name, $expected ) {
+		$should_autoload = $this->get_accessible_protected_method( $this->autoloader, 'should_autoload' );
```
mattallan: > get_accessible_protected_method()

This functionality is used a lot in our tests so I've just pushed up https://github.com/Automattic/woocommerce-subscriptions-core/pull/68/commits/2cf36b41404f715a2ea0d6a949e686eebeea8293 to add this helper.

This is the same function used in [WC Payments](https://github.com/Automattic/woocommerce-payments/blob/develop/tests/unit/helpers/class-phpunit-utils.php) so I ported it over here :)

tests/unit/test-cart-functions.php
```diff
@@ -0,0 +1,287 @@
+<?php
+/**
+ *
+ */
+class Test_Cart_Functions extends WP_UnitTestCase {
```
mattallan: File name and class should be `test-wcs-cart-functions.php` and `WCS_Cart_Functions_Test`.

tests/unit/test-early-renewal-functions.php
```diff
@@ -0,0 +1,31 @@
+<?php
+
+class Test_Early_Renewal_Functions extends WP_UnitTestCase {
```
mattallan: Early renewals are not a feature in Subscriptions Core so we don't need unit tests for this.

The function is being tested here (`wcs_get_last_non_early_renewal_order()`) is part of `wcs-renewal-functions.php` and doesn't look like it's used by Core at all, but that's okay.

Because this function is in `wcs-renewal-functions.php`, this unit test should live in `test-wcs-renewal-functions.php` class.

mattallan (commented):
Nice one @brucealdridge 

I left a couple of notes inline but just a couple of things I wanted to bring up:

**File names**

> rename the files to `test-{file-name}.php` instead of `wcs_test_{file_name}.php`

I realise this comment on the issue wasn't very clear, but what I meant by this was to add the full filename with `test-` in front of it, i.e. `test-wcs-core-autoloader.php` instead of `test-autoloader.php`.

This keeps things consistent with WC Payments but also a nice feature to this naming convention is that when trying to find files, if you search `wcs-core-autloader.php` you'll both the file and the unit tests for that file :)

**Class names**

I didn't mention this on the issue but I noticed WC and WC Pay names tests classes with `_Test` at the end, not at the start, so I think we should stick with the same naming convention.

mattallan:
> Can you see what the problem may be this this test?

I think this test is calling `wcs_create_subscription()` without sending a currency and is expecting to use the default store currency which used to be set to `"GBP"`, but in our new bootstrap we're setting the default currency to USD

build/index.asset.php
brucealdridge: looks like there was an oddity in `phpcs.dist.xml` that required it.

tests/unit/bootstrap.php
brucealdridge: I'm sure they weren't working for me at some point, but now they magically work. removed now. (bootstrap still required)

PR #70, opened 2021-12-09
Update recurring payments copy on payment gateways page

### LINKED ISSUE #52: Adjust `Recurring Payments` copy in payment gateways page
## Description

In the payment gateways settings screen, there's a section prompting merchant to install other gateways for more auto-renew payment options.

This messaging should only be displayed when the store has WooCommerce Subscriptions active. With WC Payments + Subscriptions Core, WCPay is the only supported gateway.

image

### Testing instructions

1. Make sure WC Payments is active and deactivate WC Subscriptions plugin
2. From the WP admin dashboard, navigate to WooCommerce > Settings > Payments
3. Click on the All payment method tab
3. Scroll down to the bottom of the page and notice the recurring payments section which tells store managers to "Find new gateways that support automatic subscription payments in the official WooCommerce Marketplace."

This wording should be removed from subscriptions core or changed to something else.

## Product impact

- [x] WooCommerce Payments: https://github.com/Automattic/woocommerce-payments/issues/3402
- [x] WooCommerce Subscriptions - Once this is removed from Subscriptions Core, we'll need to add this text back to WC Subscriptions if we think it's still necessary.

brianyu28:
Fixes #52
Related: https://github.com/Automattic/woocommerce-payments/issues/3402

## Description

Currently, subscriptions-core makes reference to finding "new gateways that support automatic subscription payments" on the payment gateways page. This message should only be shown for WooCommerce Subscriptions and not in WCPay Subscriptions (where WCPay is the required gateway).

This PR removes that line from subscriptions-core and passes it through a new filter. A separate PR (https://github.com/woocommerce/woocommerce-subscriptions/pull/4283) in WooCommerce Subscriptions will use the filter to add the line back for WooCommerce Subscriptions only.

## How to test this PR

Go to `/wp-admin/admin.php?page=wc-settings&tab=checkout` and verify that you no longer see the "Find new gateways that support automatic subscription payments" text.

## Product impact

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -4,6 +4,7 @@
 * Fix: Remove references to the Subscription extension in the tooltips found on the Payment Methods settings table. PR#55 wcpay#3234
 * Fix: Update the Automatic Recurring Payments column on the Payment Methods table to only show which payment methods are supported by Subscriptions Core. PR#55
 * Tweak: Update deprecation message when calling WC_Subscriptions_Coupon::cart_contains_limited_recurring_coupon() to mention the correct replacement function. PR#53
+* Tweak: Update recurring payments copy on payment gateways page.
 
 2021-11-23 - version 1.2.0
 * Fix: Update tooltip wording when deleting product variation. PR#46
--- a/includes/admin/class-wc-subscriptions-admin.php
+++ b/includes/admin/class-wc-subscriptions-admin.php
@@ -1742,7 +1742,7 @@ class WC_Subscriptions_Admin {
 					'type' => 'informational',
 				),
 			),
-			WC_Subscriptions_Admin::$option_prefix,
+			self::$option_prefix
 		);
 
 		$insert_index = array_search(
--- a/includes/admin/class-wc-subscriptions-admin.php
+++ b/includes/admin/class-wc-subscriptions-admin.php
@@ -1711,9 +1711,8 @@ class WC_Subscriptions_Admin {
 	}
 
 	/**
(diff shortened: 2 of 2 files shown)


shendy-a8c:
Great work, @brianyu28 !

I tested and it works as expected.

Here are my test cases:
1. Both WCS and WCPay do not have this change.
2. Both WCS and WCPay have this change (and 4283-gh-woocommerce/woocommerce-subscriptions).
3. Only WCPay has this change.
4. Only WCS has this change (and 4283-gh-woocommerce/woocommerce-subscriptions).

The way I tested, I changed `composer.json` of WCS / WCPay and change the version of `woocommerce/subscriptions-core` to `dev-update/52-recurring-payments-copy`. Then, I run `composer update` to fetch this branch.

By the way, good catch on fixing the comment too from `Settings->Checkout->Payment Gateways table` to `Settings->Payments->Payment Methods table`.

I see there is conflict. Fixed it with ea39eb5457afcd154820da5b1d8564e75f6559b8.
Also fixed PHP linting problems with 259f41506b23b29b33e5c5a776ff98462676a7c4.

If it looks good to you, go ahead and merge.

shendy-a8c:
Not sure why JS linting fails. The only changes are only PHP changes, so I think it's safe to merge.

Will investigate further. @brucealdridge maybe can spot what's wrong faster.

I attempt to fix here: https://github.com/Automattic/woocommerce-subscriptions-core/pull/74.

shendy-a8c (approved):

PR #72, opened 2021-12-14
Reduce the scope of the JS linting

aprea:
Hey @brucealdridge, I spent a little bit of time looking at the JS linting.

In this PR I am proposing we further reduce the scope for JS linting by ignoring all files located at `assets/js/*`.

I initially tried including these files but that resulted in a further 650 problems with only 195 errors potentially fixable with the `--fix` option.

One other thing I wanted to ask, I'm a little confused that we have both `format:js` and `lint:js-fix` scripts. The naming suggests these do similar things, do we need both? Or do they do different things?


haszari:
> One other thing I wanted to ask, I'm a little confused that we have both format:js and lint:js-fix scripts. The naming suggests these do similar things, do we need both? Or do they do different things?

Good question!

I don't see `format-js` documented in [`wp-scripts` docs](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-scripts/), so maybe it's not needed or maybe it's deprecated/old. 

If there is a real difference it might be something like:

- `format` does formatting only, like prettify
- `lint` does linter rules, some static/lexical analysis, and maybe some formatting associated with that

We should only have scripts in package-json that are useful to run regularly, so if one of these is not providing value, feel free to kill it (could be a separate PR).

brucealdridge (approved):

brucealdridge:
> > One other thing I wanted to ask, I'm a little confused that we have both format:js and lint:js-fix scripts. The naming suggests these do similar things, do we need both? Or do they do different things?
> 
> Good question!
> 
> I don't see `format-js` documented in [`wp-scripts` docs](https://developer.wordpress.org/block-editor/reference-guides/packages/packages-scripts/), so maybe it's not needed or maybe it's deprecated/old.
> 
> If there is a real difference it might be something like:
> 
> * `format` does formatting only, like prettify
> * `lint` does linter rules, some static/lexical analysis, and maybe some formatting associated with that
> 
> We should only have scripts in package-json that are useful to run regularly, so if one of these is not providing value, feel free to kill it (could be a separate PR).

that looks right to me.

PR #77, opened 2021-12-17
CI: Migrate subscription unit tests

### LINKED ISSUE #64: Migrating Unit Tests: Subscriptions, emails & other files
Working on this issue will rely on the unit test helpers being migrated first: #60
Related to woocommerce/woocommerce-subscriptions#4195

---

## Description

This issue is to track migrating the existing unit tests found on WC Subscriptions repository that are related to the main WC Subscription class, emails and other remaining unit tests like the core autoloader.

Test files that will need to be migrated over include:
- [x] `tests/unit/subscriptions/subscriptions.php`
- [x] `tests/unit/subscriptions/subscriptions-payment-count.php`
  - this is testing `subscription->get_payment_count()` and doesn't need to be a separate class so it should be moved into the above file
- [x] `tests/unit/wcs_emails.php`
- [x] `tests/unit/wcs_post_meta_cache_manager.php`
- [x] `tests/unit/wcs_test_autoloader.php`

As part of moving these files over, we should also:

- [x] use consistent naming convention i.e. `test-{file-name}.php`
- [x] each test class should extend `WP_UnitTestCase` instead of `WCS_Unit_Test_Class`
- [x] keep the `tests/unit/` file structure similar to that of `subscriptions-core/includes`. Rather than having `unit/subscriptions` this should go in `tests/unit/test-{file-name}.php`

brucealdridge:
Fixes #64 

## Description

Migrate subscriptions based tests from private subscriptions codebase.

## How to test this PR

This PR should be auto-testing in that if all actions pass it should be good to go.
Otherwise tests can be manually run by `composer test`


assets/js/admin/payment-method-restrictions.js
```diff
@@ -1,91 +1,152 @@
...
-		if ( element.parent().find( '.wc_error_tip' ).length === 0 ) {
-			element.after( '<div class="wc_error_tip ' + error_type + '">' + wcs_gateway_restrictions[ error_type ] + '</div>' );
-			element.parent().find( '.wc_error_tip' )
-				.css( 'left', offset.left + element.width() - ( element.width() / 2 ) - ( $( '.wc_error_tip' ).width() / 2 ) )
-				.css( 'top', offset.top + element.height() )
-				.fadeIn( '100' );
-		}
-	})
-} );
+jQuery(function ($) {
```
mattallan: I don't think these edits are supposed to be here? 😅 I'm assuming this was accidentally carried over from the formatting JS work?

brucealdridge: oops, likely happened due to changing branches and doing `git commit -a`

mattallan (approved):
Hey @brucealdridge , the email tests seem a bit flakey so I'm wondering if we should just remove them for now? Some things I noticed:
 - the email tests fail when I run them in isolation (`vendor/bin/phpunit --filter WCS_Subscriptions_Email_Test`)
 - when I rename the email tests file to `test-class-wc-subscriptions-email.php` they also start failing, so it seems like they only work when they're ran in a specific order? (i have no idea :joy:)

Because there's only one email test and it's testing the WC Failed Order email (a non subscription email), I'm thinking we leave these out for now and we can work on them at a later time. How does that sound to you?

Also, I noticed the tests were taking an extra 30 seconds on this branch and I think this is mainly because the setUp function in the WC Subscriptions unit tests was creating an array of 7 subscriptions every time the tests were run. Since we use a lot of data providers in this class, we must have been creating 1000s of subscriptions which wasn't necessary.

To improve this a little bit, I've just pushed up some changes which stop creating subscriptions for every test and instead, only create them once in the `setUpBeforeClass()` function. Most tests didn't change the state of these subscriptions, but for those that did, I've updated them to use their own new subscription.

Some other changes I've made:
 - renamed subscriptions test file to `test-class-wc-subscriptions.php`
 - renamed subscriptions payment count test file to `test-class-wc-subscription-payment-count.php` (Sorry if this comes across as a bit pedantic haha)
 - Moved a helper function we were using in a set to our `PHPUnit Utils`

Let me know your thoughts on removing the email tests for now.
They're passing so I'm fine to leave them as is

tests/unit/test-class-wc-subscriptions.php
```diff
@@ -3,20 +3,20 @@
...
  * Class: WC_Subscriptions_Get_Date_Test
  */
-class WCS_Subscriptions_Test extends WP_UnitTestCase {
+class WC_Subscriptions_Test extends WP_UnitTestCase {
 
 	/** An array of basic subscriptions used to test against */
-	public $subscriptions = [];
+	public static $subscriptions = [];
+
+	public static $yes;
```
brucealdridge:
```suggestion
```

Doesn't look like this is used.

brucealdridge:
Good catch on the slowdown. 
Sounds sensible on the email front as well. That would explain why I was having so much trouble with the other email tests. Flaky tests are never good to keep around, lets remove, then will be good to merge.

mattallan:
I've just pushed up a few commits which:
- removes the Email unit tests
- pulls in the latest trunk changes
- fixes some minor PHPCS issues in the unit tests

This should be good to merge now 🎉

mattallan:
Thanks again @brucealdridge for all the help with migrating over these unit tests 💯 
With all the checks passing and both of our approval on the latest changes I'm going to merge this one :)

PR #86, opened 2022-01-07
Handle relocated templates.

### LINKED ISSUE #85: Handle relocated templates
## Description

Related to 4291-gh-woocommerce/woocommerce-subscriptions.

Template files were moved from woocommerce-subscriptions/templates to woocommerce-subscriptions-core/templates, so 3rd-party plugin that tries to load WCS' templates using `wc_get_template()` will not work.

## Product impact

- [x] Does this feature affect WooCommerce Subscriptions? 4291-gh-woocommerce/woocommerce-subscriptions
- [x] Does this feature affect WooCommerce Payments? https://github.com/Automattic/woocommerce-payments/issues/3606

shendy-a8c:
Fixes #85

## Description

A continuation of 4293-gh-woocommerce/woocommerce-subscriptions#pullrequestreview-845268689.
Hook to `wc_get_template` to handle relocated subscriptions templates from WCS to subscriptions-core.

## How to test this PR

I only have testing instructions for subscriptions-core loaded from WCS. Copy-pasted from 4293-gh-woocommerce/woocommerce-subscriptions:
1. Install All Products for WooCommerce Subscriptions 3.1.31.
2. Go to Admin > WooCommerce > Settings > Subscriptions > enable "Add to Subscription" for Products and Carts.
3. Create a simple product and add subscription plan under Product Data > Subscriptions.
4. Purchase that product with subscription.
5. After purchasing, now that you have subscription for that product, open the the product page again and there will be 'Add to an existing subscription' option. With this PR branch, a table of subscriptions should show up correctly. With base branch, nothing should show up.

Screen Recording 2022-01-06 at 06 06 40

## Product impact

- [x] Will this PR affect WooCommerce Subscriptions? 4291-gh-woocommerce/woocommerce-subscriptions
- [x] Will this PR affect WooCommerce Payments? https://github.com/Automattic/woocommerce-payments/issues/3606

### DIFF
--- a/includes/class-wcs-template-loader.php
+++ b/includes/class-wcs-template-loader.php
@@ -384,7 +384,7 @@ class WCS_Template_Loader {
 			return $template;
 		}
 
-		return $subscriptions_core_path . 'templates/' . self::$relocated_templates[ $template_file ] . $template_file;
+		return $subscriptions_core_path . '/templates/' . self::$relocated_templates[ $template_file ] . $template_file;
 	}
 
 	/**
--- a/includes/class-wcs-template-loader.php
+++ b/includes/class-wcs-template-loader.php
@@ -7,6 +7,72 @@
  */
 class WCS_Template_Loader {
 
+	/**
+	 * Relocated templates from WooCommerce Subscriptions.
+	 *
+	 * @var array[]
+	 */
+	private static $relocated_templates = [
+		'admin/deprecated/html-variation-price.php',
+		'admin/deprecated/html-variation-synchronisation.php',
+		'admin/deprecated/order-shipping-html.php',
+		'admin/deprecated/order-tax-html.php',
+		'admin/html-admin-notice.php',
+		'admin/html-failed-scheduled-action-notice.php',
+		'admin/html-variation-price.php',
+		'admin/html-variation-synchronisation.php',
+		'admin/status.php',
+		'cart/cart-recurring-shipping.php',
+		'checkout/form-change-payment-method.php',
+		'checkout/recurring-coupon-totals.php',
+		'checkout/recurring-fee-totals.php',
+		'checkout/recurring-itemized-tax-totals.php',
+		'checkout/recurring-subscription-totals.php',
+		'checkout/recurring-subtotals.php',
+		'checkout/recurring-tax-totals.php',
+		'checkout/recurring-totals.php',
+		'checkout/subscription-receipt.php',
+		'emails/admin-new-renewal-order.php',
+		'emails/admin-new-switch-order.php',
+		'emails/admin-payment-retry.php',
+		'emails/cancelled-subscription.php',
+		'emails/customer-completed-renewal-order.php',
+		'emails/customer-completed-switch-order.php',
+		'emails/customer-on-hold-renewal-order.php',
+		'emails/customer-payment-retry.php',
+		'emails/customer-processing-renewal-order.php',
+		'emails/customer-renewal-invoice.php',
+		'emails/email-order-details.php',
+		'emails/expired-subscription.php',
+		'emails/on-hold-subscription.php',
+		'emails/plain/admin-new-renewal-order.php',
+		'emails/plain/admin-new-switch-order.php',
+		'emails/plain/admin-payment-retry.php',
+		'emails/plain/cancelled-subscription.php',
+		'emails/plain/customer-completed-renewal-order.php',
+		'emails/plain/customer-completed-switch-order.php',
+		'emails/plain/customer-on-hold-renewal-order.php',
+		'emails/plain/customer-payment-retry.php',
+		'emails/plain/customer-processing-renewal-order.php',
+		'emails/plain/customer-renewal-invoice.php',
+		'emails/plain/email-order-details.php',
+		'emails/plain/expired-subscription.php',
+		'emails/plain/on-hold-subscription.php',
+		'emails/plain/subscription-info.php',
+		'emails/subscription-info.php',
+		'html-modal.php',
+		'myaccount/my-subscriptions.php',
+		'myaccount/related-orders.php',
+		'myaccount/related-subscriptions.php',
+		'myaccount/subscription-details.php',
+		'myaccount/subscription-totals-table.php',
+		'myaccount/subscription-totals.php',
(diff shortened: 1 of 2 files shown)


includes/class-wcs-template-loader.php
```diff
@@ -289,4 +357,53 @@ public static function get_opc_subscription_add_to_cart() {
...
+			return $template;
+		}
+
+		$is_default_path_wcs = false;
+		if ( class_exists( 'WC_Subscriptions_Plugin' ) ) {
+			// WCS 4+ is active.
+			$wcs_templates_path  = WC_Subscriptions_Plugin::instance()->get_plugin_directory( 'templates/' );
+			$is_default_path_wcs = $default_path === $wcs_templates_path;
+		} elseif ( class_exists( 'WC_Subscriptions' ) ) {
+			// WCS 3 or older is active.
```
shendy-a8c: I'm not sure if this block is necessary because there will be no case where subscriptions-core is loaded when WCS 3 or older is active, right?

includes/class-wcs-template-loader.php
```diff
@@ -289,4 +357,53 @@ public static function get_opc_subscription_add_to_cart() {
 		global $product;
 		wc_get_template( 'checkout/add-to-cart/simple.php', array( 'product' => $product ), '', PP_One_Page_Checkout::$template_path );
 	}
+
+	/**
+	 * Handles relocated templates.
+	 *
+	 * Hooked onto 'wc_get_templates'.
+	 *
+	 * @since 1.4.0
```
shendy-a8c, line 366: Is the number for `@since` the subscriptions-core's version or WCS? I assume subscriptions-core's but looking at the other functions above, seems like it refers to WCS' version.

mattallan: Yeah, we should probably use the subscriptions-core version in `@since` for doc-block. Those other functions are copy-pasted from WCS and never updated. 

No huge priority to get those updated at this stage

includes/class-wcs-template-loader.php
mattallan: Yeah that's correct! We don't need this `elseif` for the `class_exists( 'WC_Subscriptions' )` case.

mattallan (commented):
Nice work on this @shendy-a8c. This is proving to be a difficult issue to get right!!

Something I noticed is that in some cases we (and potentially other plugins) call `wc_get_template()` with the `$template_name` including the directory path and other cases we have the directory in the `$default_path` 🙈

For example in `includes/admin/class-wcs-admin-notice.php`

```
$template_name = 'html-admin-notice.php';
$template_path = WC_Subscriptions_Core_Plugin::instance()->get_subscriptions_core_directory( 'templates/admin/' );
```

Given these cases can exist, this means the following conditions won't work as we intend them to:

```
in_array( $template_name, self::$relocated_templates, true )
```

this won't find a template, and

```
$subscriptions_core_templates_path = WC_Subscriptions_Core_Plugin::instance()->get_subscriptions_core_directory( 'templates/' );

if ( $default_path === $subscriptions_core_templates_path ) {
	return $template;
}
```

Maybe we could change the following `===` to use `str_pos` instead?

```
if ( $default_path === $subscriptions_core_templates_path ) {
	return $template;
}
```

As for the relocated template names sometimes being passed as just the file name or sometimes having the full templates/{directory}, I have no idea.
Maybe we can store the template filename only?

includes/class-wcs-template-loader.php
```diff
@@ -289,4 +357,53 @@ public static function get_opc_subscription_add_to_cart() {
...
+			if ( ! file_exists( dirname( $default_path ) . "/$maybe_a_plugin_name.php" ) ) {
+				$is_default_path_wcs = true;
+			}
+		}
+
+		if ( ! $is_default_path_wcs ) {
+			return $template;
+		}
+
+		return wc_get_template(
```
mattallan: Can we return `wc_get_template()` on the filter we're hooked onto?

This filter expects a string however this function returns void if I'm understanding this correctly.

mattallan:
Hey @shendy-a8c I had a look into this and pushed up an alternative to address some of the issues I noticed ([see the changes here](https://github.com/Automattic/woocommerce-subscriptions-core/compare/fix/handle-relocated-templates...handle-relocated-templates-2)):
 - don't return `wc_get_template()` inside the filter, return the new/updated template path instead.
 - use `strpos` instead of `===` in case the default path given includes the directory to the template file
 - returns early if the template file can be found (we don't need to do any relocating if the file exists)
 - I seperated the logic for determining if the file is a default WCS file so we can more easily write unit test this and it cleans up the main function a little bit.

Let me know of your thoughts with those changes and feel free to apply some or any of them to this PR :)
If you like those changes I'm also happy to create a PR merging into this branch.

includes/class-wcs-template-loader.php
shendy-a8c: Good catch! That's a major oversight on my part.

shendy-a8c:
All good call, @mattallan! I added fixes to [your branch](https://github.com/Automattic/woocommerce-subscriptions-core/commits/handle-relocated-templates-2).
How should we proceed with this? Merge your branch to this PR branch? or close this PR and create a new PR of your branch?

mattallan:
> I added fixes to your branch.

Thanks for this!

> How should we proceed with this? Merge your branch to this PR branch? or close this PR and create a new PR of your branch?

I'll cherry-pick those commits onto this PR and then review and test it all as a whole.

mattallan (approved):
Thanks @shendy-a8c for all the work on this!

Here's what I've tested on the latest changes:
 - Frontend templates (mainly myaccount and cart, checkout) are still working
 - Custom subscriptions templates in themes are still working
 - Fetching templates from the old default path are properly redirected

I think this is good to merge from my perspective. Did you have any concerns with the latest changes @shendy-a8c ?

shendy-a8c:
Thank you, @mattallan.

> Did you have any concerns with the latest changes

No. Merging this...

> I seperated the logic for determining if the file is a default WCS file so we can more easily write unit test this and it cleans up the main function a little bit.

This is really a good idea. It's not urgent but so it doesn't fall through the cracks, I just created https://github.com/Automattic/woocommerce-subscriptions-core/issues/88.

PR #99, opened 2022-01-18
v1.6.1 release

mattallan:
```
2022-01-18 - version 1.6.1
* Dev: Update the list of "export-ignore" in `.gitattributes` to include recent developer files. PR#97
* Dev: Set the composer package type to "wordpress-plugin". PR#96
```


mattallan:
Self-merging this to get this version bump in the next WC Payments release.
The changes are only version bumps and a date change so nothing that could possibly break anything 🤞

PR #106, opened 2022-02-02
Fixed missing order placeholder variables in subscription cancelled admin email template.

sun:
## Description

Problem
- When a user cancels a subscription, the email sent to the admin shows empty data related to that subscription.

Details
- The plugin woocommerce-email-customizer-plus expects an order object for it to load the subscription data as in all other woocommerce email templates. This is simply not set by woocommerce-subscriptions.

Proposed solution
1. Add the order object to the email template context variables.

## How to test this PR

1. Install plugin woocommerce-email-customizer-plus.
2. Customize the subscription cancelled admin email notification template, outputting information from the subscription/order.
3. Test the customized email template; the values should appear in the admin email notification.

## Product impact

- [ ] Added changelog entry

### DIFF
--- a/includes/emails/class-wcs-email-cancelled-subscription.php
+++ b/includes/emails/class-wcs-email-cancelled-subscription.php
@@ -98,6 +98,7 @@ class WCS_Email_Cancelled_Subscription extends WC_Email {
 			$this->template_html,
 			array(
 				'subscription'       => $this->object,
+				'order'              => $this->object,
 				'email_heading'      => $this->get_heading(),
 				'additional_content' => is_callable( array( $this, 'get_additional_content' ) ) ? $this->get_additional_content() : '', // WC 3.7 introduced an additional content field for all emails.
 				'sent_to_admin'      => true,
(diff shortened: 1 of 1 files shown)


brucealdridge (commented):
Hi @sun thanks for the PR.
It sounds like your proposed fix is to add/fix missing functionality in another plugin. 

Are you able to explain why it is better to make the change in `woocommerce-subscriptions-core` instead of `woocommerce-email-customizer-plus`?

Would this change also need to be made to `WCS_Email_Expired_Subscription` and `WCS_Email_On_Hold_Subscription` ?

haszari:
Thanks @sun for reaching out. Adding on to @brucealdridge comments, if you have a bug with subscriptions, or a compelling use-case to change hooks or extensibility interfaces, please open an issue with more details:

- What you're trying to do – what's the use case?
- How it works now, i.e. how to reproduce the bug or demonstrate the gap in functionality. Ideally with just WooCommerce and WooCommerce Subscriptions active.
- Your desired outcome – how you want it to work.

mattallan:
This PR has been opened for 2 years without a response and so I suspect by now that the "Email Customizer for WooCommerce" has implemented Subscriptions support.

Closing this. Please re-open if this change is still needed. Thanks

PR #111, opened 2022-02-09
Fix: Changing payment method not filtering the subscription total to be $0 in WC Payments

mattallan:
For #110 

## Description

While investigating https://github.com/Automattic/woocommerce-payments/issues/3765 we discovered that changing the payment method of a WCPay Subscription was incorrectly charging the customer.

This is caused by an issue in `subscriptions-core` where certain filters are attached to the `woocommerce_loaded` hook which has already been triggered before `subscriptions-core` library is loaded.

The fix in this PR is to move the filter outside of the `woocommerce_loaded` hook which was only being used to check if the current WC version was pre 3.0.

Since the minimum WC version we support is 4.4, we can remove the need for the:

```
add_action( 'woocommerce_loaded', __CLASS__ . '::attach_dependant_hooks' );
```

## How to test this PR

1. Make sure Subscriptions with WooCommerce Payments is enabled and the WC Subscriptions plugin is disabled.
2. Purchase a new subscription
3. Go to My Account > Subscriptions > View your new active subscription
4. Next to the list of Actions, click "Change Payment"
5. Complete the change payment method process.
6. Log into the Stripe dashboard and view Payments for this store.
7. Notice there's a charge whenever you change payment method :x:

## Product impact

- [ ] ~~Will this PR affect WooCommerce Subscriptions?~~

https://github.com/Automattic/woocommerce-payments/issues/3768

### DIFF
--- a/includes/class-wc-subscriptions-core-plugin.php
+++ b/includes/class-wc-subscriptions-core-plugin.php
@@ -54,7 +54,7 @@ class WC_Subscriptions_Core_Plugin {
 		if ( $autoloader ) {
 			$this->autoloader = $autoloader;
 		} else {
-			$this->autoloader = new WCS_Core_Autoloader( $this->get_subscriptions_core_directory() );
+			$this->autoloader = new WCS_Core_Autoloader( $this->get_subscriptions_core_directory()  );
 			$this->autoloader->register();
 		}
 
--- a/includes/class-wc-subscriptions-core-plugin.php
+++ b/includes/class-wc-subscriptions-core-plugin.php
@@ -54,7 +54,7 @@ class WC_Subscriptions_Core_Plugin {
 		if ( $autoloader ) {
 			$this->autoloader = $autoloader;
 		} else {
-			$this->autoloader = new WCS_Core_Autoloader( $this->get_subscriptions_core_directory()  );
+			$this->autoloader = new WCS_Core_Autoloader( $this->get_subscriptions_core_directory() );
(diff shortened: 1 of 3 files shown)


includes/class-wc-subscriptions-core-plugin.php
```diff
@@ -54,7 +54,7 @@ public function __construct( $autoloader = null ) {
 		if ( $autoloader ) {
 			$this->autoloader = $autoloader;
 		} else {
-			$this->autoloader = new WCS_Core_Autoloader( $this->get_subscriptions_core_directory() );
+			$this->autoloader = new WCS_Core_Autoloader( $this->get_subscriptions_core_directory()  );
```
aprea: **Nitpick:** Can probably revert this change 😄

mattallan: haha whoops 🤦 
I was testing something the other day with husky and completely forgot about this commit 😂

aprea (approved):
This is working for me ✅

I tested with:

- `woocommerce-payments` on the `develop` branch at commit [65c6d89](https://github.com/Automattic/woocommerce-payments/tree/65c6d89).
- `woocommerce-subscriptions-core` on `for/issue-110-maybe-zero-total`

I observed that the order amount is changed to $0 when processing a request to change the payment method.

I also tested with WooCommerce Subscriptions 4.0.2 enabled and didn't observe any issues.

mattallan:
Thanks @aprea !

PR #132, opened 2022-04-01
Modify the changelog entry header format

aprea:
As discussed here p1648770283229429-slack-C02BW3Z8SHK

This PR modifies the changelog header format to use the WordPress.org format rather than WooCommerce.com format.

There shouldn't have any negative implications as this plugin isn't deployed to either WordPress.org or WooCommerce.com.


shendy-a8c (approved):
Looks good.

I wonder what's the motivation of this change, to use [jetpack-changelogger](https://github.com/Automattic/jetpack-changelogger)?

aprea:
> Looks good.
> 
> I wonder what's the motivation of this change, to use [jetpack-changelogger](https://github.com/Automattic/jetpack-changelogger)?

It would be interesting to see if we could use changelogger in the version, I'm not sure if it supports both formats 🤔

The original motivation was to align us with WCPay so I could reuse this regex to extract changelog entries https://github.com/Automattic/woocommerce-payments/blob/develop/.github/workflows/pre-release-publish-on-tag.yaml#L135

Issue #131, opened 2022-03-30
Create an automation to handle creating GitHub releases


haszari:
## Description
Currently we have a manual process for "releasing" subscriptions core. It goes something like this: 

##### Prepare release PR
- Make a release branch e.g. `release/5.3.0`
- Update version numbers in various files
- Build assets with `npm`
- Push version and built assets to release branch
- Open a PR for release branch - the "release PR"

##### Build and push to GitHub releases
- Merge to `trunk` (after a final eyeball check of the changes)
- Create a GitHub release pointing at current `trunk`, named with raw release version `5.3.0`
- Paste changelog into release description field and publish

From there, that release can be pulled into client projects using composer. 

We'd like to streamline this as much as possible. There's a lot of redundant/mechanical monkey work here that is time-consuming and prone to human error.

The goal of this issue is to automate the "Build and push to GitHub releases" portion of this process, i.e. preparing the GitHub release.

The first part of this process will be tackled in #135.

haszari:
> Automation 2: Build and push to GitHub releases

I'm working on this. These are the steps that are needed:

- [Trigger when specific branch is merged](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-when-a-pull-request-merges) - `release/*` - @brucealdridge 
- Create GitHub release pointing at trunk with version number name - @haszari 
- Add changelog to release description
- Publish release

haszari:
> Create GitHub release pointing at trunk with version number name

There are lots of ways to create a release in a GitHub Actions workflow.

- Manual e.g. use curl or other client to make requests to GitHub API.
- Use an existing action from the community. See this [deprecated GitHub repo for a list](https://github.com/actions/create-release).

I'm going with an action for now, seems best bang for buck, and starting with [`softprops/action-gh-release` which seems very popular and well documented](https://github.com/softprops/action-gh-release).

Working on this in a custom fork (so I can push to trunk to test): https://github.com/haszari/woocommerce-subscriptions-core/blob/trunk/.github/workflows/tag-release.yml

PR #133, opened 2022-04-04
Add GitHub workflow to automate release PRs

### LINKED ISSUE #135: Create an automation to handle creating release pull requests
## Description
Currently we have a manual process for "releasing" subscriptions core. It goes something like this: 

##### Prepare release PR
- Make a release branch e.g. `release/5.3.0`
- Update version numbers in various files
- Build assets with `npm`
- Push version and built assets to release branch
- Open a PR for release branch - the "release PR"

##### Build and push to GitHub releases
- Merge to `trunk` (after a final eyeball check of the changes)
- Create a GitHub release pointing at current `trunk`, named with raw release version `5.3.0`
- Paste changelog into release description field and publish

From there, that release can be pulled into client projects using composer. 

We'd like to streamline this as much as possible. There's a lot of redundant/mechanical monkey work here that is time-consuming and prone to human error.

The goal of this issue is to automate the "Prepare release PR" portion of this process, i.e. preparing the release PR.

The second part of this process will be tackled in #131.

aprea:
Fixes #135

## Description

This PR adds a GitHub workflow to automate release PRs. This workflow uses the "workflow_dispatch" trigger, this means the workflow must be manually triggered via the "Actions" page (see below for an example).

Screen Shot 2022-03-29 at 4 50 18 pm

You can also watch a video demo here → p1648783295601769/1648532936.655589-slack-C02BW3Z8SHK 

The workflow performs the following steps:
- Check out the `woocommerce-subscriptions-core` repository (on the `trunk` branch)
- Check out the `woorelease` repository (the latest version, i.e. `2.2.0`)
- Enable composer dependencies caching
- Enable npm dependencies caching
- Setup PHP with composer
- Install Woorelease composer dependencies
- Create the release branch
- Build `woocommerce-subscriptions-core` assets
- Bumps version strings across the codebase
- Commit changes and create the release PR

## Goals/outcomes

The two goals of this PR are:
- **Time-saving**: The current release process is very manual. Each of the steps listed above must be carried out manually. This takes quite a bit of time, especially if the release lead has some local changes on a feature branch. In comparison, the proposed GitHub workflow takes a little over **1 minute** to run and happens entirely within a CI environment.
- **Reduce human error**: Since these steps are manual, there's always the potential to make small mistakes when preparing a release. These mistakes should be minimized/eliminated when using an automated process like the workflow proposed in this PR. 

## How to test this PR

This PR is tricky to test as the workflow must be present on the default branch (i.e. `trunk`) before you can run it.

I was mostly testing this in a fork I created → https://github.com/aprea/woocommerce-subscriptions-core

You can see the workflow runs here → https://github.com/aprea/woocommerce-subscriptions-core/actions/workflows/create-release-pr.yml

And here is an example of a PR created by the workflow → https://github.com/aprea/woocommerce-subscriptions-core/pull/7

## Product impact

### DIFF
--- a/.github/workflows/create-release-pr.yml
+++ b/.github/workflows/create-release-pr.yml
@@ -20,6 +20,9 @@ jobs:
         with:
           repository: woocommerce/woorelease
           ref: 2.2.0
+          # This token uses the "repo" scope to check out the private Woorelease repo.
+          # This personal access token is owned by the botwoo account - https://github.com/botwoo
+          # Note: GITHUB_TOKEN does not work here.
           token: ${{ secrets.GH_REPO_PAT }}
           path: woorelease
 
--- a/.github/workflows/create-release-pr.yml
+++ b/.github/workflows/create-release-pr.yml
@@ -63,21 +63,16 @@ jobs:
         run: |
           git checkout -b release/$VERSION
 
-      - name: Build assets
-        run: |
-          # Temporarily remove the postinstall script as it runs "composer install".
-          npm set-script postinstall ""
-          npm ci
-
-          # Revert the changes to the postinstall script.
-          git reset --hard
-          npm run build:js
-
       - name: Update version
         env:
           VERSION: ${{ steps.get_version.outputs.VERSION }}
         run: |
           php bin/update-version.php $VERSION
+
+      - name: Build assets
+        run: |
+          npm install
+          npm run build:js
 
       - name: Commit changes and create PR
         env:
--- a/.github/workflows/create-release-pr.yml
+++ b/.github/workflows/create-release-pr.yml
@@ -1,0 +1,95 @@
+name: Create release PR
+
+on: workflow_dispatch
+
+jobs:
+  create-release-pr:
+    runs-on: ubuntu-20.04
+    defaults:
+      run:
+        working-directory: ./main
+    steps:
+      - name: Check out the woocommerce-subscriptions-core repository
+        uses: actions/checkout@v3
+        with:
+          ref: trunk
+          path: main
+
+      - name: Check out the woorelease repository
+        uses: actions/checkout@v3
+        with:
+          repository: woocommerce/woorelease
+          ref: 2.2.0
+          token: ${{ secrets.GH_REPO_PAT }}
+          path: woorelease
+
+      - name: Enable composer dependencies caching
+        uses: actions/cache@v2
+        with:
+          path: ~/.cache/composer/
+          key: ${{ runner.os }}-composer-${{ hashFiles('woorelease/composer.lock') }}
+
+      - name: Enable npm dependencies caching
+        uses: actions/cache@v2
+        with:
+          path: ~/.npm/
+          key: ${{ runner.os }}-npm-${{ hashFiles('main/package-lock.json') }}
+
+      - name: Setup PHP with composer
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '7.4'
+          tools: composer
+          coverage: none
+
+      - name: Install Woorelease composer dependencies
+        run: |
+          cd $GITHUB_WORKSPACE/woorelease
+          composer install --prefer-dist --no-dev
+
+      - name: Get version
+        id: get_version
+        run: |
+          VERSION=$(sed -n '/^= /s/^.*[^0-9]\([0-9]*\.[0-9]*\.[0-9]*\).*$/\1/p' changelog.txt | head -1)
+
+          echo "::set-output name=VERSION::$(echo $VERSION)"
+
+      - name: Create the release branch
+        env:
+          VERSION: ${{ steps.get_version.outputs.VERSION }}
+        run: |
+          git checkout -b release/$VERSION
+
+      - name: Build assets
+        run: |
+          # Temporarily remove the postinstall script as it runs "composer install".
+          npm set-script postinstall ""
+          npm ci
+
+          # Revert the changes to the postinstall script.
+          git reset --hard
+          npm run build:js
+
+      - name: Update version
+        env:
+          VERSION: ${{ steps.get_version.outputs.VERSION }}
+        run: |
+          php bin/update-version.php $VERSION
+
+      - name: Commit changes and create PR
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          VERSION: ${{ steps.get_version.outputs.VERSION }}
+        run: |
+          git config --global user.name 'github-actions[bot]'
+          git config --global user.email 'github-actions[bot]@users.noreply.github.com'
+          git commit -am "Version $VERSION"
+          git push --set-upstream origin release/$VERSION
+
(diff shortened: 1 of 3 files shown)


.github/workflows/create-release-pr.yml
```diff
@@ -0,0 +1,95 @@
...
+        with:
+          ref: trunk
+          path: main
+
+      - name: Check out the woorelease repository
+        uses: actions/checkout@v3
+        with:
+          repository: woocommerce/woorelease
+          ref: 2.2.0
+          token: ${{ secrets.GH_REPO_PAT }}
```
aprea, line 26: TODO: add a comment about the use of this token

haszari:
> This PR is tricky to test as the workflow must be present on the default branch (i.e. trunk) before you can run it.

I _think_ you can [manually trigger and test on the branch (before merge) if you add `workflow_dispatch` trigger](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/). Though maybe that only works if a version of the workflow is on trunk already? Mentioning in case this helps testing @aprea !

haszari:
Linking in related issue: #131

@aprea  - Does this PR cover all or part of that issue? If so can you link it in and also would be a good idea to recap the outcomes / goal in this PR description. Feel free to edit/tweak the issue if it doesn't match your goals for this :)

aprea:
> > This PR is tricky to test as the workflow must be present on the default branch (i.e. trunk) before you can run it.
> 
> I _think_ you can [manually trigger and test on the branch (before merge) if you add `workflow_dispatch` trigger](https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/). Though maybe that only works if a version of the workflow is on trunk already? Mentioning in case this helps testing @aprea !

Unfortunately, you can only run workflows that are already present on `trunk`.

.github/workflows/create-release-pr.yml
```diff
@@ -0,0 +1,93 @@
...
+        env:
+          VERSION: ${{ steps.get_version.outputs.VERSION }}
+        run: |
+          git checkout -b release/$VERSION
+
+      - name: Update version
+        env:
+          VERSION: ${{ steps.get_version.outputs.VERSION }}
+        run: |
+          php bin/update-version.php $VERSION
```
brucealdridge, line 70: I think we should swap these steps around and re-run the get_version step in here. This way we would be operating based on the new version rather than the old.

From my understanding, if the previous version was `1.8.0` and we were therefore wanting to bump to `1.9.0` the branch and commit would mention `1.8.0`

.github/workflows/create-release-pr.yml
```diff
@@ -0,0 +1,93 @@
...
+        uses: actions/checkout@v3
+        with:
+          ref: trunk
+          path: main
+
+      - name: Check out the woorelease repository
+        uses: actions/checkout@v3
+        with:
+          repository: woocommerce/woorelease
+          ref: 2.2.0
```
brucealdridge, line 22: If we're locking woorelease to v2.2, should we cache it? Likely depends on how often this action is run.

brucealdridge (commented):
Looking good @aprea. Just added a couple of comments.

.github/workflows/create-release-pr.yml
aprea: Hey @brucealdridge, I'm not sure I understand your request.

I'll provide some extra context below to ensure we're on the same page.

Let's use your example, i.e. current version is `1.8.0` and the upcoming version is `1.9.0`.

The `changelog.txt` file will contain the following header:

```
= 1.9.0 - 2022-xx-xx =
```

In the "Get version" step we parse the `changelog.txt` file and grab the upcoming version (`1.9.0` in this example). We then set that version as an output parameter so we can access it in subsequent steps.

We use that version in the "Create the release branch" step to create the release branch (e.g. `release/1.9.0`)

We also use it in the "Update version" step to bump the version strings across the codebase (`package.json`, `woocommerce-subscriptions-core.php`, etc.) 

These steps do not reference the current version (i.e. `1.8.0`). The steps only ever need to reference the upcoming version (i.e. `1.9.0`).

Hopefully that makes sense 😅

brucealdridge: Thanks, that makes sense. 

So we need to make sure that we have a version number included in the changelog rather than a placeholder like
```
= 1.x.x - 20xx-xx-xx =
```
I misread the code, or assumed that woorelease was bumping the version. But rather it is just updating the version in the various places to the version that is passed in (and already in the changelog.)

brucealdridge: FYI: https://github.com/Automattic/woocommerce-subscriptions-core/commit/c44ec3ed415bb54c96e47e4ffe8b48605cba0152
It looks like it's happened a couple of times. Will just need to make sure we commit an actual version number before running the action. 
If this happens, the action will either fail or produce a `Version x.x.x` PR that we can just close so not a problem.

.github/workflows/create-release-pr.yml
aprea: I'm not sure if I believe this but GitHub is telling me that this step takes 1 second to execute 😬

Screen Shot 2022-04-05 at 2 47 12 pm

If that's to be believed then we probably don't need caching, maybe it's doing its own internal caching? 🤔

package.json
```diff
@@ -18,7 +18,7 @@
 		"url": "https://github.com/Automattic/woocommerce-subscriptions-core/issues"
 	},
 	"scripts": {
-		"postinstall": "composer install",
+		"postinstall": "is-ci || composer install",
```
shendy-a8c, line 21: I'm not entirely clear why this change is needed. Is there harm `composer install` is run on `npm install`?

shendy-a8c (approved):
Hi :wave:, Being an extra set of eyes here.

I think this PR looks good and definitely this GH action will safe time and give a peace of mind when one is releasing this library.

Without being able to test it, I can only read the diff and simulate what the GH action would do locally as much as possible, for example, running the command to get the version from changelog, run `bin/update-version.php`, etc.

Also, I try to think when things are not in the expected state, for example:
- When changelog entry isn't in the `= 1.9.0 - 2022-xx-xx =` format, or
- When there is already the same release branch (possibly created manually or GH action is run more than once)

The worst case is that the release PR will be created and there is manual validation that needs to be done by human (right?), so I'm not too worried.

This might be already in your radar but I want to point out that once this PR and https://github.com/Automattic/woocommerce-subscriptions-core/pull/136 are merged, we need to:
- Update our handbook (PdjTHR-fl-p2), and
- Set `GH_REPO_PAT`

package.json
aprea: No real harm but it's not necessary as we don't use the `woocommerce-subscriptions-core` composer dependencies in this workflow. I made this change mostly for performance reasons.

aprea:
**Note:** This workflow will not work yet as I haven't created the `GH_REPO_PAT` secret in this repo. I'm still waiting to receive access to the `botwoo` GitHub account.

PR #140, opened 2022-04-08
Add support for unit tests in php8+

### LINKED ISSUE #139: Support for Unit tests on php 8.0+
## Description

Due to limitations with phpunit, our unit tests currently only run on php versions 7.0 - 7.4.

The main reason behind this is php7.0 requires a maximum version of phpunit 6.0. [WooCommerce is bumping the minimum required version for php to 7.2](https://developer.woocommerce.com/2022/01/05/new-requirement-for-woocommerce-6-5-php-7-2/) next month so we can drop support for unit tests along with as WooCommerce Payments did [recently](https://github.com/Automattic/woocommerce-payments/pull/3726)

brucealdridge:
Fixes #139 

## Description

Adds support for unit testing in php8.0 and 8.1. This also removes the ability to run unit tests on php7.0
(compatibility with php7.0 still exists).

PHPUnit has been updated to the latest version which allows us to test against php7.1 - php8.1. Github actions have also been updated to run the test suite against all supported versions.

As part of this PR I encountered issues with the current github actions were testing against the oldest version for WC which was causing issues on newer versions of php. 
So github actions have been updated so we now run unit tests on the latest version of WP and the following versions of WC / php

WC Version | PHP Version | Notes
---|---|---
Latest | 7.1 - 8.1 |
Beta | 7.4 | Any version would do. I went with middle of the road for compatibility broad compatibility
5.5 | 7.1 | This matches the oldest version of WC we currently test

## How to test this PR

Setup unit tests as mentioned in `tests/README.md` and confirm they all pass. Additionally github actions output can be viewed.

- [x] Will this PR affect WooCommerce Subscriptions? no, developer only change.
- [x] Will this PR affect WooCommerce Payments? no, developer only change.


composer.json
```diff
@@ -5,17 +5,14 @@
...
 	},
 	"require-dev": {
-		"phpunit/phpunit": "^6.0",
+		"phpunit/phpunit": "9.5.14",
 		"woocommerce/woocommerce-sniffs": "0.1.0",
 		"dave-liddament/sarb": "^1.1",
-		"yoast/phpunit-polyfills": "^1.0",
-		"phpdocumentor/reflection-docblock": "4.3.4",
-		"phpdocumentor/type-resolver": "0.5.1",
-		"doctrine/instantiator": "1.0.5"
```
brucealdridge, line 18: Not sure why these were here. I'm guessing it was to help lock in the versions required for the older version of phpunit. Subscriptions Core doesn't require these but they are dependencies of phpunit.

includes/class-wc-subscriptions-cart.php
```diff
@@ -220,7 +220,7 @@ public static function set_subscription_prices_for_calculation( $price, $product
 				if ( $trial_length > 0 ) {
 					$price = $sign_up_fee;
 				} else {
-					$price += $sign_up_fee;
+					$price = (float) $price + $sign_up_fee;
```
brucealdridge: Without the explicit casting to float, I was getting unit test errors due to price being a string on newer versions of php. This would have been casted automatically in older versions.

changelog.txt
```diff
@@ -1,5 +1,8 @@
 *** WooCommerce Subscriptions Core Changelog ***
 
+= 1.9.0 - 20xx-xx-xx =
+* Dev: Update phpunit to v9 to allow testing against newer php versions. PR#139
```
```suggestion
* Dev: Update phpunit to v9 to allow testing against newer php versions. PR#140
```

includes/class-wc-subscriptions-cart.php
mattallan: Just making a note that this seems fine to me. I've noticed WC Core also casts prices to `(float)` with inline comments saying it's for PHP 8 compat.

mattallan (approved):
Thanks @brucealdridge for these changes!

> This also removes the ability to run unit tests on php7.0

This seems fine to me, especially since we're now able to test PHP versions from 7.1 all the way to 8.1 👍 

Small suggestion to fix typo in changelog referencing the issue number instead of the PR. Apart from that this is good to me.

Github checks also looks happy 😁

includes/class-wc-subscriptions-cart.php
haszari: Since that `$price` is the return value, is it worth adding type specifier to this function, e.g. `public static float function set_subscription_prices_for_calculation(...)`? And would that help - i.e. reduce the need for explicit cast?

brucealdridge: @haszari we still need the explicit as `$price` maybe a string.
I've spent a bit of time playing around with type hinting in php and can cause lots of issues

```php
	public static function set_subscription_prices_for_calculation( float $price, $product ) {
```
Will work find if we send `0` for example. However sending an empty string will throw a `TypeError`. 
Using a straight cast in this instance is the closest to the original intent.

Issue #157, opened 2022-04-25
Intermittent error with unit tests


brucealdridge:
## Describe the bug
The recent updated to the CI using github actions has surfaced an intermittent issue with the test suite.

in `WC_Subscriptions_Test::test_update_status_to_pending_canellation` we validate that a timestamp is saved in the database. Due to phpunit deprecating and removing the `$delta` parameter on the `assertEquals` function. This code now fails intermittently.

```
There was 1 failure:

1) WC_Subscriptions_Test::test_update_status_to_pending_canellation
Failed asserting that 1649912371 matches expected 1649912372.

/home/runner/work/woocommerce-subscriptions-core/woocommerce-subscriptions-core/tests/unit/test-class-wc-subscriptions.php:1149
```

[View the full logs](https://github.com/Automattic/woocommerce-subscriptions-core/runs/6018810069?check_suite_focus=true)

https://github.com/Automattic/woocommerce-subscriptions-core/blob/8eea2a9bd5e6cafda6655bcda3fb754e2ea7936a/tests/unit/test-class-wc-subscriptions.php#L1146

This will need to be updated to `assertEqualsWithDelta` it must have been missed in 8eea2a9bd5e6cafda6655bcda3fb754e2ea7936a

mattallan:
Closed with [a218e45](https://github.com/Automattic/woocommerce-subscriptions-core/pull/159/commits/a218e4549b1017bce6c82aecfccdc78d03ace860) in #159

PR #159, opened 2022-04-27
Update `update_post_meta()` code where an order ID is passed to use CRUD

### LINKED ISSUE #142: Replace code which uses `update_post_meta()` updating order meta data
## Description

In this issue, we're tackling only the uses of `update_post_meta()` where an order ID is passed. Subscription IDs or Product IDs are fine at this stage.

### How to update code?
These uses should be replaced with:

```
$order->update_meta_data( $meta_key, $meta_value );
$order->save();
```

or using a setter function if it exists:
```
$order->set_total();
$object->save();
```

### What code needs updating?
 - [x] [`wcs_set_objects_property()`](https://github.com/Automattic/woocommerce-subscriptions-core/blob/2a0654bf0385cc3e5b097943aea5589e843b05d0/includes/wcs-compatibility-functions.php#L198-L218)
   - Can use https://github.com/woocommerce/woocommerce-subscriptions/pull/3814/files for inspiration
 - [x] [`save_old_paypal_meta()`](https://github.com/Automattic/woocommerce-subscriptions-core/blob/2a0654bf0385cc3e5b097943aea5589e843b05d0/includes/gateways/paypal/includes/class-wcs-paypal-standard-switcher.php#L169-L170)

### Code/classes to ignore
 - [`maybe_record_subscription_payment()`](https://github.com/Automattic/woocommerce-subscriptions-core/blob/2a0654bf0385cc3e5b097943aea5589e843b05d0/includes/class-wc-subscriptions-renewal-order.php#L95-L96)
   - Surrounded by pre wc 3.0 condition
 - [`add_order_meta`](https://github.com/Automattic/woocommerce-subscriptions-core/blob/2a0654bf0385cc3e5b097943aea5589e843b05d0/includes/class-wc-subscriptions-synchroniser.php#L1498)
   - Code/function deprecated in subscriptions 2.0
 - `WC_Subscription_Legacy`
   - This class is only loaded pre wc 3.0
 - [`generate_renewal_orders()`](https://github.com/Automattic/woocommerce-subscriptions-core/blob/731d98fa36b7037e96e5e5e4f908e091ba68f735/includes/upgrades/class-wc-subscriptions-upgrader.php#L550-L558)
   - This code is in a function that runs when upgrading to WC Subscriptions 1.2 😂  We don't need to touch this code and should probably remove it.
 - `WCS_Repair_2_0_2`
 - `WCS_Upgrade_1_2`
 - 

**Upgrade scripts** can be ignored at this stage as these might be removed from the next version. See internal discussion here: pdjTHR-Y8-p2

## Product impact

- Does this issue affect WooCommerce Subscriptions? yes
- Does this issue affect WooCommerce Payments? yes

## Additional context

Product thread: pdjTHR-Xp-p2

mattallan:
Fixes #142

## Description

When order data is moved to custom tables, updating order meta using `update_post_meta()` will no longer work. This PR updates all cases of `update_post_meta()` where an Order ID is passed - thankfully there wasn't many cases in our codebase 😅 

## How to test this PR

**Testing `wcs_set_objects_property()` changes**

This function is pretty well covered in unit tests, however to test nothing is broken a simple test would be to process a subscription renewal order and make sure the renewal order has the correct meta data.

1. With an active subscription, process a renewal order
2. Confirm the new renewal order has `_billing_address_1` metadata set

**Testing `save_old_paypal_meta()` changes**

This one was a little trickier to test so I just hacked some of the conditions because the only changes we cared about is the order updating post meta:
1. Edit the `save_old_paypal_meta()` function to pass if conditions using, `true ||`
2. Manually set `$subscriptions` to be `$subscriptions = [ 101823 => wcs_get_subscription( 101823 ) ];`
3. Run the following snippet:

```
add_action( 'admin_footer', function() {
	update_post_meta( 101823, '_paypal_subscription_id', 'S-fake1234' );
	update_post_meta( 101823, '_payment_method', 'paypal' );

	WCS_PayPal_Standard_Switcher::save_old_paypal_meta( 101824, [] );
} );
```

4. Confirm the $order meta is set in the DB

image

## Product impact

### DIFF
--- a/includes/gateways/paypal/includes/class-wcs-paypal-standard-switcher.php
+++ b/includes/gateways/paypal/includes/class-wcs-paypal-standard-switcher.php
@@ -157,22 +157,24 @@ class WCS_PayPal_Standard_Switcher {
 	public static function save_old_paypal_meta( $order_id, $posted ) {
 		$order = wc_get_order( $order_id );
 
-		if ( wcs_is_order( $order ) && wcs_order_contains_switch( $order ) ) {
-			$subscriptions = wcs_get_subscriptions_for_order( $order, array( 'order_type' => 'switch' ) );
+		if ( ! wcs_is_order( $order ) || ! wcs_order_contains_switch( $order ) ) {
+			return;
+		}
 
-			foreach ( $subscriptions as $subscription ) {
+		$subscriptions = wcs_get_subscriptions_for_order( $order, array( 'order_type' => 'switch' ) );
 
-				if ( 'paypal' === $subscription->get_payment_method() ) {
+		foreach ( $subscriptions as $subscription ) {
 
-					$paypal_id = wcs_get_paypal_id( $subscription->get_id() );
+			if ( 'paypal' === $subscription->get_payment_method() ) {
 
-					if ( ! wcs_is_paypal_profile_a( $paypal_id, 'billing_agreement' ) ) {
-						$order->update_meta_data( '_old_payment_method', 'paypal_standard' );
-						$order->update_meta_data( '_old_paypal_subscription_id', $paypal_id );
-						$order->save();
+				$paypal_id = wcs_get_paypal_id( $subscription->get_id() );
 
-						update_post_meta( $subscription->get_id(), '_switched_paypal_subscription_id', $paypal_id );
-					}
+				if ( ! wcs_is_paypal_profile_a( $paypal_id, 'billing_agreement' ) ) {
+					$order->update_meta_data( '_old_payment_method', 'paypal_standard' );
+					$order->update_meta_data( '_old_paypal_subscription_id', $paypal_id );
+					$order->save();
+
+					update_post_meta( $subscription->get_id(), '_switched_paypal_subscription_id', $paypal_id );
 				}
 			}
 		}
--- a/includes/gateways/paypal/includes/class-wcs-paypal-standard-switcher.php
+++ b/includes/gateways/paypal/includes/class-wcs-paypal-standard-switcher.php
@@ -155,9 +155,10 @@ class WCS_PayPal_Standard_Switcher {
 	 * @since 2.0.15
 	 */
 	public static function save_old_paypal_meta( $order_id, $posted ) {
+		$order = wc_get_order( $order_id );
 
-		if ( wcs_order_contains_switch( $order_id ) ) {
-			$subscriptions = wcs_get_subscriptions_for_order( $order_id, array( 'order_type' => 'switch' ) );
+		if ( wcs_is_order( $order ) && wcs_order_contains_switch( $order ) ) {
+			$subscriptions = wcs_get_subscriptions_for_order( $order, array( 'order_type' => 'switch' ) );
 
 			foreach ( $subscriptions as $subscription ) {
 
@@ -166,8 +167,10 @@ class WCS_PayPal_Standard_Switcher {
 					$paypal_id = wcs_get_paypal_id( $subscription->get_id() );
 
 					if ( ! wcs_is_paypal_profile_a( $paypal_id, 'billing_agreement' ) ) {
-						update_post_meta( $order_id, '_old_payment_method', 'paypal_standard' );
-						update_post_meta( $order_id, '_old_paypal_subscription_id', $paypal_id );
+						$order->update_meta_data( '_old_payment_method', 'paypal_standard' );
+						$order->update_meta_data( '_old_paypal_subscription_id', $paypal_id );
+						$order->save();
+
 						update_post_meta( $subscription->get_id(), '_switched_paypal_subscription_id', $paypal_id );
 					}
 				}
--- a/tests/unit/test-class-wc-subscriptions.php
+++ b/tests/unit/test-class-wc-subscriptions.php
@@ -1137,13 +1137,13 @@ class WC_Subscriptions_Test extends WP_UnitTestCase {
 			if ( in_array( $status, $expected_to_pass, true ) ) {
 
 				try {
-
-					$start_date = gmdate( 'Y-m-d H:i:s', strtotime( '-1 month' ) );
+					$expected_end_date = gmdate( 'Y-m-d H:i:s' );
+					$start_date        = gmdate( 'Y-m-d H:i:s', strtotime( $expected_end_date . ' -1 month' ) );
 					$subscription->update_dates( [ 'start' => $start_date ] );
 
 					$subscription->update_status( 'pending-cancel' );
 
-					$this->assertEquals( time(), $subscription->get_time( 'end' ), '', 2 );
+					$this->assertEquals( strtotime( $expected_end_date ), $subscription->get_time( 'end' ), '', 2 );
 				} catch ( Exception $e ) {
 
 					$this->fail( $e->getMessage() );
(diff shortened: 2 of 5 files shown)


brucealdridge:
@mattallan The new unit test you added is currently failing.

mattallan:
> The new unit test you added is currently failing.

Whoops! Thanks for the heads up! I copied the unit test changes from https://github.com/woocommerce/woocommerce-subscriptions/pull/3814/files but didn't realise some of the unit test changes were for `wcs_get_object_property()` which isn't changed in this PR.

I've removed the expected deprecated unit test but kept the test additions.

@brucealdridge this should be good to review again. Thanks

mattallan:
One of the existing tests is failing with the unexpected calculated date comparison (off by 1 second). I'll try run the test again but this failing test shouldn't block this PR

mattallan:
I just pushed up [3d1e6f0](https://github.com/Automattic/woocommerce-subscriptions-core/pull/159/commits/3d1e6f087345ed79bdd458a6ece25e43bd31b730) which should help improve test flakiness 🤞

brucealdridge:
> I just pushed up [3d1e6f0](https://github.com/Automattic/woocommerce-subscriptions-core/pull/159/commits/3d1e6f087345ed79bdd458a6ece25e43bd31b730) which should help improve test flakiness 🤞

This is issue #157  
I did some research on it and it looks like `assertEquals` has removed the delta parameter in the newer version of phpunit so we will need to use `assertEqualsWithDelta` instead.

mattallan:
> This is issue https://github.com/Automattic/woocommerce-subscriptions-core/issues/157

Oh nice! Apologies I didn't see an issue was already created for this!

I'm happy to take it off this PR and handle it separately to avoid confusion that said, I kinda have an issue with using `assertEqualsWithDelta()` where it can be avoided.

In this instance, the issue with this test sometimes failing and sometimes passing is because we're setting up a subscription's start date with `time() -1 month` and then we're calling a bunch of subscription operations (updating dates and status) and then testing that the end date is `assertEqual` to `time()`.

Depending on how slow the test environment is and how much we set as a delta (i.e. 1-2 seconds), we could still this test failing in the future unless we blow out the delta to 5-10 seconds.

The solution I went with was to use a set/stored version `time()` variable to assert expected date calculations, rather than assuming the operations made during the test will be completed within a given delta 😄

includes/gateways/paypal/includes/class-wcs-paypal-standard-switcher.php
```diff
@@ -155,9 +155,10 @@ public static function maybe_set_payment_method( $payment_processing_result, $or
 	 * @since 2.0.15
 	 */
 	public static function save_old_paypal_meta( $order_id, $posted ) {
+		$order = wc_get_order( $order_id );
 
-		if ( wcs_order_contains_switch( $order_id ) ) {
-			$subscriptions = wcs_get_subscriptions_for_order( $order_id, array( 'order_type' => 'switch' ) );
+		if ( wcs_is_order( $order ) && wcs_order_contains_switch( $order ) ) {
```
brucealdridge: I'd be tempted to change this condition to do a early return and remove a level of indentation from the method 
```suggestion
		if ( ! wcs_is_order( $order ) || ! wcs_order_contains_switch( $order ) ) {
		    return;
		}
```

tests/unit/test-wcs-compatibility-functions.php
```diff
@@ -23,12 +23,18 @@ public function test_wcs_get_objects_property() {
 		$this->assertEquals( get_post( $simple_product->get_id() ), wcs_get_objects_property( $simple_product, 'post' ) );
 		$this->assertEquals( get_post( $variation_product->get_parent_id() ), wcs_get_objects_property( $variation_product, 'post' ) );
 
-		// post_status.
+		// post_status - now deprecated as a property
```
brucealdridge: If these are deprecated can we remove them from the unit tests?
Or move to another function and mark as deprecated?

tests/unit/test-class-wc-subscriptions.php
```diff
@@ -1137,13 +1137,13 @@ public function test_update_status_to_pending_canellation() {
 			if ( in_array( $status, $expected_to_pass, true ) ) {
 
 				try {
-
-					$start_date = gmdate( 'Y-m-d H:i:s', strtotime( '-1 month' ) );
+					$expected_end_date = gmdate( 'Y-m-d H:i:s' );
+					$start_date        = gmdate( 'Y-m-d H:i:s', strtotime( $expected_end_date . ' -1 month' ) );
```
brucealdridge: I've tested this by adding a sleep(1); between lines 1141-1142 and 1142-1143 and in both cases they fail

brucealdridge (approved):
Just a few minor nitpicks

tests/unit/test-class-wc-subscriptions.php
mattallan: I see 🙈 Putting sleeps in the test still causes the same issues that I was trying to prevent.

Because the subscription doesn't have a next payment date, the new calculated end date is being set to the current time, therefore putting sleeps between these lines causes the expected end date I saved to be in the past and an end date can't be in the past, so it still fails.

I had a solution to this by using dates in the future and setting a next payment date on the subscription (which would be the new expected end date), but I think the original purpose of this test is to make sure any end date is set as a result of updating the subscription status to pending-cancel (not validating the end date per se). My approach would slow the test down by updating more dates etc so I'll revert back to using withDelta and go back to the original code 😅

mattallan: For reference, here's the updated test I had:

```
$start_time        = time();
$expected_end_time = strtotime( '+1 month', $start_time );

$subscription->update_dates( [
	'start'        => gmdate( 'Y-m-d H:i:s', $start_time ),
	'next_payment' => gmdate( 'Y-m-d H:i:s', $expected_end_time ),
] );

$subscription->update_status( 'pending-cancel' );

$this->assertEquals( $expected_end_time, $subscription->get_time( 'end' ) );
```

includes/gateways/paypal/includes/class-wcs-paypal-standard-switcher.php
mattallan: Fixed in 886238ca. Thanks for the suggestion!

tests/unit/test-class-wc-subscriptions.php
mattallan: Fixed in [a218e45](https://github.com/Automattic/woocommerce-subscriptions-core/pull/159/commits/a218e4549b1017bce6c82aecfccdc78d03ace860)

tests/unit/test-wcs-compatibility-functions.php
mattallan: This param isn't deprecated (copy/paste fail from original PR made to WC Subscriptions). I've removed this wording from the comment to avoid confusion: [4c72ed4](https://github.com/Automattic/woocommerce-subscriptions-core/pull/159/commits/4c72ed41d623c4ed5f43f3980e845b24b07f0f6e)

mattallan:
Thanks for the review @brucealdridge . I've addressed the issues you've raised. Since you've approved this PR i'll merge it when everything is green :+1:

Fixes #157
^ is fixed by [a218e45](https://github.com/Automattic/woocommerce-subscriptions-core/pull/159/commits/a218e4549b1017bce6c82aecfccdc78d03ace860)

Issue #152, opened 2022-04-21
EPIC: Use CRUD APIs when interacting with Order objects


mattallan:
## Description

Custom Order Tables are coming to WooCommerce sometime this year and with that, any code we have that is fetching order data from posts or postmeta will no longer work as intended.

This epic is to keep track of the progress with updating our Subscriptions Core library to remove usages of the following functions when trying to fetch/update orders:
 - `get_post_meta()` - issue opened
 - `update_post_meta()` - issue opened
 - `get_posts()` - issue opened
 - `add_post_meta()` - no affected code found
 - `get_post_type()` - issue opened
 - `get_post_type_object()` - no affected code found

## Issues

- [x] https://github.com/Automattic/woocommerce-subscriptions-core/issues/144
- [x] https://github.com/Automattic/woocommerce-subscriptions-core/issues/142
- [x] https://github.com/Automattic/woocommerce-subscriptions-core/issues/141
- [x] https://github.com/Automattic/woocommerce-subscriptions-core/issues/153

## Additional context

This issue is part of internal project thread: pdjTHR-Xp-p2

brucealdridge:
marking as closed as all sub-tasks are now complete.

PR #161, opened 2022-04-28
Replace `get_posts()` with `wc_get_orders()` in WC_Subscriptions_Order::get_users_subscription_orders()

### LINKED ISSUE #144: Replace code which uses `get_posts()` to get Orders post types with `wc_get_orders()`
## Description

In this issue we're tackling uses of `get_posts()` where `shop_order` post type is passed. Subscriptions and products are fine as they're staying in the wp posts table.

## How to update code?

We should be able to replace most cases of `get_posts` with `wc_get_orders()`. To search for custom parameters we can use the approach outlined here: https://github.com/woocommerce/woocommerce/wiki/wc_get_orders-and-WC_Order_Query#adding-custom-parameter-support

## Changes needed

- [x] [`get_users_subscription_orders`](https://github.com/Automattic/woocommerce-subscriptions-core/blob/97c279bdb6131462de9e90161a6a172ce0447293/includes/class-wc-subscriptions-order.php#L617)
  - This function is not being used by any of our code including in WC Payments and WC Subscriptions so updating this is very low priority/impact
 - [x] [`wcs_get_subscription_orders`](https://github.com/Automattic/woocommerce-subscriptions-core/blob/2a0654bf0385cc3e5b097943aea5589e843b05d0/includes/wcs-order-functions.php#L425)

## Code to ignore

- `class WCS_Related_Order_Store_CPT` - will be handled separately in #145 
- [`WCS_PayPal_Standard_IPN_Handler::get_order_id_and_key()`](https://github.com/Automattic/woocommerce-subscriptions-core/blob/2a0654bf0385cc3e5b097943aea5589e843b05d0/includes/gateways/paypal/includes/class-wcs-paypal-standard-ipn-handler.php#L604)
  - This can probably be ignored for now because `'shop_order'` post type is never queried (even though it's the default option when calling the function). I believe this default option was pre-WC subscriptions 2.0 before a subscription post type existed and we had to use the parent order ID.
- `class WCS_Repair_2_0` - We're going to disable old repair scripts when HPOS is enabled
- `class WCS_Upgrade_2_0` - We're going to disable old upgrade scripts when HPOS is enabled

## Product impact

shendy-a8c:
Fixes #144 partially.

## Description

This PR only replaces `get_posts()` call in [get_users_subscription_orders()](https://github.com/Automattic/woocommerce-subscriptions-core/blob/97c279bdb6131462de9e90161a6a172ce0447293/includes/class-wc-subscriptions-order.php#L617) with `wc_get_orders()`.

Some key points:
- I follow [this guide](https://github.com/woocommerce/woocommerce/wiki/wc_get_orders-and-WC_Order_Query#adding-custom-parameter-support) to pass the `meta_query` bit.
- I change the field name following [this map](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/wc-order-functions.php#L27-L36).
- I use `customer_id` [parameter](https://github.com/woocommerce/woocommerce/wiki/wc_get_orders-and-WC_Order_Query#customer), instead of passing it in `meta_query`.

## How to test this PR

1. Install `woocommerce-subscriptions-core` as a plugin.
2. Since I [get_users_subscription_orders()](https://github.com/Automattic/woocommerce-subscriptions-core/blob/97c279bdb6131462de9e90161a6a172ce0447293/includes/class-wc-subscriptions-order.php#L617) isn't called anywhere, I use [wp-console](https://wordpress.org/plugins/wp-console/) to call it.

Screen Shot 2022-04-28 at 22 52 10

Actually, I wondered if we can just delete `get_users_subscription_orders()` but then I think someone out there (another plugin or custom code) might depend on it.

3. Run `WC_Subscriptions_Order::get_users_subscription_orders();` and make sure base branch and PR branch returns the same output.

4. I also add this code as a [mu-plugin](https://wordpress.org/support/article/must-use-plugins/), in a file `wp-content/mu-plugins/something.php`.

```
define('SAVEQUERIES', true);
add_action( 'shutdown', function() {
    global $wpdb;
    foreach($wpdb->queries as $q) {
        file_put_contents('/tmp/debug.log', "-----start\n" . $q[0] . "\n-----end\n");
    }
} );
```

To log queries going to the database. I search for `subscription_renewal`. I format the query and diff the query made using base branch vs PR branch.

There is some diff but looks OK:

Screen Shot 2022-04-28 at 23 07 01

## Product impact

### DIFF
--- a/includes/class-wc-subscriptions-order.php
+++ b/includes/class-wc-subscriptions-order.php
@@ -622,35 +622,39 @@ class WC_Subscriptions_Order {
 		}
 
 		// Get all the customers orders which are not subscription renewal orders
-		$order_ids = get_posts( array(
-			'posts_per_page' => 1,
-			'post_type'      => 'shop_order',
-			'post_status'    => 'any',
-			'fields'         => 'ids',
-			'orderby'        => 'date',
-			'order'          => 'DESC',
-			'meta_query'     => array(
-				array(
-					'key'     => '_customer_user',
-					'compare' => '=',
-					'value'   => $user_id,
-					'type'    => 'numeric',
-				),
-				array(
+		$custom_query_var_handler = function( $query, $query_vars ) {
+			if ( ! empty( $query_vars['_non_subscription_renewal'] ) ) {
+				$query['meta_query'][] = array(
 					'key'     => '_subscription_renewal',
 					'compare' => 'NOT EXISTS',
-				),
-			),
-		) );
+				);
+			}
 
-		foreach ( $order_ids as $index => $order_id ) {
-			if ( ! wcs_order_contains_subscription( $order_id, 'parent' ) ) {
-				unset( $order_ids[ $index ] );
+			return $query;
+		};
+		add_filter( 'woocommerce_order_data_store_cpt_get_orders_query', $custom_query_var_handler, 10, 2 );
+
+		$args = array(
+			'limit'    => 1,
+			'type'     => 'shop_order',
+			'status'   => 'any',
+			'orderby'  => 'date',
+			'order'    => 'DESC',
+			'customer' => $user_id,
+		);
+
+		$args['_non_subscription_renewal'] = true;
+
+		$orders = wc_get_orders( $args );
+
+		remove_filter( 'woocommerce_order_data_store_cpt_get_orders_query', $custom_query_var_handler, 10 );
+
+		$order_ids = [];
+		foreach ( $orders as $order ) {
+			if ( wcs_order_contains_subscription( $order->id, 'parent' ) ) {
+				$order_ids[] = $order->id;
 			}
 		}
-
-		// Normalise array keys
-		$order_ids = array_values( $order_ids );
 
 		return apply_filters( 'users_subscription_orders', $order_ids, $user_id );
 	}
--- a/includes/class-wc-subscriptions-order.php
+++ b/includes/class-wc-subscriptions-order.php
@@ -644,20 +644,23 @@ class WC_Subscriptions_Order {
 			'orderby'     => 'date',
 			'order'       => 'DESC',
 			'customer_id' => $user_id,
+			'return'      => 'ids',
 		);
 
 		$args['_non_subscription_renewal'] = true;
 
-		$orders = wc_get_orders( $args );
+		$order_ids = wc_get_orders( $args );
 
 		remove_filter( 'woocommerce_order_data_store_cpt_get_orders_query', $custom_query_var_handler, 10 );
 
-		$order_ids = [];
-		foreach ( $orders as $order ) {
-			if ( wcs_order_contains_subscription( $order->id, 'parent' ) ) {
-				$order_ids[] = $order->id;
+		foreach ( $order_ids as $index => $order_id ) {
+			if ( ! wcs_order_contains_subscription( $order_id, 'parent' ) ) {
+				unset( $order_ids[ $index ] );
 			}
 		}
+
+		// Normalise array keys
+		$order_ids = array_values( $order_ids );
 
 		return apply_filters( 'users_subscription_orders', $order_ids, $user_id );
 	}
--- a/includes/class-wc-subscriptions-order.php
+++ b/includes/class-wc-subscriptions-order.php
@@ -628,6 +628,7 @@ class WC_Subscriptions_Order {
 					'key'     => '_subscription_renewal',
 					'compare' => 'NOT EXISTS',
 				);
+				unset( $query_vars['_non_subscription_renewal'] );
 			}
 
 			return $query;
--- a/includes/class-wc-subscriptions-order.php
+++ b/includes/class-wc-subscriptions-order.php
@@ -635,12 +635,12 @@ class WC_Subscriptions_Order {
 		add_filter( 'woocommerce_order_data_store_cpt_get_orders_query', $custom_query_var_handler, 10, 2 );
 
 		$args = array(
-			'limit'    => 1,
-			'type'     => 'shop_order',
-			'status'   => 'any',
-			'orderby'  => 'date',
-			'order'    => 'DESC',
-			'customer' => $user_id,
+			'limit'       => 1,
+			'type'        => 'shop_order',
+			'status'      => 'any',
+			'orderby'     => 'date',
+			'order'       => 'DESC',
+			'customer_id' => $user_id,
 		);
 
 		$args['_non_subscription_renewal'] = true;
--- a/includes/class-wc-subscriptions-order.php
+++ b/includes/class-wc-subscriptions-order.php
@@ -638,7 +638,6 @@ class WC_Subscriptions_Order {
 		$all_possible_statuses = array_values( array_unique( array_keys( wc_get_order_statuses() ) ) );
 
 		$args = array(
-			'limit'       => 1,
 			'type'        => 'shop_order',
 			'status'      => $all_possible_statuses,
 			'orderby'     => 'date',
--- a/includes/class-wc-subscriptions-order.php
+++ b/includes/class-wc-subscriptions-order.php
@@ -635,10 +635,12 @@ class WC_Subscriptions_Order {
 		};
 		add_filter( 'woocommerce_order_data_store_cpt_get_orders_query', $custom_query_var_handler, 10, 2 );
 
+		$all_possible_statuses = array_values( array_unique( array_keys( wc_get_order_statuses() ) ) );
+
 		$args = array(
 			'limit'       => 1,
 			'type'        => 'shop_order',
-			'status'      => 'any',
+			'status'      => $all_possible_statuses,
 			'orderby'     => 'date',
 			'order'       => 'DESC',
 			'customer_id' => $user_id,


shendy-a8c:
@mattallan @brucealdridge how do we handle changelog for this project? I can think of summarizing into 1 changelog entry for several PRs isn't a bad idea. I'm fine with doing 1 entry for each bit of work too though.

includes/class-wc-subscriptions-order.php
```diff
@@ -622,36 +622,40 @@ public static function get_users_subscription_orders( $user_id = 0 ) {
...
+		$args = array(
+			'limit'       => 1,
+			'type'        => 'shop_order',
+			'status'      => 'any',
+			'orderby'     => 'date',
+			'order'       => 'DESC',
+			'customer_id' => $user_id,
+		);
+
+		$args['_non_subscription_renewal'] = true;
```
shendy-a8c, line 649: I use a unique string `_non_subscription_renewal` here, so I can check for it in `$custom_query_var_handler`.

shendy-a8c:
I have a doubt about what minimum version of PHP that supports assigning a callable to a variable since in my change I use that feature. I'm still trying to find that out.

Jinksi:
> what minimum version of PHP that supports assigning a callable to a variable

@shendy-a8c I think you can safely use this.

[Anonymous function → variable assignment require PHP >= 5.3.0](https://www.php.net/manual/en/functions.anonymous.php). `subscriptions-core` currently supports WP >= 5.6 which requires PHP 5.6

Edit: on the other hand, I don't think this syntax is used anywhere else in the codebase. Is there a benefit to using the variable assigned function syntax?

includes/class-wc-subscriptions-order.php
mattallan: Makes sense!

I wonder if we're going to see a number of duplicated meta query filters being used for the same filter. If we do start seeing this, I curious to know what we could do to prevent duplicated code.

Something like a new file `wcs-order-meta-query-filters.php` which has a bunch of public functions like `wcs_filter_orders_non_subscription_renewal()` so we can call them from anywhere.

I don't think we should implement this just yet but just thinking out aloud while reviewing the changes here :)

Nice one!

mattallan:
> how do we handle changelog for this project?

Good question! I think 1 changelog entry for the epic (https://github.com/Automattic/woocommerce-subscriptions-core/issues/152) that this issue apart of is fine! For the other issues my vote would be for them to have their own changelog entry just so it's easy to find individual changes with a separate linked PR for each line

includes/class-wc-subscriptions-order.php
shendy-a8c: That's a good point.

I can't think of any other way. Seems like we will do a lot of:
```
add_filter( 'woocommerce_order_data_store_cpt_get_orders_query', $callable );
wc_get_orders( $args );
remove_filter( 'woocommerce_order_data_store_cpt_get_orders_query', $callable );
```

We can have a 'central' place for the callables and just add filter once, per your suggestion with `wcs-order-meta-query-filters.php` (if I understand it correctly) but the downside is that part of the code will not be 'close' to the relevant code that calls `wc_get_orders()`.

shendy-a8c:
@Jinksi thank you for pointing that out.

> Is there a benefit to using the variable assigned function syntax?

No particularly. I did so to increase readability.

Function assigned to a variable is already there in this project, for example https://github.com/Automattic/woocommerce-subscriptions-core/blob/trunk/tests/unit/test-class-wc-subscriptions-payment-count.php#L59-L61 but it's just for testing.

Jinksi:
> I did so to increase readability.

@shendy-a8c Nice, yeah I find this to be much easier to follow than callable strings normally used in WP hooks.

shendy-a8c:
Added more changes:
- [53997d0](https://github.com/Automattic/woocommerce-subscriptions-core/pull/161/commits/53997d0c11bafbd3762393d0e7abd5f48483277c) to make sure we're not appending meta query more than once.
- [c47664c](https://github.com/Automattic/woocommerce-subscriptions-core/pull/161/commits/c47664c84ca2bd28541207044a347dc200f5a1c5) to follow `wc_get_orders() docs`, ie value should be an array of string and it doesn't mention `any`. `any` works for now but since it's not mentioned in the docs, it might not work later on.

The diff of the query becomes like this after the above change:
Screen Shot 2022-05-05 at 04 11 12

includes/class-wc-subscriptions-order.php
brucealdridge: If this is a repeated pattern we should look at how to remove the repetition. This is likely outside the scope of the current work but would aid in readability and would simplify code.
```php
function filter_get_orders( $args , $callable ) {
    add_filter( 'woocommerce_order_data_store_cpt_get_orders_query', $callable ); 
    $orders = wc_get_orders( $args ); 
    remove_filter( 'woocommerce_order_data_store_cpt_get_orders_query', $callable );
    return $orders;
}

$orders = filter_get_orders(
    $args,
    function ( $query ) {
       // alter query here
    }
);
```

shendy-a8c: @brucealdridge after giving it a thought, though your suggestion is a good idea, I'm afraid this repetition won't come up again and in that case, having a helper function as suggested would be over-engineering. We can address it later as needed.

shendy-a8c:
@Jinksi @mattallan @brucealdridge is there any other feedback to this PR?

includes/class-wc-subscriptions-order.php
```diff
@@ -622,36 +622,43 @@ public static function get_users_subscription_orders( $user_id = 0 ) {
...
+			return $query;
+		};
+		add_filter( 'woocommerce_order_data_store_cpt_get_orders_query', $custom_query_var_handler, 10, 2 );
 
-		foreach ( $order_ids as $index => $order_id ) {
-			if ( ! wcs_order_contains_subscription( $order_id, 'parent' ) ) {
-				unset( $order_ids[ $index ] );
+		$all_possible_statuses = array_values( array_unique( array_keys( wc_get_order_statuses() ) ) );
+
+		$args = array(
```
mattallan, line 640: In these args, can we add `'return' => 'ids',` since we don't need order objects to be returned?

mattallan (approved):
Thanks @shendy-a8c! I left a suggestion [here](https://github.com/Automattic/woocommerce-subscriptions-core/pull/161/files#r869827924) but apart from that I think this is good to be merged once  addressed.

Approving!

Jinksi (approved):
LGTM Shendy! 🚀

Side-note: thanks for the tips in your testing instructions, I learned some great debugging techniques!

includes/class-wc-subscriptions-order.php
shendy-a8c: Thank you for pointing that out, @mattallan! Addressed in [cec2cb9](https://github.com/Automattic/woocommerce-subscriptions-core/pull/161/commits/cec2cb9acfbe7234b2fa760b87700c85c2836fe7).

shendy-a8c:
After [addressing @mattallan's suggestion](https://github.com/Automattic/woocommerce-subscriptions-core/pull/161#discussion_r870100630), I tested again and found strange that last time I tested it, that function returned an id and now it returned an empty array. Even in `trunk`, that function returns empty.

That is because of the query is limiting to only 1 result and since my last order in my dev store is not a subscription order, it returns an empty array. Reading the comment of the function, I don't think that it should behave like that, so I'm taking out that limit here [e9bb9c7](https://github.com/Automattic/woocommerce-subscriptions-core/pull/161/commits/e9bb9c7db920b9d2c31d63c363392a5977ad9470). That changes the behavior of the original function but risk is minimal here since that function isn't called anywhere at least from WCS or WCPay.

includes/class-wc-subscriptions-order.php
```diff
@@ -622,26 +622,35 @@ public static function get_users_subscription_orders( $user_id = 0 ) {
...
-		) );
+				);
+				unset( $query_vars['_non_subscription_renewal'] );
+			}
+
+			return $query;
+		};
+		add_filter( 'woocommerce_order_data_store_cpt_get_orders_query', $custom_query_var_handler, 10, 2 );
+
+		$all_possible_statuses = array_values( array_unique( array_keys( wc_get_order_statuses() ) ) );
```
saas786, line 638: Shouldn't it be:
```suggestion
		$all_possible_statuses = array_unique( array_keys( wc_get_order_statuses() ) );
```
To me `array_values` seems redundant here...

shendy-a8c:
PHP 7.1 linting check [fails](https://github.com/Automattic/woocommerce-subscriptions-core/runs/6385651745?check_suite_focus=true#step:5:177) because WC 6.5 was just released yesterday that requires PHP 7.2.

Address a fix [here](https://github.com/Automattic/woocommerce-subscriptions-core/pull/165). Merging this PR for now.

includes/class-wc-subscriptions-order.php
shendy-a8c: Sorry I missed your comment before merging this PR. Thank you for suggestion.

However, I think `array_values()` is necessary to avoid non-incremental array index.

Consider this:
```php
<?php
$a = ['a', 'b', 'c', 'a', 'd'];
var_dump(array_unique($a));
var_dump(array_values(array_unique($a)));
```

Output is:
```
array(4) {
  [0]=>
  string(1) "a"
  [1]=>
  string(1) "b"
  [2]=>
  string(1) "c"
  [4]=>
  string(1) "d"
}
array(4) {
  [0]=>
  string(1) "a"
  [1]=>
  string(1) "b"
  [2]=>
  string(1) "c"
  [3]=>
  string(1) "d"
}
```

Notice the first output array is has no array index 3.

saas786: Why do we care about indexes? keys / indexes are irrelevant in this context.

Ref: https://github.com/woocommerce/woocommerce/blob/6.5.0/plugins/woocommerce/includes/data-stores/class-wc-order-data-store-cpt.php#L782

saas786: After some exploration, I think even `array_unique` is irrelevant...

```
		$all_possible_statuses = array_values( array_unique( array_keys( wc_get_order_statuses() ) ) );

		$args = array(
			'type'        => 'shop_order',
			'status'      => $all_possible_statuses,
			'orderby'     => 'date',
			'order'       => 'DESC',
			'customer_id' => $user_id,
			'return'      => 'ids',
		);
```

could have been
```
		$args = array(
			'type'        => 'shop_order',
			'status'      => array_keys( wc_get_order_statuses() ),
			'orderby'     => 'date',
			'order'       => 'DESC',
			'customer_id' => $user_id,
			'return'      => 'ids',
		);
```

shendy-a8c: You're right that it might not matter but not calling `array_unique()` or `array_values()` isn't much of an optimization. I won't know how that values would be used down the line now or even more so later, so I thought just to be safe, I'd pass a proper array with proper index and non-duplicate values.

saas786: > You're right that it might not matter but not calling `array_unique()` or `array_values()` isn't much of an optimization.

It can have impact on performance, but for me it was more off cognitive load, than some optimization concern.

PR #162, opened 2022-05-03
Replace `get_post_meta()` with `$order->get_meta()` or `wcs_get_objects_property()`

### LINKED ISSUE #141: Replace `get_post_meta()` uses where we are fetching order meta data
Product thread: pdjTHR-Xp-p2

---

## Description

In this issue we're tackling only the uses of `get_post_meta()` where an order ID is passed. Subscription ID or Product ID are fine at this stage.

### How to update code?

These uses should be replaced with `$object->get_meta( $property )` or `$object->get_{property_name}()` if the function exists.

### What code needs replacing?

Code that needs to be updated:
- [x] `WC_Subscriptions_Order::get_meta()`
- [x] `WCS_Post_Meta_Cache_Manager::maybe_update_for_post_change()`
  - When `$post_id` belongs to a `shop_order`
- [x] `wcs_get_objects_property()`
  - Can use https://github.com/woocommerce/woocommerce-subscriptions/pull/3814/files for inspiration

### Code to ignore 
- `WCS_Related_Order_Store_CPT::get_related_subscription_ids()`
- `WCS_Related_Order_Store_CPT::add_relation()`
- `test-class-wcs-related-order-store-cpt.php`
- `test-class-wcs-related-order-store-cached-cpt.php`
- `WCS_PayPal_Standard_IPN_Handler::get_order_id_and_key()`
  - This can probably be ignored for now because 'shop_order' post type is never queried (even though it's the default option when calling the function). I believe this default option was pre-WC subscriptions 2.0 before a subscription post type existed and we had to use the parent order ID.
- `WCS_Post_Meta_Cache_Manager_Many_To_One::meta_updated()`
  - This class is only used for `'shop_subscription'` post types
- `WC_Subscriptions_Synchroniser::order_contains_synced_subscription()`
  - This function can be removed as part of removing functions deprecated in 2.0

These will be tackled in a separate issue with a refactor of the WCS_Related_Order_Store_CPT class. See internal discussion here: pdjTHR-Y8-p2

- `WCS_Upgrade_1_2::init()`
- `class WCS_Upgrade_2_0`

Upgrade scrips can be ignored at this stage as these might be removed from the next version. See internal discussion here: pdjTHR-Y8-p2
 

## Product impact

mattallan:
Fixes #141 

## Description

This PR updates sections of our code that calls `get_post_meta()` where an $order ID is passed:

- `WC_Subscriptions_Order::get_meta()`

This function wasn't being used in our code anyway, apart from in very old upgrade scripts, so I updated the function to remove the `get_post_meta()` usage but I also deprecated it. Since WC Subscriptions 2.0, this function has been replaced with `wcs_get_objects_property()`

- `WCS_Post_Meta_Cache_Manager::maybe_update_for_post_change()`

In order to make this function work with Orders in a custom order table, I needed to update it to get the order or post object first, then get the meta using our `wcs_get_objects_property()` function.

Because this class has the word `Post` in it (`WCS_Post_Meta_Cache_Manager`), we could look at creating a whole new class for handling object meta cache (i.e. `WCS_Object_Meta_Cache_Manager`). Keen to get others thoughts on this :)

- `wcs_get_objects_property()`

The other change in this PR is to our helper function which is used to get meta data for a given object (subscriptions, order, or post). Rather than getting rid of the `get_post_meta()` uses entirely from this function, I updated the function so that it always uses `$object->get_meta()` function for orders and subscriptions.

#### Testing changes to `WCS_Post_Meta_Cache_Manager::maybe_update_for_post_change()`

Affected flow: Viewing subscriptions related orders in WP Admin and from My Account page
1. Purchase a new subscription
2. Process a renewal
3. View the Edit Subscription > Related Order meta box
4. Process another renewal
5. View the number of orders for the subscription is correct in **WooCommerce > Subscriptions** table
6. Visit My Account > Subscriptions > View
7. Scroll down and check related orders table
8. Delete a related order

#### Testing`wcs_get_objects_property()` changes

The above testing instructions will work for these changes as well.

## Product impact

### DIFF
--- a/includes/class-wc-subscriptions-order.php
+++ b/includes/class-wc-subscriptions-order.php
@@ -325,40 +325,6 @@ class WC_Subscriptions_Order {
 	}
 
 	/**
-	 * A unified API for accessing subscription order meta, especially for sign-up fee related order meta.
-	 *
-	 * Because WooCommerce 2.1 deprecated WC_Order::$order_custom_fields, this function is also used to provide
-	 * version independent meta data access to non-subscription meta data.
-	 *
-	 * @param WC_Order|int $order The WC_Order object or ID of the order for which the meta should be sought.
-	 * @param string $meta_key The key as stored in the post meta table for the meta item.
-	 * @param mixed $default (optional) The default value to return if the meta key does not exist. Default 0.
-	 * @since 1.0
-	 */
-	public static function get_meta( $order, $meta_key, $default = 0 ) {
-
-		if ( ! is_object( $order ) ) {
-			$order = wc_get_order( $order );
-		}
-
-		$meta_key = preg_replace( '/^_/', '', $meta_key );
-
-		if ( isset( $order->$meta_key ) ) { // WC 2.1+ magic __isset() & __get() methods
-			$meta_value = $order->$meta_key;
-		} elseif ( is_array( $order->order_custom_fields ) && isset( $order->order_custom_fields[ '_' . $meta_key ][0] ) && $order->order_custom_fields[ '_' . $meta_key ][0] ) {  // < WC 2.1+
-			$meta_value = maybe_unserialize( $order->order_custom_fields[ '_' . $meta_key ][0] );
-		} else {
-			$meta_value = get_post_meta( wcs_get_objects_property( $order, 'id' ), '_' . $meta_key, true );
-
-			if ( empty( $meta_value ) ) {
-				$meta_value = $default;
-			}
-		}
-
-		return $meta_value;
-	}
-
-	/**
 	 * Displays a few details about what happens to their subscription. Hooked
 	 * to the thank you page.
 	 *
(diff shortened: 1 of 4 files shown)


mattallan:
Making a note that the one test that is failing should be fixed in [3d1e6f0](https://github.com/Automattic/woocommerce-subscriptions-core/pull/159/commits/3d1e6f087345ed79bdd458a6ece25e43bd31b730)

shendy-a8c (approved):
Looks good!

- [x] Tested as instructed and everything looks good.
- [x] Confirmed that affected flow has been addeed to our shared doc for all the flows to be tested for this COT project.
- [x] I re-ran the GitHub checks and now they are all green, so confirmed @mattallan's comment [here](https://github.com/Automattic/woocommerce-subscriptions-core/pull/162#issuecomment-1116961889).

I cannot test with WCPay Subscription because got a fatal error when trying to use Test Clock. Filed an issue here: 59-gh-Automattic/woocommerce-payments-dev-tools

> Because this class has the word Post in it (WCS_Post_Meta_Cache_Manager), we could look at creating a whole new class for handling object meta cache (i.e. WCS_Object_Meta_Cache_Manager). Keen to get others thoughts on this :)

In my opinion, I don't think we need to do something about it.

shendy-a8c:
+cc @brucealdridge as an extra set of eyes since I'm committing [7228a7e](https://github.com/Automattic/woocommerce-subscriptions-core/pull/162/commits/7228a7e4d42a7223ccd0be932c03898061e7e71f) even though very minor.

brucealdridge (commented):
LGTM

Issue #155, opened 2022-04-25
Allow subscriptions database upgrades on headless sites (was Cron / Action Scheduler processing disabled until manual intervention on update)


haszari:
## Description

There is no way to upgrade the database (when required) on headless sites. For previously required upgrades, they required either manually clicking a button or by a user visiting `wp-admin`. 

### Investigation

In upgrader `init()`, subscriptions upgrader puts a lock on WP cron if the version of subscriptions set in the database is lower than the currently installed version.  

This lock is only removed once an upgrade is completed by logging in via wp-admin.  This won't cause any issues if the plugin is manually updated. However automatic updates require this manual intervention of logging in before any cron actions will proceed.

https://github.com/Automattic/woocommerce-subscriptions-core/blob/9ac1ae66fc1bc90b1abbfb9791eb880a9148bdda/includes/upgrades/class-wc-subscriptions-upgrader.php#L51-L56

https://github.com/Automattic/woocommerce-subscriptions-core/blob/9ac1ae66fc1bc90b1abbfb9791eb880a9148bdda/includes/upgrades/class-wc-subscriptions-upgrader.php#L132-L141

This could impact other cron jobs and correct cron processing on larger or more complex sites.

## Product impact

- [x] Does this issue affect WooCommerce Subscriptions? yes 4343-gh-woocommerce/woocommerce-subscriptions
- [x] Does this issue affect WooCommerce Payments? yes https://github.com/Automattic/woocommerce-payments/issues/4195

p81Rsd-J3-p2#comment-1026
p81Rsd-J3-p2#comment-1025

brucealdridge:
**NB:** This issue originally occurred on a "headless site" - one where `wp-admin` will no be accessed.

haszari:
Adding all info from duplicate/similar issue:

- #158 

As I understand it this is an unusual use case – a site where wp-admin is rarely visited! That said, we want to ensure our upgrade processes are robust and work well for a variety of use cases 🚀. 

Marking this as low priority as I haven't seen many reports of affected merchants. There's an easy workaround – visit wp-admin to kick off the scheduler. 

Next steps (if this comes up again):

- Add reproduce steps for a new / clean site.
- The impact of the bug – i.e. specifically what breaks.
- Specific relevant details of how this impacted sites we know of – i.e. real use case for "headless".
- Workarounds for the issue - e.g. visit or automatically request an admin page after plugin update.
[...]

PR #163, opened 2022-05-06
Deprecate upgrade scripts for versions before version 3

### LINKED ISSUE #150: Remove old upgrade/repair scripts from the Subscription Core codebase
## Description

With Custom Order Tables being worked on by the WC Core team ([public announcement](https://developer.woocommerce.com/2022/01/17/the-plan-for-the-woocommerce-custom-order-table/)), some of our old subscription/order repair and upgrade scripts that use direct database queries on orders will no longer work.

Rather than refactoring/updating these functions, we discussed (pdjTHR-H1-p2#comment-859) the idea of removing these older functions and then have a clear documented path for stores to upgrade from very old versions of Subscriptions.

## Proposed changes

In the link shared above, we decided to remove any upgrade/repair scripts from before 3.0 (last major version) so the changes to close this issue would be:
 - [ ] Deprecate any upgrade/repair scripts classes from before 3.0.
 - [ ] Make sure no deprecated upgrade/repair classes are ran (update the `WC_Subscriptions_Upgrader` class to not trigger an old upgrades)
 - [ ] Update the `WC_Subscriptions_Upgrader` class to display an admin notice or some sort of error admin page when a store is upgrading from an unsupported version with a link to documentation with Steps on how to upgrade.

## Product impact

  - While unlikely, there's potential for these changes to impact WC Pay in that a store could have been using an old version of subscriptions and then switched to the latest version bundled in WC Payments. 

## Dev notes

Something worth mentioning is that none of the repair scripts have code that will break when orders are moved to custom tables, so we could technically keep them. My thinking is that we just remove them because it would seem odd if the upgrade script has been removed for a particular version, but the repair script still runs.

## Additional context

Project thread: pdjTHR-Xp-p2
Discussion on removing scripts before 3.0: pdjTHR-H1-p2

brucealdridge:
Fixes #150

## Description

⚠️ This code intentionally prevents database updates from happening when upgrading from older versions.

This PR deprecates upgrading procedures from versions prior to woocommerce-subscriptions-database version 3.0 (which was released Jan 2020.) Some code has been removed to prevent upgrades from running, where possible code has just been marked as deprecated. 

An admin notice has been added to warn users about the database upgrade.

image

Due to the removal of some of the really outdated upgrades I was able to disable the code that caused issue #155 

 **NB** Only the database updates are prevented from running. Other subscriptions functionality can still be used including adding & purchasing products.

## How to test this PR

1. On a clean WP/WC instance, install an old version of WooCommerce Subscriptions (prior to 3.0.0)
2. Deactivate WooCommerce Subscriptions
3. Build woocommerce-subscriptions-core, install & activate.
4. Install WooCommerce Payments. (to see Upgrade Notice)
5. Disable subscriptions-core. (this will use WCPay bundled version of subscriptions-core which will upgrade successfully)
6. Re-activate subscriptions-core

You may also want to check `woocommerce_subscriptions_active_version` during this testing to check that the version updates correctly.

## Product impact

This change will only affect those that have WooCommerce Subscriptions 2.6.5 or before installed. Prior to upgrading to the latest version of either WooCommerce Subscriptions or WooCommerce Payments.


assets/js/wcs-upgrade.js
```diff
@@ -1,3 +1,6 @@
+/**
+ * @deprecated x.x.x This file is no longer in use and can be removed in future.
```
shendy-a8c: Is `x.x.x` intentional? Would it be replaced by a version? What would it be?

includes/upgrades/class-wc-subscriptions-upgrader.php
```diff
@@ -48,15 +61,24 @@ public static function init() {
...
+			add_action(
+				'admin_notices',
+				function () {
+					self::show_unsupported_upgrade_path_notice();
+				}
+			);
+			return;
 		}
 
+		// Set the cron lock on every request with an out of date version, regardless of authentication level, as we can only lock cron for up to 10 minutes at a time, but we need to keep it locked until the upgrade is complete, regardless of who is browing the site
```
shendy-a8c: Should this comment be deleted?
I'm guessing it explains the
```
if ( $version_out_of_date ) {
    self::set_cron_lock();
}
```
block.

includes/upgrades/class-wc-subscriptions-upgrader.php
```diff
@@ -157,9 +178,9 @@ public static function upgrade() {
 		 */
 		do_action( 'woocommerce_subscriptions_before_upgrade',  WC_Subscriptions_Core_Plugin::instance()->get_plugin_version(), self::$active_version );
 
-		// Update the hold stock notification to be one week (if it's still at the default 60 minutes) to prevent cancelling subscriptions using manual renewals and payment methods that can take more than 1 hour (i.e. PayPal eCheck)
-		if ( '0' == self::$active_version || version_compare( self::$active_version, '1.4', '<' ) ) {
```
shendy-a8c, line 151: I'm not too clear why you removed the `version_compare()` call. Could you explain?

assets/js/wcs-upgrade.js
brucealdridge: Left x.x.x as we need to resolve the issue with version numbers or we would have the issue where a docblock reads like the following. I believe @mattallan is looking into this.
~~~
/**
 * @since 2.3.0
 * @deprecated 1.9.0
 */
 ~~~

includes/upgrades/class-wc-subscriptions-upgrader.php
brucealdridge: As this is removing support for upgrades from versions prior to `3.0.0`. An upgrade from a version prior to 1.4 is no longer supported and therefore the check is not needed. We do still need to check for initial installs which is why I left the `'0' == self::$active_version` in.

brucealdridge:
Adding blocked status until versioning issues have been resolved.

Original discussion: pdjTHR-10A-p2

brucealdridge:
Removing blocked status as required [versioning changes](https://github.com/Automattic/woocommerce-subscriptions-core/issues/239) have been released.

mattallan:
Hey @wjrosa, I've cleaned out the `WC_Subscriptions_Upgrader` class quite a lot. Here's a breakdown of the latest changes:

- Removed all irrelevant code from the `WC_Subscriptions_Upgrader::init()` function that was there to support our deprecated AJAX upgrade process.
- Moved all deprecated functions to the bottom of the file so keep all relevant functions at the top of the class.
- Updated all deprecated x.x.x tags and deprecated all other functions that are no longer used.
- Removed all unnecessary hooks that no longer are used

To test this PR:

#### Testing the admin notice
1. Set option `woocommerce_subscriptions_active_version` to 2.6.0
2. Confirm you see the admin notice explaining that this is not a supported version to upgrade from.

#### Confirm upgrades are working
1. Create an old cron event: `wp_schedule_event( time(), 'weekly', 'wcs_cleanup_big_logs' );`
2. Set option `woocommerce_subscriptions_active_version` to `3.1.0`
3. Reload an admin page
4. Call `wp_next_scheduled( 'wcs_cleanup_big_logs' );` and confirm the upgrade script for 6.8 was ran

#### Confirm defaults are set by the upgrade script on new stores
1. Delete the `woocommerce_subscriptions_active_version` option
2. Set option `woocommerce_subscriptions_multiple_purchase` to `no`
3. Reload an admin page to run the upgrade scripts
4. Confirm `woocommerce_subscriptions_multiple_purchase` was set to 'yes'

wjrosa (approved):
Thanks for the latest updates and the testing instructions, @mattallan! Works as expected:

### Testing the admin notice

> 2. Confirm you see the admin notice explaining that this is not a supported version to upgrade from.

Screenshot 2024-11-12 at 12 15 42

### Confirm upgrades are working

> 4. Call wp_next_scheduled( 'wcs_cleanup_big_logs' ); and confirm the upgrade script for 6.8 was ran

Screenshot 2024-11-12 at 13 07 50

### Confirm defaults are set by the upgrade script on new stores

> 4. Confirm woocommerce_subscriptions_multiple_purchase was set to 'yes'

Screenshot 2024-11-12 at 13 09 17

PR #165, opened 2022-05-11
Fix failed PHP 7.1 linting check

shendy-a8c:
## Description

Since WooCommerce just released `6.5` that requires PHP 7.2, our PHP 7.1 linting check [fails](https://github.com/Automattic/woocommerce-subscriptions-core/runs/6385651745?check_suite_focus=true#step:5:177).

This PR set WC version `6.4.1` to be installed when PHP version is <= 7.1.

## How to test this PR

Make sure `PHP linting and tests / PHP testing (7.1)` check is green.

## Product impact

### DIFF
--- a/.github/workflows/php-lint-test.yml
+++ b/.github/workflows/php-lint-test.yml
@@ -87,7 +87,14 @@ jobs:
           php-version: '7.1'
           tools: composer
       # run CI checks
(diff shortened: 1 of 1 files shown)


mattallan (approved):
I don't have any feedback for this one but the changes LGTM :+1:
Thanks Shendy for fixing this!

PR #177, opened 2022-06-02
Division by zero error during renewals

### LINKED ISSUE #175: Division by Zero fatal error in renewal logic (PHP8)
## Describe the bug

In php8 a Division By Zero [now throws an error](https://3v4l.org/8rait) whereas in older versions of php it only triggered a warning.
While there is no clear way to reproduce this error there have been instances of customers being affected.

A customer has flagged the following specific line during renewals 
https://github.com/Automattic/woocommerce-subscriptions-core/blob/97b448a082c61d684bbd29a2033788ffb6bced4d/includes/class-wcs-cart-renewal.php#L454

Wrapping this line in a try/catch, catching `DivisionByZeroError` seems to be the best solution as this will not change the behavior on older versions of php.

## How to reproduce
(Steps pasted from other issue)

It is tough to reproduce this issue as it is intermittent. In one of the older versions of WooCommerce Subscriptions I was able to reproduce this issue by:

1. Create two active subscription orders
1. Create pending renewal orders for both of them.
1. Open the customer payment page side by side in the browser for both renewals.
1. Close the one you opened last, and proceed to pay for the earlier one.
1. You'll see the DivisionByZeroError in logs. (This may be due to cart hashes clashing.)

But now with the latest version of WooCommerce Subscriptions, I am not able to reproduce the error using the steps above.

Next, I came across this ticket 5228964-zen, where the user is still facing the same error using the latest version of WooCommerce Subscriptions.

I decided to do some tests on my site (using Subs 4.2.0). For an active subscription order,

1. I created a pending renewal order.
1. Then, I deleted the item in the on-hold subscription order.
1. I tried making a payment for the pending renewal order using the "Customer payment page" link.
1. I was able to get the DivisionByZeroError in logs after the payment.

## Product impact

- [x] Does this issue affect WooCommerce Subscriptions? Yes, 4359-gh-woocommerce/woocommerce-subscriptions
- [x] Does this issue affect WooCommerce Payments? **Assuming yes – unconfirmed**

## Additional context

5228964-zen

brucealdridge:
Fixes #175

## Description

In php8 Division by zero throws an exception whereas in previous versions only a warning is thrown. This has caused issues in rare cases where the quantity is zero during a renewal

## How to test this PR

I haven't been able to reproduce this issue so have been unable to test.

There are some reproduction steps in #175 but results may vary.

## Product impact

- [x] Will this PR affect WooCommerce Subscriptions? 4359-gh-woocommerce/woocommerce-subscriptions
- [x] Will this PR affect WooCommerce Payments? Assuming yes – unconfirmed

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Subscriptions Core Changelog ***
 
+= 2.1.0 - 2022-xx-xx =
+* Fix - Fatal Error caused in rare cases where quantity is zero during renewal.
+
 = 2.0.0 - 2022-05-20 =
 * Dev - Retrieving users subscriptions order has been updated to use the WooCommerce specific APIs in WC_Subscriptions_Order.
 * Dev - Deprecate the WC_Subscriptions_Order::get_meta() function. Use wcs_get_objects_property( $order, $meta_key, "single", $default ) instead.
--- a/includes/class-wcs-cart-renewal.php
+++ b/includes/class-wcs-cart-renewal.php
@@ -451,7 +451,10 @@ class WCS_Cart_Renewal {
 					}
 				}
 
-				$_product->set_price( $price / $item_to_renew['qty'] );
+				// In rare cases quantity can be zero. Check first to prevent triggering a fatal error in php8+
+				if ( 0 !== $item_to_renew['qty'] ) {
+					$_product->set_price( $price / $item_to_renew['qty'] );
+				}
 
 				// Don't carry over any sign up fee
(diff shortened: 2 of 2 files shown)


mattallan (approved):
Nice one @brucealdridge !

I had some thoughts when reviewing these changes:
 - should we set a default qty to 1 if it's 0 or empty?
 - is it okay to skip updating the product price?

After thinking through these questions I had, I think your approach is fine because when the quantity is 0, the cart item won't get added to the cart from the session anyway and so we don't want to change that behavior by setting the qty to 1 and skipping updating the product price is fine because the item won't get added to the cart because it has 0 quantity 🤪

Issue #186, opened 2022-08-02
Refactor `wcs_get_subscription()` to not use Order Factory (Unable to purchase subscription with COT enabled)


mattallan:
> **Warning** - This issue may not need to be implemented.
> Now that we are going to support/migrate Subscriptions over COT/HPOS tables, the Order Factory should work as we expect it to and so we no longer need to refactory this function.
>
> Once we have implemented a HPOS Subscriptions datastore class and updated `wcs_create_subscription()` to create subscriptions via this datastore when HPOS is enabled, we will need to confirm whether we still need to fix this issue. I suspect we won't

---

## Describe the bug

This issue was discovered by trying to purchase a subscription when custom order tables is enabled on a store. But this is only one of many issues caused by this problem.

Error received:
```
2022-08-02T05:40:06+00:00 CRITICAL Uncaught Error: Call to a member function get_items() on bool in /srv/users/usereaea7e7b/apps/usereaea7e7b/public/wp-content/plugins/woocommerce-subscriptions/vendor/woocommerce/subscriptions-core/includes/class-wc-subscriptions-synchroniser.php:1148
Stack trace:
#0 /srv/users/usereaea7e7b/apps/usereaea7e7b/public/wp-includes/class-wp-hook.php(309): WC_Subscriptions_Synchroniser::maybe_add_subscription_meta(1170)
#1 /srv/users/usereaea7e7b/apps/usereaea7e7b/public/wp-includes/class-wp-hook.php(331): WP_Hook->apply_filters(NULL, Array)
#2 /srv/users/usereaea7e7b/apps/usereaea7e7b/public/wp-includes/plugin.php(476): WP_Hook->do_action(Array)
#3 /srv/users/usereaea7e7b/apps/usereaea7e7b/public/wp-includes/post.php(4673): do_action('save_post', 1170, Object(WP_Post), false)
#4 /srv/users/usereaea7e7b/apps/usereaea7e7b/public/wp-content/plugins/woocommerce-subscriptions/vendor/woocommerce/subscriptions-core/wcs-functions.php(180): wp_insert_post(Array, true)
#5 /srv/users/usereaea7e7b/apps/usereaea7e7b/public/wp-c in /srv/users/usereaea7e7b/apps/usereaea7e7b/public/wp-content/plugins/woocommerce-subscriptions/vendor/woocommerce/subscriptions-core/includes/class-wc-subscriptions-synchroniser.php on line 1148
```

[...]

PR #189, opened 2022-08-03
When COT is enabled, allow using wc_get_order() with a Subscription ID to get a subscription object

### LINKED ISSUE #187: Using wc_get_order() with a subscription ID doesn't work when Custom Order Tables are enabled (Unable to purchase Subscription with WCPay)
## Describe the bug

After enabling COT on your store and trying to purchase a subscription with WooCommerce Payments, there's a fatal error that occurs on the checkout:

```
PHP Fatal error:  Uncaught Error: Call to a member function get_payment_tokens() on bool in /Users/matt/local/woo/wp-content/plugins/woocommerce-payments/includes/class-wc-payment-gateway-wcpay.php:1478
Stack trace:
#0 /Users/matt/local/woo/wp-content/plugins/woocommerce-payments/includes/compat/subscriptions/trait-wc-payment-gateway-wcpay-subscriptions.php(678): WC_Payment_Gateway_WCPay->get_payment_token(false)
#1 /Users/matt/local/woo/wp-content/plugins/woocommerce-payments/includes/class-wc-payment-gateway-wcpay.php(2530): WC_Payment_Gateway_WCPay->maybe_schedule_subscription_order_tracking(213, false)
#2 /Users/matt/local/woo/wp-includes/class-wp-hook.php(307): WC_Payment_Gateway_WCPay->schedule_order_tracking(213)
#3 /Users/matt/local/woo/wp-includes/class-wp-hook.php(331): WP_Hook->apply_filters('', Array)
#4 /Users/matt/local/woo/wp-includes/plugin.php(476): WP_Hook->do_action(Array)
#5 /Users/matt/local/woo/wp-content/plugins/woocommerce-subscriptions-core/includes/data-stores/class-wcs-subscription-data-store-cpt.php in /Users/matt/local/woo/wp-content/plugins/woocommerce-payments/includes/class-wc-payment-gateway-wcpay.php on line 1478
```

This error is caused by [this line](https://github.com/Automattic/woocommerce-payments/blob/5615b8854175b81f580a905159f2225884b73e59/includes/compat/subscriptions/trait-wc-payment-gateway-wcpay-subscriptions.php#L675) in WooCommerce Payments which is calling `wc_get_order( $order_id )` where Order ID is actually a Subscription ID.

The reason this is failing is because subscriptions are staying in the WP Posts table and when the COT feature is enabled, the new `OrderTablesDataStore` data store is loaded and so using `wc_get_order()` can't find the order type of a subscriptionID that isn't in the COT.

### Possible Solution?
One workaround I'm thinking is we could filter `woocommerce_order_class` and check if the `$order_type` arg is empty and that `$order_id` is a subscription post ID. If it is then return 'WC_Subscription' classname which should fix these issues

## To Reproduce

```
function enable_cot(){
   }
}
```
7. Try to purchase a subscription on the checkout with WC Payments

## Product impact

mattallan:
Fixes #187

## Description

When a store has Custom Order tables enabled (coming in WC 6.9), trying to purchase a subscription on the checkout results in the following fatal error:

```
2022-08-02T05:40:06+00:00 CRITICAL Uncaught Error: Call to a member function get_items() on bool in /srv/users/usereaea7e7b/apps/usereaea7e7b/public/wp-content/plugins/woocommerce-subscriptions/vendor/woocommerce/subscriptions-core/includes/class-wc-subscriptions-synchroniser.php:1148
```

This error is caused by our `wcs_get_subscription()` using the `WC()->order_factory->get_order( $subscription_id )` which no longer works to fetch a Subscription object. This is because Subscription object are staying in the wp posts table and when COT is enabled, the new order data store is loaded and so calling `WC_Data_Store::load( 'order' )->get_order_type( $subscription_id )` returns an empty string as the 'shop_subscription' order type is unknown in this new context.

Using `wc_get_order()` with a subscription ID is not only used within our own codebase but is used widely by other plugins (like WC Payments for one) and so this PR aims to provide a workaround for this problem.

As per this internal slack discussion (p1659414499655349-slack-C9QDV4TL4), it was suggested to refactor our `wcs_get_subscriptions()` and remove the Order Factory usage, however, after doing that there was other issues outside of Subscriptions Core where `wc_get_orders()` was being called with a subscription ID and returning `false`. See #187 which was an issue in WC Payments.

---

With this PR, the issue in #186 is also fixed but since it was suggested we remove Order Factory code from `wcs_get_subscription()`, I'm going to leave that issue open for us to still get to.

## How to test this PR

1. Checkout the latest WooCommerce `cot-branch` and build it
```
function enable_cot(){
   }
}
```
7. Try to purchase a subscription on the checkout with WC Payments

## Product impact

### DIFF
--- a/includes/data-stores/class-wcs-subscription-data-store-cpt.php
+++ b/includes/data-stores/class-wcs-subscription-data-store-cpt.php
@@ -225,7 +225,7 @@ class WCS_Subscription_Data_Store_CPT extends WC_Order_Data_Store_CPT implements
 	 * @return string
 	 * @since 2.2.0
 	 */
-	public function get_total_refunded( $subscription ) : float {
+	public function get_total_refunded( $subscription ) {
 
 		$total = 0;
 
--- a/includes/data-stores/class-wcs-subscription-data-store-cpt.php
+++ b/includes/data-stores/class-wcs-subscription-data-store-cpt.php
@@ -225,7 +225,7 @@ class WCS_Subscription_Data_Store_CPT extends WC_Order_Data_Store_CPT implements
 	 * @return string
 	 * @since 2.2.0
 	 */
-	public function get_total_refunded( $subscription ) {
+	public function get_total_refunded( $subscription ) : float {
 
 		$total = 0;
 
--- a/includes/class-wc-subscriptions-core-plugin.php
+++ b/includes/class-wc-subscriptions-core-plugin.php
@@ -201,6 +201,11 @@ class WC_Subscriptions_Core_Plugin {
 		if ( class_exists( 'WC_Abstract_Privacy' ) ) {
 			new WCS_Privacy();
 		}
+
+		// Load Subscriptions COT compatibility class for stores that have COT enabled.
+		if ( method_exists( 'Automattic\WooCommerce\Utilities\OrderUtil', 'custom_orders_table_usage_is_enabled' ) && Automattic\WooCommerce\Utilities\OrderUtil::custom_orders_table_usage_is_enabled() ) {
+			WC_Subscriptions_COT_Compatibility::init();
+		}
 	}
 
 	/**
--- a/includes/class-wcs-core-autoloader.php
+++ b/includes/class-wcs-core-autoloader.php
@@ -247,6 +247,8 @@ class WCS_Core_Autoloader {
 			$path .= '/privacy';
 		} elseif ( false !== strpos( $class, 'upgrade' ) || false !== strpos( $class, 'repair' ) ) {
 			$path .= '/upgrades';
+		} elseif ( false !== strpos( $class, 'compatibility' ) ) {
+			$path .= '/compatibility';
 		}
 
 		return trailingslashit( $path );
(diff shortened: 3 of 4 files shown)


haszari:
> Using wc_get_order() with a subscription ID is not only used within our own codebase but is used widely by other plugins (like WC Payments for one) and so this PR aims to provide a workaround for this problem.

This seems like a major issue – @mattallan  can you start a P2 on this topic to get alignment about the approach? At first glance, patching to make `wc_get_order` work may prolong the pain. Can we introduce a `wc_get_subscription` and mandate plugins move to this before being "Custom Order Tables Certified"? (to be discussed!)

includes/data-stores/class-wcs-subscription-data-store-cpt.php
```diff
@@ -225,7 +225,7 @@ protected function update_post_meta( &$subscription ) {
 	 * @return string
 	 * @since 2.2.0
 	 */
-	public function get_total_refunded( $subscription ) {
+	public function get_total_refunded( $subscription ) : float {
```
shendy-a8c: This is to fix fatal error `Declaration of WCS_Subscription_Data_Store_CPT::get_total_refunded($subscription) must be compatible with WC_Order_Data_Store_CPT::get_total_refunded($order): float`.

mattallan:
Closing this in favour of adding support for Subscriptions in Custom Order Tables.

PR #204, opened 2022-08-17
Move shipping meta fields

### LINKED ISSUE #203: One time shipping admin fields have broken markup
## Describe the bug
The [shipping fields close a div and then leave a new one opened but unclosed](https://github.com/Automattic/woocommerce-subscriptions-core/blob/c383a591f7686991615593475247ee5dfd935cdd/includes/admin/class-wc-subscriptions-admin.php#L389-L390).

This can have unintended consequences on other plugins that are attaching fields on the `woocommerce_product_options_shipping` hook. Basically if they don't mimic your markup pattern then the admin page can have broken markup.

As of WC 6.0 woo has added the [`woocommerce_product_options_shipping_product_data` filter](https://github.com/woocommerce/woocommerce/blob/3611d4643791bad87a0d3e6e73e031bb80447417/plugins/woocommerce/includes/admin/meta-boxes/views/html-product-data-shipping.php#L66) outside of the `<div class="options_group">`

This hook is generally better since you can properly open and close div elements which is more reliable for other extensions.

helgatheviking:
Move fields admin shipping fields to `woocommerce_product_options_shipping_product_data` hook. 

Fixes #203

## Description

As of 6.0, Woo has the `woocommerce_product_options_shipping_product_data` which is outside of an option_group div. Attaching there, with properly closed markup, makes it predictable for other plugins to also attach there with proper markup and nobodies fields get accidentally swallowed up in an unclosed div.

## How to test this PR

My memory is a bit dodgy, but I think prior to the PR you could add something like this:

```
function kia_test_subs_shipping() {
    echo '<div>please send tacos</div>';
}
add_action( 'woocommerce_product_options_shipping', 'kia_test_subs_shipping' );
```

and it would be sucked up into the Subs shipping div and therefore displayed according to subs' hide/show script rules. 

## Product impact

### DIFF
--- a/includes/admin/class-wc-subscriptions-admin.php
+++ b/includes/admin/class-wc-subscriptions-admin.php
@@ -392,7 +392,7 @@ class WC_Subscriptions_Admin {
 
 		$needs_html_fix = 'woocommerce_product_options_shipping' === current_filter();
 
-		 // Old hook is nested and requires invalid html markup to be compatible with other plugins.
+		// Old hook is nested and requires invalid html markup to be compatible with other plugins.
 		if ( $needs_html_fix ) {
 			echo '</div>';
 		}
--- a/includes/admin/class-wc-subscriptions-admin.php
+++ b/includes/admin/class-wc-subscriptions-admin.php
@@ -71,7 +71,11 @@ class WC_Subscriptions_Admin {
 		add_action( 'woocommerce_page_wc-status', __CLASS__ . '::clear_subscriptions_transients' );
 
 		// Add subscription shipping options on edit product page
-		add_action( 'woocommerce_product_options_shipping', __CLASS__ . '::subscription_shipping_fields' );
+		if ( wcs_is_woocommerce_pre( '6.0' ) ) {
+			add_action( 'woocommerce_product_options_shipping', __CLASS__ . '::subscription_shipping_fields' );
+		} else {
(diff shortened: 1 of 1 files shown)


mattallan (approved):
Hey @helgatheviking, thanks for submitting this PR! I've tested this on the latest WC and pre 6.0 and both look good.

---

For others, here's the PR which added this change to WC Core: https://github.com/woocommerce/woocommerce/pull/30876

includes/admin/class-wc-subscriptions-admin.php
```diff
@@ -386,7 +390,13 @@ public static function subscription_pricing_fields() {
 	public static function subscription_shipping_fields() {
 		global $post;
 
-		echo '</div>';
+		$needs_html_fix = 'woocommerce_product_options_shipping' === current_filter();
+
+		 // Old hook is nested and requires invalid html markup to be compatible with other plugins.
```
mattallan:
```suggestion
		// Old hook is nested and requires invalid html markup to be compatible with other plugins.
```

mattallan: I noticed a small whitespace issue before merging

mattallan:
Making a note that I also added a changelog entry for this one. See [5b1baa4](https://github.com/Automattic/woocommerce-subscriptions-core/commit/5b1baa4ced00225823bc2f91ff731ba4f6c6f5fb).

Feel free to let me know if the wording should be changed.

PR #206, opened 2022-08-24
Fix for rare cases when division by zero error happens during renewal

### LINKED ISSUE #197: Fatal divide by zero error when quantity is 0 in PHP8
## Describe the bug

In #177, a fix was implemented to prevent a divide by zero error if the store is running on PHP8. The following code is used to check for this:

```php
// In rare cases quantity can be zero. Check first to prevent triggering a fatal error in php8+
                if ( 0 !== $item_to_renew['qty'] ) {
                    $_product->set_price( $price / $item_to_renew['qty'] );
                }
```

However, it looks like `$item_to_renew['qty']` is not an integer (presumably it is a string), so this check doesn't have the desired effect. We had a customer report (5461670-zd-woothemes) reporting a fatal error and proposing the following fix:

```php
// In rare cases quantity can be zero. Check first to prevent triggering a fatal error in php8+
                if ( 0 !== intval($item_to_renew['qty']) ) {
                    $_product->set_price( $price / intval($item_to_renew['qty']) );
                }
```

This should prevent this error happening in all cases.

## To Reproduce

See #175 (alternatively, the customer who raised this issue suggested it occurs when a new customer logs into their store).

### Expected behavior

We don't see any fatal errors related to division by zero errors.

- [x] Does this issue affect WooCommerce Subscriptions? yes: 5461670-zd-woothemes
- [ ] Does this issue affect WooCommerce Payments? TBC

brucealdridge:
Fixes #197 

## Description

A DivisionByZero exception was thrown in rare cases during renewals. We were unable to replicate the circumstances that triggered the bug. #177 was raised to fix this bug.

As replication wasn't possible I tested around the affected code and confirmed that `$item_to_renew['qty']` was always an `int`. New bug reports (#197) show that this is not always the case.  This is likely the cause of the error, a missing or invalid quantity.

The solution to prevent these fatal errors is to ensure the variable is cast to an `int`
```diff
-if ( 0 !== $item_to_renew['qty'] ) {
+if ( 0 !== (int) $item_to_renew['qty'] ) {
```
 
## Product impact

- [x] Does this issue affect WooCommerce Subscriptions? Yes, 4359-gh-woocommerce/woocommerce-subscriptions
- [x] Does this issue affect WooCommerce Payments? **Assuming yes – unconfirmed**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Subscriptions Core Changelog ***
 
+= 2.3.0 - 2022-xx-xx =
+* Fix - Fatal Error caused in rare cases where quantity is zero during renewal.
+
 = 2.2.0 - 2022-08-03 =
 * Fix - Update subscription address when changed with renewals on block checkout.
 
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,6 @@
 *** WooCommerce Subscriptions Core Changelog ***
 
-= 2.3.0 - 2022-xx-xx =
+= 2.2.1 - 2022-xx-xx =
 * Fix - Fatal Error caused in rare cases where quantity is zero during renewal.
 
 = 2.2.0 - 2022-08-03 =
--- a/includes/class-wcs-cart-renewal.php
+++ b/includes/class-wcs-cart-renewal.php
@@ -454,7 +454,7 @@ class WCS_Cart_Renewal {
 				}
 
 				// In rare cases quantity can be zero. Check first to prevent triggering a fatal error in php8+
-				if ( 0 !== $item_to_renew['qty'] ) {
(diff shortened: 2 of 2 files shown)


mattallan (approved):
Thanks @brucealdridge 💯 

Just noting that I also tried to reproduce this and had no luck. I thought maybe there was a difference with getting the `$item_to_renew` object from `$subscription->get_items()` vs `WC_Order_Factory::get_order_item()` where one was returning stringed integer, but I checked this and there are no issues.

I also had a look over our codebase to double-check we're not setting the quantity of a subscription item to `'0'` but I couldn't see anything. I'm guessing there's some third party causing this which is probably why we're struggling to reproduce this.

Your changes and approach LGTM. Marking as good to merge :)

PR #210, opened 2022-09-15
Add order note when subscription address is changed.

tommyshellberg:
Addresses 3747-gh-woocommerce/woocommerce-subscriptions

## Description
When a customer updates their address from the My Account->My Subscriptions page, add an order note such that the store owner can be made aware of the change. Currently, an address change made by a customer is silent; the store owner may not notice and taxes and/or shipping could be incorrectly charged on subsequent renewals. This PR creates a note and prompts the store owner that a recalculation of the subscription may be necessary.

Questions for PR author:
- How can this code break?
This code relies on `$subscription` being an instance of `WC_Subscription` in order to call `add_order_note()`. It's possible to try to get a WC_Subscription object and instead get a `false` return value but the code already in place above this new code should handle that( it checks if it has a valid status or if `$subscription` is truthy). So we shouldn't accidentally cause a fatal by trying to call an undefined method.

## How to test this PR

1. Create a Simple Subscription product and purchase it to activate the subscription.
2. Go to My Account->My Subscriptions->Change Address, enter a new address, save.
3. Go to WP Admin->Subscriptions, click on the subscription, and note that the Subscription notes area on the right has a note that says:
> Subscription address has been changed. You may need to recalculate taxes and/or shipping.

Screen Shot 2022-09-15 at 15 31 01

- [x] Will this PR affect WooCommerce Payments? unsure

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 *** WooCommerce Subscriptions Core Changelog ***
 
 = 2.3.0 - 2022-xx-xx =
+* Add - Add a subscription/order note when the Subscription address has been changed.
 * Fix - Move One Time Shipping metabox fields to use the woocommerce_product_options_shipping_product_data hook introduced in WC 6.0.
 
 = 2.2.1 - 2022-08-25 =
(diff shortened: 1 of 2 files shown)


tommyshellberg:
@mattallan it seems like you've been doing all of the reviews lately for this repo, please let me know if it should be re-assigned to someone else. :)

haszari:
Thanks @tommyshellberg ! I've requested a review from `helix`, that's how we handle reviews.

brucealdridge:
Thanks for the PR @tommyshellberg 

The team has spent some time on this issue and has concluded that rather than merge this PR it would be better to fix the issue properly. In normal circumstances the note created in this PR would unlikely been seen.
 We have a number of PR's in progress that will mitigate the issue. 

p1664854771299639/1664853789.973789-slack-C02BW3Z8SHK

PR #217, opened 2022-10-11
Load the Subscriptions HPOS data store when the feature is enabled and in-use

### LINKED ISSUE #187: Using wc_get_order() with a subscription ID doesn't work when Custom Order Tables are enabled (Unable to purchase Subscription with WCPay)

### LINKED ISSUE #214: When HPOS and custom Order tables feature is enabled, load the new Subscriptions datastore
## Describe the bug

When HPOS and custom order tables is enabled, we need to make sure any code calling `WC_Data_Store::load( 'subscription' )` is loading the proper datastore class.

There are a few options we can explore to tackle this:
 1. Update our [`WC_Subscriptions_Core_Plugin->add_data_stores()`](https://github.com/Automattic/woocommerce-subscriptions-core/blob/trunk/includes/class-wc-subscriptions-core-plugin.php#L390) function to check if HPOS is enabled and if so, set the `'subscription'` datastore to the new datastore class added in https://github.com/Automattic/woocommerce-subscriptions-core/issues/213
 2. Create a new HPOS/COT controller class similar to what WC Core does with their [`CustomOrdersTableController`](https://github.com/woocommerce/woocommerce/blob/3c8ca2887e8a9566065569fb3c8d5aba99c3d9f4/plugins/woocommerce/src/Internal/DataStores/Orders/CustomOrdersTableController.php#L23) class and then attach a filter to `woocommerce_subscription_data_store` which will override the default CPT datastore class.

I kind of like the idea of keeping this logic in a HPOS controller class (option 2) which only gets loaded when HPOS is enabled. This helps us contain the logic in a single class and allow us to unhook and load in functionality when needed. Open to more ideas with how to implement this

## Impact

Throughout our codebase we currently use `WC_Data_Store::load( 'subscription' )` only 4 times:
1. One in our `wcs_subscription_search()` helper function
2. Three times in our privacy manager/eraser classes

After we support Subscriptions in HPOS, we will be need to replace all of our `get_post_type( $subscription_id )` code with something like `WC_Data_Store::load( 'subscription' )->get_order_type( $post_id )` so this change will impact a lot of our code after HPOS has been implemented.

## Additional context

pdjTHR-1Fu-p2

mattallan:
For #214
For #187

> **Note** This PR is being merged into our HPOS base branch: `hpos_subscriptions_datastore`

## Description

This PR introduces a new class `WCS_Orders_Table_Data_Store_Controller` to load the new Subscriptions Order Tables datastore class when HPOS is enabled and WooCommerce order tables are set as the authoritative data source.

image

When testing these changes, I checked that `wcs_get_subscription( $id );` and `wc_get_order( $id )` will now search for `shop_subscription` objects from `wp_wc_orders` table when HPOS is active on the store which means we can also close #187 with this PR.

Other code snippets I checked include:
 - `$subscription->data_store->get_current_class_name()` correctly returns the new data store class `WCS_Orders_Table_Subscription_Data_Store`
 - `WC_Data_Store::load( 'subscription' )->get_order_type( $subscription_id );` now returns 'shop_subscription'

## How to test this PR

To test this PR, you'll need to:
 - Install the latest [WooCommerce Core HPOS zip](https://github.com/woocommerce/woocommerce/releases/tag/feature-custom-order-table) and generate the HPOS tables
 - Add a `shop_subscription` into the new `wp_wc_orders` (to do this I just purchased an order via the checkout and then changed the `type` column in the table from `shop_order` to `shop_subscription`
 - Have a `shop_subscription` in WP Posts as well to test flicking the authoritative order data store between COT and WP Posts
 - I also installed WP Console to test some dummy code

On this branch I ran the following code snippets to test fetching and checking subscription objects are getting fetched using the proper datastore depending on whether HPOS is enabled: https://gist.github.com/mattallan/c1d5ea96211cc5e76fe25251f856e918

## Product impact


mattallan:
The current PHPCS issues are coming from the Orders Table Datastore coming from the base branch (not introduced in this PR)

includes/wcs-compatibility-functions.php
```diff
@@ -570,3 +570,39 @@ function wcs_add_woocommerce_dependent_action( $tag, $function, $woocommerce_ver
...
+ *  - HPOS is the authoriative source of order data
+ *
+ * @return bool
+ */
+function wcs_is_custom_order_tables_usage_enabled() {
+	if ( ! class_exists( '\Automattic\WooCommerce\Utilities\OrderUtil' ) || ! wcs_is_wc_feature_enabled( 'custom_order_tables' ) ) {
+		return false;
+	}
+
+	return \Automattic\WooCommerce\Utilities\OrderUtil::custom_orders_table_usage_is_enabled();
```
brucealdridge, line 607: both `wcs_is_wc_feature_enabled( 'custom_order_tables' )` and `OrderUtil::custom_orders_table_usage_is_enabled();` use a simple `get_option()` so not very taxing. 
Will there ever be a case where `wcs_is_wc_feature_enabled( 'custom_order_tables' )` is false while `OrderUtil::custom_orders_table_usage_is_enabled();` is true? I'm wondering if we can remove the `wcs_is_wc_feature_enabled` check.

mattallan: > Will there ever be a case where `wcs_is_wc_feature_enabled( 'custom_order_tables' )` is false while `OrderUtil::custom_orders_table_usage_is_enabled();` is true? 

From my understanding and testing this is possible. If I have the COT tables turned on in settings (https://d.pr/i/QzlhJ9) and then turn off the HPOS feature (https://d.pr/draw/X9ZBnX/), the `woocommerce_custom_orders_table_enabled` option in the database is still set to 'yes'.

So we can't just use `OrderUtil::custom_orders_table_usage_is_enabled()` and have to check both the feature is enabled and the COT tables are in-use.

PR #226, opened 2022-10-16
Refactor wcs_create_subscription to use CRUD methods

brucealdridge:
## Description

Before we can start adding support for HPOS in subscriptions-core `wcs_create_subscription` needs to be refactored. 
This is the entry point when creating a subscription object in the database which is used to create renewals.
Currently, this calls WP API's (`wp_insert_post`) rather than WC specific API's to create the subscription and add relevant metadata. 
Once this is refactored to work with the CPT datastore directly (using WC API's), we will be able to switch to the HPOS datastore when ready.

This PR has been split off from the original HPOS PR (#224) to aid in testing and merging this work sooner. 

## How to test this PR

Create subscriptions in a variety of circumstances both before and after checking out this PR. 
This PR is working as expected when there are no unexpected changes in data between the two.

Sample steps
1. On `trunk` Purchase a subscription. Note down subscription id 
2. On `refactor-create-subscription` purchase a subscription using the same payment method. Note the subscription id.
3. Using the below SQL, compare differences in results for anything unexpected.

Using this testing method we have confirmed the following work without issues
 - Manual creation
 - Using BACS payment method
 - Using WCPay billing & tokenised subscriptions
 - Using Square
 - Using Cart/Checkout blocks

```sql
SET @control_id=xxxx; /* Subscription id from trunk */
SET @refactor_id=yyyy; /* Subscription id from refactor-create-subscription  */
SELECT mk.meta_key, c.meta_value as control_value, r.meta_value as refactor_value
FROM (select DISTINCT meta_key from wp_postmeta WHERE post_id in (@control_id, @refactor_id)) mk
LEFT JOIN wp_postmeta c ON c.meta_key=mk.meta_key && c.post_id=@control_id
LEFT JOIN wp_postmeta r ON r.meta_key=mk.meta_key && r.post_id=@refactor_id
WHERE c.meta_value != r.meta_value || c.meta_value IS NULL || r.meta_value IS NULL;
```

**Sample results**
```
meta_key				control_value			refactor_value
_schedule_next_payment			2022-10-28 00:48:59		2022-10-28 00:49:03
_schedule_start				2022-10-26 00:37:52		2022-10-26 00:38:57
_subscription_switch_data		a:0:{}	
_order_key				wc_order_6zVJYYYPluZCd		wc_order_oTxSuXBJQatSY
_billing_address_index	         	   test@example.org 		John Doe  60 29th Street #343  San Francisco CA 94110 US test@example.org (555) 555-5555
_subscription_renewal_order_ids_cache	a:1:{i:0;i:4857;}	a:1:{i:0;i:4858;}
```

**NB:** There are current issues with `_billing_address_index` and `shipping_address_index` which is resolved in #228 

## Product impact

- [x] Will this PR affect WooCommerce Payments?


includes/data-stores/class-wcs-subscription-data-store-cpt.php
```diff
@@ -94,10 +94,27 @@ public function __construct() {
...
+		$default_subscription_data = array(
+			'post_status' => 'wc-' . apply_filters( 'woocommerce_default_subscription_status', 'pending' ),
+		);
+
+		/**
+		 * This function is called on the `woocommerce_new_order_data` filter.
+		 * We hook into this function, calling our own filter `woocommerce_new_subscription_data` to allow overriding the default subscription data.
+		 */
+		$new_subscription_data = function ( $args ) use ( $default_subscription_data ) {
+			$args = wp_parse_args( $default_subscription_data, $args );
```
james-allan: If I understand this correctly the `wp_parse_args( $default_subscription_data, $args )` here will mean the status of every subscription created will be `pending`. 

I don't think we'll be able to use `wp_parse_args` in this case because there will always be a status by the time the `woocommerce_new_order_data` hook runs. This is because the parent create function will generate a default status via the `$this->get_post_status()` here.  [abstract-wc-order-data-store-cpt.php#L88 ](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/data-stores/abstract-wc-order-data-store-cpt.php#L88)

With that in mind I think there are 2 possible solutions here: 

1. Introduce a `get_post_status()` to the subscriptions datastore which basically is a replicate of the order one. But instead it will trigger the `woocommerce_default_subscription_status` status.
2. At the top of this create function make the status check ourselves. see example below. 

```php
$subscription_status = $order->get_status( 'edit' );

if ( ! $subscription_status ) {
	$subscription->set_status( 'wc' . apply_filters( 'woocommerce_default_subscription_status', 'pending' );
}
```

wcs-functions.php
```diff
@@ -149,62 +152,32 @@ function wcs_create_subscription( $args = array() ) {
...
-	$subscription_data['post_author']   = 1;
-	$subscription_data['post_password'] = uniqid( 'order_' );
-	// translators: Order date parsed by strftime
-	$post_title_date = strftime( _x( '%b %d, %Y @ %I:%M %p', 'Used in subscription post title. "Subscription renewal order - <this>"', 'woocommerce-subscriptions' ) ); // phpcs:ignore WordPress.WP.I18n.UnorderedPlaceholdersText
-	// translators: placeholder is order date parsed by strftime
-	$subscription_data['post_title']    = sprintf( _x( 'Subscription &ndash; %s', 'The post title for the new subscription', 'woocommerce-subscriptions' ), $post_title_date );
-	$subscription_data['post_date_gmt'] = $args['date_created'];
-	$subscription_data['post_date']     = get_date_from_gmt( $args['date_created'] );
+	$subscription = new \WC_Subscription();
+	$subscription->set_status( $args['status'] );
```
james-allan: It's possible that the status isn't actually passed in and so this could be `''`. We should do something similar to how `wc_create_order` handle the setting by only calling the setters if it's set and it validates.

wcs-functions.php
```diff
@@ -149,62 +152,32 @@ function wcs_create_subscription( $args = array() ) {
...
+	$subscription->set_customer_note( $args['customer_note'] ?? '' );
+	$subscription->set_customer_id( $args['customer_id'] );
+	$subscription->set_date_created( $args['date_created'] );
+	$subscription->set_created_via( $args['created_via'] );
+	$subscription->set_version( $args['order_version'] );
+	$subscription->set_currency( $args['currency'] );
+	$subscription->set_prices_include_tax( $args['prices_include_tax'] );
+	$subscription->set_billing_period( $args['billing_period'] );
+	$subscription->set_billing_interval( absint( $args['billing_interval'] ) );
+	$subscription->set_schedule_start( $args['start_date'] );
```
james-allan, line 178: I assume you had to introduce this function because you needed a way to set the start date without calling `$subscription->update_dates()` without saving it. Is that correct?

includes/data-stores/class-wcs-subscription-data-store-cpt.php
james-allan: The second option looks cleaner to me and doesn't involve copying a whole function. 1 thing to keep in mind though is that we'll need to make sure that calling `set_status()` doesn't trigger status changed events/hooks. Given the previous status will be empty, it shouldn't.

james-allan:
I've started testing this PR from the frontend. So far I've only tested this using the wp post infrastructure and compared the subscription data with this PR vs `trunk`. That check mostly passed. https://www.diffchecker.com/MUZCLize. The differences were on the `_billing_address_index` and `_subscription_switch_data` fields. I don't know what would cause that in this PR though. I'll continue testing tomorrow.

wcs-functions.php
brucealdridge: All these setters were already in the function, they have just been reordered. 

https://github.com/Automattic/woocommerce-subscriptions-core/blob/8017438f8e0cfc16494344f783fb134f77284d58/wcs-functions.php#L192-L199

james-allan: Oh yeah I was referring to the `set_schedule_start()` function specifically. That function was introduced in this PR and I was curious as to why it was added instead of using the `update_dates()` function. eg `$subscription->update_dates( [ 'start' => $args['start_date'] ] )`.

I think it was because `update_dates()` actually saves the dates to the database whereas adding this  `set_schedule_start` function just sets the start date meta and it is all saved at once. 

I think this is a good idea, I was just curious if that was the reason behind it.

james-allan:
> The differences were on the _billing_address_index and _subscription_switch_data fields. I don't know what would cause that in this PR though. I'll continue testing tomorrow.

I've discovered what is causing the billing and shipping index issues. I'll file a separate issue for this given the indexes are only used searching and aren't critical. 

Issue: https://github.com/Automattic/woocommerce-subscriptions-core/issues/227

wcs-functions.php
brucealdridge: I've wrapped `set_status()` in an `if`

includes/data-stores/class-wcs-subscription-data-store-cpt.php
```diff
@@ -94,10 +94,28 @@ public function __construct() {
...
+		}
+
+		/**
+		 * This function is called on the `woocommerce_new_order_data` filter.
+		 * We hook into this function, calling our own filter `woocommerce_new_subscription_data` to allow overriding the default subscription data.
+		 */
+		$new_subscription_data = function ( $args ) {
+			return apply_filters( 'woocommerce_new_subscription_data', $args );
+		};
+		add_filter( 'woocommerce_new_order_data', $new_subscription_data );
```
james-allan, line 111: This closure, `add_filter()` and the following `remove_filter()` can all be removed now given the `$args` variable is undefined. :)

james-allan (changes requested):
I've left 1 small request inline. Also I hope you don't mind but I submitted 5 small commits to make uncontroversial changes. The commits are these: 

-  7afb346f0bf1becbb0a82077f8eb469e244a0d0a & d0ad5ccc7037fb2c87da0c91774d8ec1c469f46c - Replaces order references in the `WC_Subscription` contructor. I also just reworded it to read better for a constructor.
- d36486a2ab4628af75542f2dc0e79959d9648877 - Removes an empty line added as part of this PR. 
- f75527c354cd4106b90c40a147907f0eaad396a0 - adds whitespace for what would be phpcs violations except for the fact that the line is ignored. 
- d73dfcd8f9bc12f0569b2528dccb32467a5da662 - Adds/removes some whitespace and applys sentence case to `wcs_create_subscription()` inline with [WP coding standards](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#5-1-single-line-comments).

wcs-functions.php
```diff
@@ -149,62 +154,36 @@ function wcs_create_subscription( $args = array() ) {
...
 	 * Filter the newly created subscription object.
 	 *
 	 * @since 2.2.22
 	 * @param WC_Subscription $subscription
 	 */
-	$subscription = apply_filters( 'wcs_created_subscription', wcs_get_subscription( $subscription_id ) );
+	$subscription = apply_filters( 'wcs_created_subscription', $subscription );
+
+	// we need to fetch the subscription from the database as the current object state doesn't match the loaded state.
+	$subscription = wcs_get_subscription( $subscription );
```
james-allan: This `wcs_get_subscription( $subscription );` will need to return to its original place in the `apply_filters` above.


```
$subscription = apply_filters( 'wcs_created_subscription', wcs_get_subscription( $subscription_id ) );
```

This is because if a 3rd party hooks onto the filter and alters the subscription in some way, any changes they made would be overridden by the `wcs_get_subscription()`.

includes/data-stores/class-wcs-subscription-data-store-cpt.php
james-allan: Oh nvm I misread this. This is still necessary because it triggers the `woocommerce_new_subscription_data` hook when creating the subscription via the `parent::create`

james-allan:
These changes are working well for me in my testing locally.

I've tested purchasing a subscription using Stripe, a manual payment method (bacs) and Stripe with manual payments forced (see screenshot below). In all my tests the subscriptions look complete in the database and compatible with how the data looked on `trunk`. 

Screen Shot 2022-10-25 at 4 01 54 pm

I've left a small request regarding backwards compatibility with the `wcs_created_subscription` filter, but other than that and this known issue ([**HPOS: Make sure billing and shipping address indexes are set on checkout #227**](https://github.com/Automattic/woocommerce-subscriptions-core/issues/227)), this looks good.

mattallan:
I've ran through some compatibility tests on these changes:

- [x] PayPal Checkout - Purchasing Subscription and processing a renewal
  - [x] Purchased subscription on `trunk` and this branch and used James' and Bruce's queries shared in slack to compare postmeta
- [x] AutomateWoo - Simple subscription created workflow (updated status and added customer note on this event)
- [x] Confirmed `subscription.created` webhook payload sent
- [x] Checked any third-party plugins aren't affected by any new or changed hooks
   - [x] No third-party plugins on my local site (30+ WC Extensions) are hooking onto the `woocommerce_new_order_data` filter and unexpectedly setting subscription data (this is a new hook that is called when creating a subscription now that we're calling the Orders CPT's `create()` method)

All my checks passed and LGTM :+1:

includes/data-stores/class-wcs-subscription-data-store-cpt.php
Jinksi: Is this hook just used for unit testing?

wcs-functions.php
```diff
@@ -114,13 +113,19 @@ function wcs_create_subscription( $args = array() ) {
...
 
-	$args              = wp_parse_args( $args, $default_args );
-	$subscription_data = array();
+	if ( isset( $args['order_version'] ) ) {
+		wcs_deprecated_argument( __FUNCTION__, '2.0', 'The "order_version" argument is no longer changeable.' );
+	}
+
+	$args = wp_parse_args( $args, $default_args );
+
+	if ( ! empty( $args['status'] ) && ! array_key_exists( 'wc-' . $args['status'], wcs_get_subscription_statuses() ) ) {
```
Jinksi, line 130: Nitpick: It might be nice to add a comment like `// Validate the status arg.` for consistency with comments below.

wcs-functions.php
```diff
@@ -134,77 +139,53 @@ function wcs_create_subscription( $args = array() ) {
...
-	}
+	$subscription = new \WC_Subscription();
 
-	// Only set the status if creating a new subscription, use wcs_update_subscription to update the status
 	if ( $args['status'] ) {
-		if ( ! in_array( 'wc-' . $args['status'], array_keys( wcs_get_subscription_statuses() ) ) ) {
-			return new WP_Error( 'woocommerce_invalid_subscription_status', __( 'Invalid subscription status given.', 'woocommerce-subscriptions' ) );
-		}
-		$subscription_data['post_status']  = 'wc-' . $args['status'];
+		$subscription->set_status( $args['status'] );
```
Jinksi, line 167: Suggest keeping the comment unless it's irrelevant (it's helping me understand what's happening here 😝)
`// Only set the status if creating a new subscription, use wcs_update_subscription to update the status`

wcs-functions.php
```diff
@@ -134,77 +139,53 @@ function wcs_create_subscription( $args = array() ) {
...
 
-	// check customer id is set
+	// Check customer id is set.
 	if ( empty( $args['customer_id'] ) || ! is_numeric( $args['customer_id'] ) || $args['customer_id'] <= 0 ) {
 		return new WP_Error( 'woocommerce_subscription_invalid_customer_id', _x( 'Invalid subscription customer_id.', 'Error message while creating a subscription', 'woocommerce-subscriptions' ) );
 	}
 
-	// check the billing period
+	// Check the billing period.
 	if ( empty( $args['billing_period'] ) || ! in_array( strtolower( $args['billing_period'] ), array_keys( wcs_get_subscription_period_strings() ) ) ) {
```
Jinksi:
```suggestion
	if ( empty( $args['billing_period'] ) || ! in_array( strtolower( $args['billing_period'] ), array_keys( wcs_get_subscription_period_strings() ), true ) ) {
```
PHPCS warning fix: Add third argument (`$strict = true`) to `in_array` to fix `StrictInArray.MissingTrueStrict` (Not using strict comparison for in_array; supply true for third argument).

Jinksi:
There is a PHPCS error that we can fix in `wcs_create_subscription`:

```
Don't use current_time() for retrieving a Unix (UTC) timestamp. Use time() instead. Found: current_time( 'timestamp', true )
```

https://github.com/Automattic/woocommerce-subscriptions-core/blob/1a56b9fd893ced4866b332d67ce2b5d3c3518775/wcs-functions.php#L133

```diff
- 	} elseif ( wcs_date_to_time( $args['date_created'] ) > current_time( 'timestamp', true ) ) {
+	} elseif ( wcs_date_to_time( $args['date_created'] ) > time() ) {
```

These kinds of phpcs issues will be recurring as we continue refactoring. It might be easier if I commit these fixes to this PR or future PRs, leaving a self-review comment. What do you think?

**Update**: Fixed in ae559089

wcs-functions.php
Jinksi: Fixed in ae559089

wcs-functions.php
Jinksi: Fixed in ae559089 using the even more awesome `array_key_exists` function

wcs-functions.php
Jinksi: Resolved in eede953fbe6ed9f8bcab26559db60ce0c16bff89

wcs-functions.php
```diff
@@ -99,24 +99,27 @@ function wcs_create_subscription( $args = array() ) {
...
 	);
 
+	// If we are creating a subscription from an order, we use some of the order's data as defaults.
+	if ( $order instanceof \WC_Order ) {
+		$default_args['customer_id']        = $order->get_user_id();
+		$default_args['created_via']        = $order->get_created_via( 'edit' );
+		$default_args['currency']           = $order->get_currency( 'edit' );
+		$default_args['prices_include_tax'] = $order->get_prices_include_tax( 'edit' ) ? 'yes' : 'no';
+		$default_args['date_created']       = wcs_get_datetime_utc_string( $order->get_date_created( 'edit' ) );
+	}
```
brucealdridge, line 121: Rather than having lots of ternary operators, I decided to split this out to mark it a lot more readable while also removing the references to wcs_get_objects_property.

Jinksi (commented):
LGTM @brucealdridge 🙌 

Thanks for the descriptive comments throughout 🏅

I conducted some basic backwards-compatibility smoke-testing with HPOS disabled on a JN site with WC 7.1.0-beta-2:
- [x] WCPay 4.9.0 subscription purchase
- [x] WCPay 4.9.0 + WCS 4.6.0 subscription purchase + renewal

james-allan (approved):

PR #237, opened 2022-11-01
Fixes issue caused by wcs_copy_order_address() using new WC Order APIs that aren't available in versions of WC we support

### LINKED ISSUE #236: Uncaught Error: Call to undefined method WC_Subscription::set_shipping_address()
## Describe the bug

WooCommerce core is introducing a new function to `WC_Order`, `set_shipping_address` 
 to version 7.1. 
This new function was used in a recent PR (#204) which caused regressions on flows using older versions of WooCommerce.

## To Reproduce

1. Create any test site using the JN site.
1. Install and activate all the required plugins.
1. Install and activate WCPay Dev plugin.
1. Go to WCPay Dev.
1. Check the following checkboxes:
    Enable grouped settings
    Enable UPE checkout
    Add UPE additional payment methods
    Enable WCPay subscriptions
1. Click the Submit button.
1. Go to Products → Add new
1. Create Simple subscription product.
1. Go to shop Page.
1. Add subscription product in cart.
1. Go to Checkout Page.
1. Fill address and card details. 
1. Click on Sign up Now Button.
1. Observe that, "Error processing checkout. Please try again." notice display.

### Expected behavior
"Error processing checkout. Please try again." notice displayed after checking out with Simple subscription product on "Checkout" page.

### Expected Result:
Order should be placed without any error after checking out with Simple subscription product on "Checkout" page.

## Product impact

- [x] Does this issue affect WooCommerce Payments? yes https://github.com/Automattic/woocommerce-payments/issues/5033

mattallan:
Fixes #236

## Description

In #228 we updated our `wcs_copy_order_address()` function to use CRUD methods, however, the functions we switched over to using were only just introduced in 7.1 (which is still in release candidate stages of development and hasn't been released yet).

This PR updates our `wcs_copy_order_address()` to use single address setter functions which were introduced in WC 3.0.
In these changes I've also added a `$to_order->save()` to fix a backward compatibility issue where the refactored version of this function wasn't writing the changes to the database.

I thought about adding an extra "save_address" parameter to this function which defaults to true and saves the address but not sure if we'd see much use for not wanting to save the address. Keen to get others thoughts though.

## How to test this PR

1. Install WC 6.8 (earliest supported version in WC Pay)
2. Try to purchase a subscription on the checkout.
3. On `trunk` you will see error: `Fatal error: Uncaught Error: Call to undefined method WC_Subscription::set_shipping_address()
in /srv/users/userf10dd6d1/apps/userf10dd6d1/public/wp-content/plugins/woocommerce-payments/vendor/woocommerce/subscriptions-core/includes/wcs-order-functions.php on line 99`
4. On this branch, you shouldn't see any issues when purchasing a subscription.
5. On this branch, also confirm the address data has been correctly copied from order to subscription.

## Product impact

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Subscriptions Core Changelog ***
 
+= 2.4.1 - 2022-xx-xx =
+* Fix - Undefined method WC_Order::set_shipping_address() on stores running pre-7.1 of WooCommerce which prevented subscriptions from being purchased.
+
 = 2.4.0 - 2022-10-28 =
 * Update - The subscription creation function `wcs_create_subscription` has been updated to use WooCommerce CRUD methods in preparation for supporting High Performance Order Storage (HPOS).
 * Update - Improve wcs_copy_order_address() to use modern APIs for setting address fields.
--- a/includes/wcs-order-functions.php
+++ b/includes/wcs-order-functions.php
@@ -96,37 +96,29 @@ function wcs_get_subscriptions_for_order( $order, $args = array() ) {
 function wcs_copy_order_address( $from_order, $to_order, $address_type = 'all' ) {
 
 	if ( 'all' === $address_type || 'shipping' === $address_type ) {
-		$to_order->set_shipping_address(
-			array(
-				'first_name' => $from_order->get_shipping_first_name( 'edit' ),
-				'last_name'  => $from_order->get_shipping_last_name( 'edit' ),
-				'company'    => $from_order->get_shipping_company( 'edit' ),
-				'address_1'  => $from_order->get_shipping_address_1( 'edit' ),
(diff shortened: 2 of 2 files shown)


james-allan (approved):
Here's a list of checks I ran: 

- I've confirmed the issue by running `trunk` on with WC 6.8.2 and purchasing a subscription.
- I've confirmed your changes fix that error.
- I've compared a subscription purchased on your branch in WC 6.8.2 with a subscription purchased on `trunk`  in 7.1 and the meta matched.
- I've enabled shipping and checked that the shipping address fields is saved in meta with this branch.
- I did code review to make sure the setter and getters line up. 

From a purely code change, this has little to no impact because the previous approach of calling `set_shipping_address()` just looped over all the key values and called `$order->set_address_prop()`. Calling the individual setters (eg `set_billing_first_name()`) instead does the exact same thing by calling  `$order->set_address_prop()`.

PR #246, opened 2022-11-04
Update our related orders cache class to not use deprecated renewal meta query hooks

mattallan:
## Description

After tagging a 2.5.0 version of Subscriptions Core, I realized we were using filters that were recently deprecated in James' https://github.com/Automattic/woocommerce-subscriptions-core/pull/229 PR. This PR updates our `remove_related_order_cache_keys` function to:
 - be attached to the new updated hook `wc_subscriptions_renewal_order_data`
 - remove the cache meta keys from the updated formatted array

## How to test this PR

1. To test this PR I used a similar snippet of code to: https://gist.github.com/mattallan/35a59d6c0f67e2c84c45eeb0425cfb6e
2. On processing a renewal, I confirmed the following metadata isn't copied from subscription to the renewal order:

```
'_subscription_renewal_order_ids_cache',
'_subscription_resubscribe_order_ids_cache',
'_subscription_switch_order_ids_cache',
```

## Product impact

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Subscriptions Core Changelog ***
 
+= 2.5.1 - 2022-xx-xx =
+* Dev - Replace the use of the deprecated wcs_renewal_order_meta hook with wc_subscription_renewal_order_data in the WCS_Related_Order_Store_Cached_CPT class.
+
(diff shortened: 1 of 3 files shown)


james-allan (approved):
Changes look good to me. Thanks @mattallan!

Thanks for also including the fix to a copy paste error. :)

Issue #239, opened 2022-11-02
Update subscriptions-core version


brucealdridge:
## Description

subscriptions-core code and APIs are currently tightly coupled to WC Subscriptions version numbers (e.g. `@since`, `@deprecated`.) This tight coupling is problematic: subscriptions-core is used and released independently of WC Subscriptions, and currently has a lower version number than WC Subscriptions. We need to explicitly version the code in subscriptions-core as it is no longer coupled to WC S (or WCPay).

The next version that is released on subscriptions-core needs to be greater than the version of WooCommerce Subscriptions (currently 4.6.0)

- [x] Update subscriptions-core library to 5.0.0
- [x] #240
- [x] #241
- [x] Update any WC Pay to use get_library_version instead of get_plugin_version - https://github.com/Automattic/woocommerce-payments/pull/5227
- [x] Update any WC S to use get_library_version instead of get_plugin_version - https://github.com/woocommerce/woocommerce-subscriptions/pull/4435   
- [x] #242
- [x] #243

mattallan:
All parts to this issue have been completed. Closing it.

PR #254, opened 2022-11-07
Update naming of library version getter.

### LINKED ISSUE #240: Add a new method `WC_Subscriptions_Core_Plugin::get_library_version()`.
### The Problem
subscriptions-core code and APIs are currently tightly coupled to WC Subscriptions version numbers (e.g. `@since`, `@deprecated`.) This tight coupling is problematic: subscriptions-core is used and released independently of WC Subscriptions, and currently has a lower version number than WC Subscriptions. We need to explicitly version the code in subscriptions-core as it is no longer coupled to WC S (or WCPay).

Currently for version detection we use `WC_Subscriptions_Core_Plugin::get_plugin_version()` however this is not clear as subscriptions-core is not a plugin but a library.

### Solution
Add new function `WC_Subscriptions_Core_Plugin::get_library_version()` to return the version of the subscriptions-core library and replace usages of `get_plugin_version` with `get_library_version`

### LINKED ISSUE #241: Deprecate `WC_Subscriptions_Core_Plugin::get_plugin_version()`
### The Problem
subscriptions-core code and APIs are currently tightly coupled to WC Subscriptions version numbers (e.g. `@since`, `@deprecated`.) This tight coupling is problematic: subscriptions-core is used and released independently of WC Subscriptions, and currently has a lower version number than WC Subscriptions. We need to explicitly version the code in subscriptions-core as it is no longer coupled to WC S (or WCPay).

Currently for version detection we use `WC_Subscriptions_Core_Plugin::get_plugin_version()` however this is not clear as subscriptions-core is not a plugin but a library.

### Solution

This is partly solved by #240 however we need to mark `get_plugin_version` as deprecated.

brucealdridge:
Fixes #240 
Fixes #241 

## Description

For version detection, we use WC_Subscriptions_Core_Plugin::get_plugin_version() and there was confusion around whether this referred to the version of WC Subscriptions plugin (or equivalent) or the version of the subscriptions-core library.
After much discussion, we have decided to remove references to the WC Subscription version. 

This PR introduces a soft `@deprecation` docblock comment for `WC_Subscriptions_Core_Plugin::get_plugin_version()` This function has been recommended by use on many plugins and cannot be removed easily in the short-term.
A replacement function `WC_Subscriptions_Core_Plugin::get_library_version()` has been added which is recommended for future use.

`WC_Subscriptions_Core_Plugin::get_library_version()` is setup to be bumped automatically on every release.

`WC_Subscriptions_Core_Plugin::get_plugin_version()` was previously set to return `3.1.6`. This version is used for migrations and for checking against features introduced in previous versions. Because of this we must only increment this version so this will require a bumping of the library version as noted in #239 

## How to test this PR

1. Go to …
2. Click on …

## Product impact


mattallan:
Hey @brucealdridge just a heads up that it looks like there are some conflicts with this PR and trunk caused by latest docblock changes :)

package.json
```diff
@@ -8,7 +8,8 @@
 	"homepage": "https://github.com/Automattic/woocommerce-subscriptions-core",
 	"main": "Gruntfile.js",
 	"config": {
-		"translate": true
+		"translate": true,
+		"version_replace_paths": ["includes/"]
```
aprea, line 12: Is this change necessary given the additions in `bin/update-version.php` to manually bump the version number in `includes/class-wc-subscriptions-core-plugin.php`?

aprea (approved):
This LGTM ✅ I tested this locally with `php bin/update-version.php 7.0.0` and I observed the library version was updated as expected.

aprea:
FYI I sent a small commit with some tidying 3357cc0832d0ae92a80cce219cec27f54c530a28

package.json
brucealdridge: this allows usage of `@version x.x.x` and `@since x.x.x` in the code and `update-version.php` will fill these in on release.

aprea: Ah gotcha, thanks!

PR #264, opened 2022-11-15
Support for Collection in Block Checkout

mikejolley:
New Local Pickup functionality is being built in blocks: https://github.com/woocommerce/woocommerce-blocks/pull/7433

With this selected, packages are not shown individually which break Subscriptions. The current workaround is to support `woocommerce_store_api_cart_select_shipping_rate` and set the rate ID for all recurring packages when a package ID is not given.

cc @senadir 

## How to test this PR

Requires the https://github.com/woocommerce/woocommerce-blocks/pull/7433 branch.

1. Add various products and subscriptions to cart.
2. Go to checkout.
3. Choose "local pickup"
4. Select a pickup location
5. Confirm the selected pickup location applies to the cart and all recurring carts

## Product impact

### DIFF
--- a/includes/class-wc-subscriptions-cart.php
+++ b/includes/class-wc-subscriptions-cart.php
@@ -105,6 +105,7 @@ class WC_Subscriptions_Cart {
 		// When WooCommerce determines the taxable address only return pick up shipping methods chosen for the recurring cart being calculated.
 		add_filter( 'woocommerce_local_pickup_methods', __CLASS__ . '::filter_recurring_cart_chosen_shipping_method', 100, 1 );
 		add_filter( 'wc_shipping_local_pickup_plus_chosen_shipping_methods', __CLASS__ . '::filter_recurring_cart_chosen_shipping_method', 10, 1 );
+		add_action( 'woocommerce_store_api_cart_select_shipping_rate', __CLASS__ . '::set_recurring_cart_chosen_shipping_method', 10, 2 );
 
 		// Validate chosen recurring shipping methods
 		add_action( 'woocommerce_after_checkout_validation', __CLASS__ . '::validate_recurring_shipping_methods' );
@@ -1125,6 +1126,30 @@ class WC_Subscriptions_Cart {
 		}
 
 		return $shipping_methods;
+	}
+
+	/**
+	 * Updates all packages in the recurring cart to use a given shipping method if requested via Store API.
+	 * 
+	 * @see woocommerce_store_api_cart_select_shipping_rate
+	 *
+	 * @param string|null $package_id The sanitized ID of the package being updated. Null if all packages are being updated.
+	 * @param string $rate_id The sanitized chosen rate ID for the package.
+	 */
+	public static function set_recurring_cart_chosen_shipping_method( $package_id, $rate_id ) {
+		if ( ! is_null( $package_id ) ) {
+			return;
+		}
+		$chosen_shipping_methods = wc()->session->get( 'chosen_shipping_methods' ) ? wc()->session->get( 'chosen_shipping_methods' ) : [];
+		
+		foreach ( WC()->cart->recurring_carts as $recurring_cart_key => $recurring_cart ) {
+			foreach ( $recurring_cart->get_shipping_packages() as $package_index => $recurring_cart_package ) {
+				$recurring_cart_package_key = self::get_recurring_shipping_package_key( $recurring_cart_key, $package_index );
+				$chosen_shipping_methods[ $recurring_cart_package_key ] = $rate_id;
+			}
+		}
+
+		wc()->session->set( 'chosen_shipping_methods', $chosen_shipping_methods );
 	}
 
 	/**
--- a/includes/class-wc-subscriptions-cart.php
+++ b/includes/class-wc-subscriptions-cart.php
@@ -1130,7 +1130,7 @@ class WC_Subscriptions_Cart {
 
 	/**
 	 * Updates all packages in the recurring cart to use a given shipping method if requested via Store API.
-	 * 
+	 *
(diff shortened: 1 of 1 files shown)


haszari:
Thanks @mikejolley  - requesting a review from @Automattic/helix

Jinksi:
@mikejolley in the testing steps, you mention "Select a pickup location". I'm assuming this is a new option included in the checkout block from the https://github.com/woocommerce/woocommerce-blocks/pull/7433 branch, but I can't find where to enable this.

senadir:
Hey @Jinksi , you can add new Local Pickup rates from the Local Pickup tab in WooCommerce -> Settings -> Shipping

`wp-admin/admin.php?page=wc-settings&tab=shipping&section=pickup_location`

Jinksi:
Thanks @senadir. I was missing this wp-config variable:

`define('JETPACK_AUTOLOAD_DEV', true);`

Jinksi:
Also, it turns out I had an older version (name/slug?) of the repo (`woocommerce-gutenberg-products-block`) that prevented building the file `wc-shipping-method-pickup-location.js` for some reason.

Cloning the `woocommerce-blocks` repo fresh did the trick.

image

includes/class-wc-subscriptions-cart.php
```diff
@@ -105,6 +105,7 @@ public static function init() {
 		// When WooCommerce determines the taxable address only return pick up shipping methods chosen for the recurring cart being calculated.
 		add_filter( 'woocommerce_local_pickup_methods', __CLASS__ . '::filter_recurring_cart_chosen_shipping_method', 100, 1 );
 		add_filter( 'wc_shipping_local_pickup_plus_chosen_shipping_methods', __CLASS__ . '::filter_recurring_cart_chosen_shipping_method', 10, 1 );
+		add_action( 'woocommerce_store_api_cart_select_shipping_rate', __CLASS__ . '::set_recurring_cart_chosen_shipping_method', 10, 2 );
```
Jinksi, line 108: This hook `woocommerce_store_api_cart_select_shipping_rate` isn't executing for me. I'm not sure what I'm missing.

I attempted searching for it with no luck on WC/WC Blocks. Is it a dynamic hook?

My env:
WP 6.1
WC 7.1.0 (also tried trunk)
WC Blocks branch [add/local-pickup-tax-locations](https://github.com/woocommerce/woocommerce-blocks/tree/add/local-pickup-tax-locations) (tried `npm run build` and `npm run dev`)
WC Subs 4.6.0
`define('JETPACK_AUTOLOAD_DEV', true);` in `wp-config.php`

mikejolley: It should be here now if you want to checkout again: https://github.com/woocommerce/woocommerce-blocks/pull/7433/files#diff-36db0c99c9ec11b21a09848d8335af16df6aa687f32414484df6363b4e707051R97

shendy-a8c:
I have some questions after testing this PR.

Here's my setup:
- I created a checkout block page by going to wp-admin > Pages > Add New > Choose `Checkout`.

Screenshot 2022-11-23 at 02 59 23

- I created a Shipping Zone from wp-admin > WooCommerce > Settings > Shipping > Add shipping zone

- Added `define('JETPACK_AUTOLOAD_DEV', true);` to wp-config.php per @Jinksi's [comment](https://github.com/Automattic/woocommerce-subscriptions-core/pull/264#issuecomment-1319503055). Very useful. Thanks!

- Have WooCommerce Subscriptions active and WooCommerce Subscriptions Core installed as a plugin.

Then, here's the step on how I tested this PR:
- Purchased a simple product and a subscription product in one cart.
- Make sure `WC_Subscriptions_Cart::set_recurring_cart_chosen_shipping_method()` is called during the checkout process.
- Edit the subscription and choose `Process renewal` from the action box on the right.
- Make sure the Local Pickup shipping fee is present in the renewal order.

Everything seems to be working but I found problems and I have some questions:
- What's the difference behavior I should be expecting here between `trunk` and this PR's branch, `add/support-for-local-pickup-in-blocks`? When I use `trunk`, I also see the Local Pickup shipping fee carried correctly to the subscription renewal order.

- `woocommerce-blocks`'s `add/local-pickup-tax-locations` from https://github.com/woocommerce/woocommerce-blocks/pull/7433 is older version compared to the production version. Is this expected?
This is what I see from my plugins list page.
Screenshot 2022-11-23 at 03 10 17

- The checkout process using checkout block is broken. See my video below. Basically, somehow shipping address is not asked and the checkout request got an error back because shipping address is empty. I don't see the problem with classic checkout page. Also, phone is required in classic checkout but not in checkout block.


https://user-images.githubusercontent.com/73803630/203412095-ef5665bf-b6be-44b5-a954-6607b9ba1e79.mov

shendy-a8c:
I also tested when WooCommerce Subscriptions is not active, so we will create a WCPay Subscription and renewal needs to be manual where customer needs to go through the checkout process again to complete the renewal process. I suspect that this PR is dealing with that flow because of the 'recurring cart' mentioned.

However, I see that Local Pickup is not selected. Is that intentional?

See my video below:


https://user-images.githubusercontent.com/73803630/203413594-785970cd-0787-410a-9658-21bc7ac2607e.mov

brucealdridge:
Removing Helix review as awaiting author reply.
Are you able to look into this @mikejolley then re-request review?

mikejolley:
Local Pickup has now merged into Blocks trunk so there is no need to switch to special branches. All you need to do is enable Local Pickup in your shipping settings and have at least one other shipping method active (flat rate), add a subscriptions product and a regular product to your cart, and then go to checkout.

This solution will no longer work however, so I've created an issue with the details: 

https://github.com/Automattic/woocommerce-subscriptions-core/issues/399

PR #266, opened 2022-11-16
Version 2.5.2

mattallan:
```
= 2.5.2 - 2022-11-15 =
* Fix - When creating a subscription via the checkout, make sure a new instance of the subscription is attached to the `woocommerce_checkout_subscription_created` action hook.
```

---

### Why is this version 2.5.2 and not 5.0.1?

We discovered a bug in WooCommerce Payments 5.0 which shipped with `subscriptions-core` v2.5.1 which resulted in us releasing a new patch of WC Payments 5.0.3.

Subscriptions Core 5.0.0 has a lot of changes that were too risky to release as part of this patch release so
I decided to tag a new 2.5.2 which is a patch release on the latest subscriptions core version shipped in WC Pay 5.0


mattallan:
No version bumps needed to merge this into trunk. Only the fix and changelog is necessary since we've already tagged a 5.0.0 of subscriptions-core

This PR is just to get the 2.5.2 changes into `trunk`

mattallan:
I've confirmed the changes in this PR are only adding the changelog and the fix introduced with https://github.com/Automattic/woocommerce-subscriptions-core/pull/263 therefore I'm going to self merge this one

Issue #221, opened 2022-10-13
Replace _post hooks


brucealdridge:
## Describe the bug

There are a number of places in the code that hook into the `post` actions. These methods will not work with [HPOS](https://developer.woocommerce.com/2022/09/14/high-performance-order-storage-progress-report/) turned on and syncing to CPT turned off. There is already an issue to [Replace 'save_post' hooks](https://github.com/Automattic/woocommerce-subscriptions-core/issues/220) but there are other hooks we will need to replace such as `before_delete_post` `wp_trash_post` `deleted_post` `trashed_post`

A sample "find in files" regex search: `'(\w+)_post'`

Examples:
https://github.com/Automattic/woocommerce-subscriptions-core/blob/3c7ae3489cacd2bf29ea59daf8702c8a914aafc8/includes/class-wc-subscriptions-checkout.php#L92
https://github.com/Automattic/woocommerce-subscriptions-core/blob/3c7ae3489cacd2bf29ea59daf8702c8a914aafc8/includes/class-wc-subscriptions-manager.php#L51
https://github.com/Automattic/woocommerce-subscriptions-core/blob/3c7ae3489cacd2bf29ea59daf8702c8a914aafc8/includes/class-wc-subscriptions-manager.php#L57

mattallan:
Thanks Bruce for opening this issue.

There are a few of these in the `WCS_Post_Meta_Cache_Manager` class as well but I've opened a separate issue to tackle that class as a whole: https://github.com/Automattic/woocommerce-subscriptions-core/issues/194


I wonder if there are CRUD-type replacement hooks we can replace these with that will work for Subscriptions stored in both Posts or Order Tables, or whether we have to surround these hooks with a `if (HPOS is not enabled); do _post hooks` condition 🤔

mattallan:
Some details about these delete/trash hooks in slack: p1659617204621119-slack-C9QDV4TL4

Question raised:
[...]

PR #305, opened 2022-11-24
Remove all get_post_type() from WC_Subscriptions_Manager.

### LINKED ISSUE #288: [HPOS] Replace `get_post_type()` with WC_Data_Store::load( 'subscription' )->get_order_type( $id )

shendy-a8c:
Partially fixes https://github.com/Automattic/woocommerce-subscriptions-core/issues/288.

## Description

Replace `get_post_type()` usage from `WC_Subscriptions_Manager`.

## How to test this PR

Not sure what's the best way to test but I use WP Console to call the affected function directly, eg `WC_Subscriptions_Manager::maybe_cancel_subscription( $id )`.
I tested with $id that belongs to a subscription and one that does not. Also, I tested with HPOS disabled and enabled and with WC version 6.8 (prior HPOS).

Note: I haven't added a changelog. Will add one after 5.1 is released.

## Product impact

### DIFF
--- a/includes/class-wc-subscriptions-manager.php
+++ b/includes/class-wc-subscriptions-manager.php
@@ -841,20 +841,29 @@ class WC_Subscriptions_Manager {
 	/**
 	 * Make sure a subscription is cancelled before it is trashed or deleted
 	 *
-	 * @param int $post_id
+	 * @param int $id
 	 * @since 1.0.0 - Migrated from WooCommerce Subscriptions v2.0
 	 */
-	public static function maybe_cancel_subscription( $post_id ) {
+	public static function maybe_cancel_subscription( $id ) {
 
-		if ( 'shop_subscription' == get_post_type( $post_id ) && 'auto-draft' !== get_post_status( $post_id ) ) {
+		$subscription = wcs_get_subscription( $id );
+		if ( ! $subscription ) {
+			return;
+		}
 
-			$subscription = wcs_get_subscription( $post_id );
+		// Check its type because $subscription might be WC_Subscription's child class.
+		if ( $subscription->get_type() !== 'shop_subscription' ) {
+			return;
+		}
 
-			if ( $subscription->can_be_updated_to( 'cancelled' ) ) {
+		if ( 'auto-draft' === $subscription->get_status() ) {
+			return;
+		}
 
-				$subscription->update_status( 'cancelled' );
+		if ( $subscription->can_be_updated_to( 'cancelled' ) ) {
 
-			}
+			$subscription->update_status( 'cancelled' );
+
 		}
 	}
 
@@ -868,12 +877,22 @@ class WC_Subscriptions_Manager {
 	 * This function fixes that by setting '_wp_trash_meta_status' to 'wc-cancelled' whenever its former status
 	 * is something that can not be restored.
 	 *
-	 * @param int $post_id
+	 * @param int $id
 	 * @since 1.0.0 - Migrated from WooCommerce Subscriptions v2.0
 	 */
-	public static function fix_trash_meta_status( $post_id ) {
+	public static function fix_trash_meta_status( $id ) {
 
-		if ( 'shop_subscription' == get_post_type( $post_id ) && ! in_array( get_post_meta( $post_id, '_wp_trash_meta_status', true ), array( 'wc-pending', 'wc-expired', 'wc-cancelled' ) ) ) {
+		$subscription = wcs_get_subscription( $id );
+		if ( ! $subscription ) {
+			return;
+		}
+
+		// Check its type because $subscription might be WC_Subscription's child class.
+		if ( $subscription->get_type() !== 'shop_subscription' ) {
+			return;
+		}
+
+		if ( ! in_array( get_post_meta( $post_id, '_wp_trash_meta_status', true ), array( 'wc-pending', 'wc-expired', 'wc-cancelled' ), true ) ) {
 			update_post_meta( $post_id, '_wp_trash_meta_status', 'wc-cancelled' );
 		}
 	}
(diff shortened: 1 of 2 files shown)


includes/class-wc-subscriptions-manager.php
```diff
@@ -841,20 +841,29 @@ public static function maybe_delete_subscription( $post_id ) {
...
 
-		if ( 'shop_subscription' == get_post_type( $post_id ) && 'auto-draft' !== get_post_status( $post_id ) ) {
+		$subscription = wcs_get_subscription( $id );
+		if ( ! $subscription ) {
+			return;
+		}
 
-			$subscription = wcs_get_subscription( $post_id );
+		// Check its type because $subscription might be WC_Subscription's child class.
+		if ( $subscription->get_type() !== 'shop_subscription' ) {
```
mattallan, line 854: Call `wcs_get_subscription()` checks if the subscription is a subscription so we don't need to check the type here now.

includes/class-wc-subscriptions-manager.php
```diff
@@ -841,20 +841,29 @@ public static function maybe_delete_subscription( $post_id ) {
...
+			return;
+		}
 
-			if ( $subscription->can_be_updated_to( 'cancelled' ) ) {
+		if ( 'auto-draft' === $subscription->get_status() ) {
```
mattallan: @james-allan recently changed our `$subscription->set_status()` function (https://github.com/Automattic/woocommerce-subscriptions-core/pull/258) to always make sure 'auto-draft' and 'draft' isn't set as the subscriptions status when it's read from the DB so I don't think we need this check in here anymore.

i.e if I run the following code:

```
$id = 503;

wp_update_post( [ 'ID' =>  $id, 'post_status' => 'draft' ] );
$subscription = wcs_get_subscription( $id );
echo $subscription->get_status() . "\n";

wp_update_post( [ 'ID' =>  $id, 'post_status' => 'auto-draft' ] );
$subscription = wcs_get_subscription( $id );
echo $subscription->get_status();
```

It shows `get_status()` always returns `'pending'` when the status is auto-draft/draft

includes/class-wc-subscriptions-manager.php
```diff
@@ -868,26 +877,36 @@ public static function maybe_cancel_subscription( $post_id ) {
...
-	public static function fix_trash_meta_status( $post_id ) {
+	public static function fix_trash_meta_status( $id ) {
+
+		$subscription = wcs_get_subscription( $id );
+		if ( ! $subscription ) {
+			return;
+		}
+
+		// Check its type because $subscription might be WC_Subscription's child class.
+		if ( $subscription->get_type() !== 'shop_subscription' ) {
```
mattallan, line 885: As above, calling `wcs_get_subscription( $id );` returns `false` if the object isn't a subscription type so we don't need to check this after fetching the subscription object.

mattallan (changes requested):
Thanks @shendy-a8c!

I've tested the following flows that are impacted by these changes:
 - [x] When trashing an active subscription, make sure the subscription is canceled first
 - [x] After a subscription is trashed make sure `_wp_trash_meta_status` is set to `wc-cancelled`

I've tested the following changes on WC 7.2 with HPOS disabled and also WC 6.8.

This code doesn't need to be tested on HPOS-enabled sites as we're still hooking onto `wp_trash_post` and `before_delete_post`. These will be addressed in https://github.com/Automattic/woocommerce-subscriptions-core/issues/221


---

I've made a few comments to clean up some of these functions:
 - [ ] We don't need to call `$subscription->get_type() !== 'shop_subscription'` after fetching the subscription using `wcs_get_subscription()`
 - [ ] We don't need to check `'auto-draft' === $subscription->get_status()` since the subscription status will be returned as `'pending'` for draft statuses.
 - [ ] Changelog is needed

includes/class-wc-subscriptions-manager.php
shendy-a8c: Isn't is possible another plugin maybe extends `WC_Subscription` and the returned value of `wcs_get_subscription()` is the child class of `WC_Subscription`?

Consider this code:
```
$subscription = wcs_get_subscription( 56 );

class WC_Subscription_Child extends WC_Subscription {
    public function __construct( $subscription = 0 ) {
		parent::__construct( $subscription );
		$this->order_type = 'shop_subscription2';
	}
	
	public function get_type() {
		return $this->order_type;
	}
}

wc_register_order_type(
	'shop_subscription2',
	apply_filters(
		'woocommerce_register_post_type_subscription',
		array(
			'class_name' => 'WC_Subscription_Child',
		)
	)
);

$a = wcs_get_subscription( 111111 );
var_dump( $a->get_type() );

$wc_subscription_child = new WC_Subscription_Child( 111111 );
$b = new WC_Subscription_Child( $wc_subscription_child );
var_dump( $b->get_type() );
```

where order with id 56 is a WC_Subscription but 111111 I manually inserted
```
INSERT INTO wp_wc_orders (id, status, currency, type) VALUES (111111, 'wc-active', 'USD', 'shop_subscription2');
```

The PHP code will print `shop_subscription2`. Hence, I thought checking the type and make sure we only handle `shop_subscription` is necessary.

includes/class-wc-subscriptions-manager.php
shendy-a8c: Thank you for pointing that out! Addressed in 057fa07abfbc5c016574e6534529fbac7dd15f87.

I was wondering if it would affect this PR when I read a changelog entry yesterday that says:
```
Replaces the need for the overriding WC_Subscription::get_status() which has been deleted.
```

shendy-a8c:
> Changelog is needed

Addressed in 27ba924bebcf5d057834db7530893758a4f82800.

includes/class-wc-subscriptions-manager.php
mattallan: Good point :) Let's leave the type check in there.

includes/class-wc-subscriptions-manager.php
```diff
@@ -868,26 +873,36 @@ public static function maybe_cancel_subscription( $post_id ) {
...
 	 */
-	public static function fix_trash_meta_status( $post_id ) {
+	public static function fix_trash_meta_status( $id ) {
+
+		$subscription = wcs_get_subscription( $id );
+		if ( ! $subscription ) {
+			return;
+		}
+
+		// Check its type because $subscription might be WC_Subscription's child class.
```
mattallan: > WC_Subscription's child class.

I don't think we need to emphasize that this check is for custom subscription types. I don't think we need this comment but if we want to keep it, we can just say something like: 

```suggestion
		// Check the object type is shop_subscription
```

shendy-a8c: Sounds good! Addressed in [6a1f79c](https://github.com/Automattic/woocommerce-subscriptions-core/pull/305/commits/6a1f79c17117d8599b8bd34f627c4252c9d5df34)

mattallan (approved):
Thanks Shendy! This one is good to merge

PR #317, opened 2022-11-28
[HPOS] Fix customers changing subscription addresses via My Account when HPOS is enabled

### LINKED ISSUE #316: [HPOS] Cannot change an address on a subscription via the My Account > Subscription actions when HPOS is enabled.
## Describe the bug

When HPOS is enabled, going through the change address flows on My Account > Subscriptions and My Account > Addresses does not update any subscription addresses.

This is caused by our code using outdated methods for updating a subscription/orders address.

In our `maybe_update_subscription_addresses()` function we are using the [`$subscription->set_address()`](https://github.com/woocommerce/woocommerce/blob/4cf10ad2174de8bde62fa1b753691e2f091db413/plugins/woocommerce/includes/legacy/abstract-wc-legacy-order.php#L334) method:

```
public function set_address( $address, $type = 'billing' ) {
	foreach ( $address as $key => $value ) {
		update_post_meta( $this->get_id(), "_{$type}_" . $key, $value );
		if ( is_callable( array( $this, "set_{$type}_{$key}" ) ) ) {
			$this->{"set_{$type}_{$key}"}( $value );
		}
	}
}
```

After looking at this method you can see it's doing a few things wrong that won't work in an HPOS environment:

 - it's using `update_post_meta()`
 - It's calling the address setter without saving

Because of this issues, trying to change an address when HPOS is enabled results in the address not being updated and the address fields being saved to postmeta instead of in the new order address table:

image

To fix this we'll need to update our `$subscription->save_address()` to use newer WC Order APIs like `$subscription->set_billing_address()`

## To Reproduce

1. Enable HPOS and purchase a non-virtual subscription (must need shipping to change the address)
2. Make sure to run `DELETE FROM wp_usermeta WHERE meta_key = '_wcs_subscription_ids_cache';` so that the cache is forced to update with the latest users subscriptions.
3. Go into **My Account > Subscriptions**
4. Click on the **Change Address** action on the View Subscription page
5. Change the shipping address
6. Scroll down on the My Account View Subscription page and confirm the subscription has not been updated
7. Look in `wp_wc_order_addresses` and confirm the new address is not present
8. Look in `wp_postmeta` and notice the updated address is sitting in postmeta 😭  

## Product impact

mattallan:
Fixes #316 

## Description

Customers changing subscription addresses via the My Account flow is broken on stores that enable HPOS. This is because our `maybe_update_subscription_addresses()` function was using a [WC Order Legacy `set_address()`](https://github.com/woocommerce/woocommerce/blob/4cf10ad2174de8bde62fa1b753691e2f091db413/plugins/woocommerce/includes/legacy/abstract-wc-legacy-order.php#L334) method which was saving/updating the address data using `update_post_meta().

This issue affects the following flows:
 - Updating a single subscription address via the My Account > Subscription > Change Address
 - Update all subscription addresses via the My Account > Addresses and then check "Update the X Address used for all future renewals of my active subscriptions (optional)"

image

This PR fixes all issues with customers not being able to save/update their subscription addresses.
To fix these issues we just needed to use newer WC Order APIs to set address data.

Changes in this PR:

 - New helper function `wcs_set_order_address( $order, $address, $type )`
   - The most efficient function in WC to set an order address is using `set_billing()` and `set_shipping()` however these weren't introduced until WC 7.1, therefore this helper function acts as a compatibility function in which it will use these methods if they exist, otherwise use the alternative which is to call each individual setter (ie. `set_billing_last_name()`)
 - Update `maybe_update_subscription_addresses()` to replace Order Legacy set address() with new `wcs_set_order_address()`

Another issue I discovered was when HPOS was enabled with data syncing off, we were still setting the address fields in postmeta:

image

This issue has also been fixed in this PR.

## How to test this PR

**Changing A Single Subscription Address**

1. Enable HPOS and purchase a non-virtual subscription (must need shipping to change the address)
2. Make sure to run `DELETE FROM wp_usermeta WHERE meta_key = '_wcs_subscription_ids_cache';` so that the cache is forced to update with the latest users subscriptions.
3. Go into **My Account > Subscriptions**
4. Click on the **Change Address** action on the View Subscription page
5. Change the shipping address
6. Scroll down on the My Account View Subscription page and confirm the subscription has the updated shipping address

**Updating the address of all the users active susbcriptions**

Similarly to the above steps:

1. Enable HPOS and purchase a non-virtual subscription (must need shipping to change the address)
2. Make sure to run `DELETE FROM wp_usermeta WHERE meta_key = '_wcs_subscription_ids_cache';` so that the cache is forced to update with the latest users subscriptions.
3. Go into **My Account > Addresses**
4. Click on the icons next to Shipping/Billing address to edit them
5. After updating the address, scroll to the bottom and click on "Update the X Address used for all future renewals of my active subscriptions (optional)"
6. Save the address and check your subscriptions have the latest chanegs.

## Product impact

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -2,6 +2,7 @@
 
 = 5.2.0 - 2022-xx-xx =
 * Fix - Refactor `WCS_Meta_Box_Subscription_Data::save` to support HPOS stores, fixing a PHP warning notice when updating an order via the Edit Order screen.
+* Fix - Set the `download_permissions_granted` value when purchasing a downloadable subscription product when HPOS is enabled.
 * Update - Replace `get_post_type()` calls in `WC_Subscriptions_Manager` for HPOS support.
 * Update - Replace `get_post_type()` calls in `WC_Subscriptions_Synchroniser` for HPOS support.
 
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,8 +1,10 @@
 *** WooCommerce Subscriptions Core Changelog ***
 
 = 5.2.0 - 2022-xx-xx =
+* Add - New wcs_set_order_address() helper function to set an array of address fields on an order or subscription.
 * Fix - Refactor `WCS_Meta_Box_Subscription_Data::save` to support HPOS stores, fixing a PHP warning notice when updating an order via the Edit Order screen.
 * Fix - Set the `download_permissions_granted` value when purchasing a downloadable subscription product when HPOS is enabled.
+* Fix - When a customer changes their address on their account or subscription, make sure the new address is saved when HPOS is enabled.
 * Update - Replace `get_post_type()` calls in `WC_Subscriptions_Manager` for HPOS support.
 * Update - Replace `get_post_type()` calls in `WC_Subscriptions_Synchroniser` for HPOS support.
 
--- a/includes/class-wc-subscriptions-addresses.php
+++ b/includes/class-wc-subscriptions-addresses.php
@@ -159,7 +159,6 @@ class WC_Subscriptions_Addresses {
 		}
 
 		if ( isset( $_POST['update_all_subscriptions_addresses'] ) ) {
-
 			$users_subscriptions = wcs_get_users_subscriptions( $user_id );
 
 			foreach ( $users_subscriptions as $subscription ) {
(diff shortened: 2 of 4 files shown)


brucealdridge (approved):
LGTM.
Have tested both situations and confirmed working as expected.
Have included a small docblock fix I found when stepping through the code.

mattallan:
Thanks very much @brucealdridge for the test and the docblock fixes!
I've just pushed up a changelog entry for this PR. Merging!

PR #318, opened 2022-11-28
Replace all remaining `get_post_type()`

shendy-a8c:
https://github.com/Automattic/woocommerce-subscriptions-core/issues/288

## Description

- Change changelog entry per [@mattallan's suggestion](https://github.com/Automattic/woocommerce-subscriptions-core/pull/306#issuecomment-1327077739).
- Replace `get_post_type( $id )` with `WC_Data_Store::load( 'subscription' )->get_order_type( $id )`;

## How to test this PR

For `WC_Subscriptions_Product::user_can_not_delete_subscription` (`includes/class-wc-subscriptions-product.php`):
- Create a subscription product (let's call it `Subs1`).
- Purchase `Subs1`.
- Trash `Subs1`.
- Create another subscription product, `Subs2`.
- Trash `Subs2`.
- Go to the trash list (/wp-admin/edit.php?post_status=trash&post_type=product) and make sure you cannot delete permanently `Subs1` but can do so for `Subs2`.

Note: with HPOS on, `Delete Permanently` show up even for subscription products that are associated to an order. I created a ticket [here](https://github.com/Automattic/woocommerce-subscriptions-core/issues/319).

<br />

For `WCS_Admin_Post_Types::print_bulk_actions_script` (`includes/admin/class-wcs-admin-post-types.php`):
- Open the admin edit page of an active subscription.
- Look at the HTML source of that page, make sure you can find `.appendTo( "select[name='action'], select[name='action2']" );`.

Note: as of currently, we might not be able to test this with HPOS on.

<br />

For `WCS_Cached_Data_Manager::purge_delete` (`includes/class-wcs-cached-data-manager.php`):
No need to test as it's deprecated. Make sure code diff looks good.

<br />

For `WCS_Admin_Post_Types::print_bulk_actions_script` (`includes/admin/class-wcs-admin-post-types.php`):
- Make sure bulk actions show up on subscriptions admin list (`/wp-admin/edit.php?post_type=shop_subscription`).
Screenshot 2022-12-01 at 11 59 48

## Product impact

### DIFF
--- a/includes/admin/class-wcs-admin-post-types.php
+++ b/includes/admin/class-wcs-admin-post-types.php
@@ -249,7 +249,7 @@ class WCS_Admin_Post_Types {
 
 		$post_status = ( isset( $_GET['post_status'] ) ) ? $_GET['post_status'] : '';
 
-		$subscription_id = $GLOBALS['post'] ?? '';
+		$subscription_id = ( ! empty( $GLOBALS['post']->ID ) ) ? $GLOBALS['post']->ID : '';
 		if ( ! $subscription_id ) {
 			return;
 		}
--- a/includes/admin/class-wcs-admin-post-types.php
+++ b/includes/admin/class-wcs-admin-post-types.php
@@ -249,7 +249,12 @@ class WCS_Admin_Post_Types {
 
 		$post_status = ( isset( $_GET['post_status'] ) ) ? $_GET['post_status'] : '';
 
-		if ( 'shop_subscription' !== WC_Data_Store::load( 'subscription' )->get_order_type() || in_array( $post_status, array( 'cancelled', 'trash', 'wc-expired' ), true ) ) {
+		$subscription_id = $GLOBALS['post'] ?? '';
+		if ( ! $subscription_id ) {
+			return;
+		}
+
+		if ( 'shop_subscription' !== WC_Data_Store::load( 'subscription' )->get_order_type( $subscription_id ) || in_array( $post_status, array( 'cancelled', 'trash', 'wc-expired' ), true ) ) {
 			return;
 		}
 
--- a/includes/admin/class-wcs-admin-post-types.php
+++ b/includes/admin/class-wcs-admin-post-types.php
@@ -249,7 +249,7 @@ class WCS_Admin_Post_Types {
 
 		$post_status = ( isset( $_GET['post_status'] ) ) ? $_GET['post_status'] : '';
 
-		if ( 'shop_subscription' !== get_post_type() || in_array( $post_status, array( 'cancelled', 'trash', 'wc-expired' ) ) ) {
+		if ( 'shop_subscription' !== WC_Data_Store::load( 'subscription' )->get_order_type() || in_array( $post_status, array( 'cancelled', 'trash', 'wc-expired' ), true ) ) {
 			return;
 		}
 
--- a/includes/class-wc-subscriptions-product.php
+++ b/includes/class-wc-subscriptions-product.php
@@ -724,7 +724,7 @@ class WC_Subscriptions_Product {
 	public static function user_can_not_delete_subscription( $allcaps, $caps, $args ) {
 		global $wpdb;
 
-		if ( isset( $args[0] ) && in_array( $args[0], array( 'delete_post', 'delete_product' ) ) && isset( $args[2] ) && ( ! isset( $_GET['action'] ) || 'untrash' != $_GET['action'] ) && 0 === strpos( WC_Data_Store::load( 'subscription' )->get_order_type( $args[2] ), 'product' ) ) { // phpcs:ignore
+		if ( isset( $args[0] ) && in_array( $args[0], array( 'delete_post', 'delete_product' ), true ) && isset( $args[2] ) && ( ! isset( $_GET['action'] ) || 'untrash' !== $_GET['action'] ) && 0 === strpos( WC_Data_Store::load( 'subscription' )->get_order_type( $args[2] ), 'product' ) ) { // phpcs:ignore
 
 			$user_id = $args[2];
 			$post_id = $args[2];
--- a/includes/class-wc-subscriptions-product.php
+++ b/includes/class-wc-subscriptions-product.php
@@ -724,7 +724,7 @@ class WC_Subscriptions_Product {
 	public static function user_can_not_delete_subscription( $allcaps, $caps, $args ) {
 		global $wpdb;
 
-		if ( isset( $args[0] ) && in_array( $args[0], array( 'delete_post', 'delete_product' ), true ) && isset( $args[2] ) && ( ! isset( $_GET['action'] ) || 'untrash' !== $_GET['action'] ) && 0 === strpos( WC_Data_Store::load( 'subscription' )->get_order_type( $args[2] ), 'product' ) ) { // phpcs:ignore
+		if ( isset( $args[0] ) && in_array( $args[0], array( 'delete_post', 'delete_product' ) ) && isset( $args[2] ) && ( ! isset( $_GET['action'] ) || 'untrash' != $_GET['action'] ) && 0 === strpos( get_post_type( $args[2] ), 'product' ) ) {
 
 			$user_id = $args[2];
 			$post_id = $args[2];
--- a/includes/class-wc-subscriptions-product.php
+++ b/includes/class-wc-subscriptions-product.php
@@ -724,7 +724,7 @@ class WC_Subscriptions_Product {
 	public static function user_can_not_delete_subscription( $allcaps, $caps, $args ) {
 		global $wpdb;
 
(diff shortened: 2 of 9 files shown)


includes/class-wc-subscriptions-product.php
```diff
@@ -724,7 +724,7 @@ public static function maybe_set_variations_price_html( $variation_details, $var
 	public static function user_can_not_delete_subscription( $allcaps, $caps, $args ) {
 		global $wpdb;
 
-		if ( isset( $args[0] ) && in_array( $args[0], array( 'delete_post', 'delete_product' ) ) && isset( $args[2] ) && ( ! isset( $_GET['action'] ) || 'untrash' != $_GET['action'] ) && 0 === strpos( get_post_type( $args[2] ), 'product' ) ) {
+		if ( isset( $args[0] ) && in_array( $args[0], array( 'delete_post', 'delete_product' ), true ) && isset( $args[2] ) && ( ! isset( $_GET['action'] ) || 'untrash' !== $_GET['action'] ) && 0 === strpos( WC_Data_Store::load( 'subscription' )->get_order_type( $args[2] ), 'product' ) ) { // phpcs:ignore
```
shendy-a8c: Following similar [suggestion](https://github.com/Automattic/woocommerce-subscriptions-core/pull/306#discussion_r1031932685), I tried `WC_Data_Store::load( 'product' )->get_order_type()` but it didn't work. It returns `null`. I'm not sure why.

includes/admin/class-wcs-admin-post-types.php
```diff
@@ -249,7 +249,7 @@ public function print_bulk_actions_script() {
 
 		$post_status = ( isset( $_GET['post_status'] ) ) ? $_GET['post_status'] : '';
 
-		if ( 'shop_subscription' !== get_post_type() || in_array( $post_status, array( 'cancelled', 'trash', 'wc-expired' ) ) ) {
+		if ( 'shop_subscription' !== WC_Data_Store::load( 'subscription' )->get_order_type() || in_array( $post_status, array( 'cancelled', 'trash', 'wc-expired' ), true ) ) {
```
shendy-a8c: `til:` if no id is passed, it will use `$GLOBALS['post']` ([ref](https://github.com/WordPress/WordPress/blob/25ea9680aecaf4791c870658fd8085c10622ad05/wp-includes/post.php#L991-L993)).

includes/class-wcs-cached-data-manager.php
```diff
@@ -73,16 +73,14 @@ public function cache_and_get( $key, $callback, $params = array(), $expires = WE
 	public function purge_delete( $post_id, $post = null ) {
 		wcs_deprecated_function( __METHOD__, '2.3.0' );
 
-		$post_type = get_post_type( $post_id );
-
-		if ( 'shop_order' === $post_type ) {
+		if ( 'shop_order' === WC_Data_Store::load( 'order' )->get_order_type( $post_id ) ) {
```
shendy-a8c, line 76: Won't take an extra mile to rename the `$post_id` variable to `$id` here as this function is deprecated. This change is so that we cannot find `get_post_type()` in our code base. Let me know if I should go ahead and change the variable name.

shendy-a8c:
@mattallan could you give some pointers on how to deal with `includes/class-wcs-post-meta-cache-manager.php`? That class seems to be very specific for post meta.

includes/admin/class-wcs-admin-post-types.php
mattallan: Thanks for pointing this out @shendy-a8c. This complicates things a little because if this function is called on stores that have HPOS enabled, the global post won't exist and the [OrdersTable data store class](https://github.com/woocommerce/woocommerce/blob/a734772c69f953fa466e861670f39732f0c83c7f/plugins/woocommerce/src/Internal/DataStores/Orders/OrdersTableDataStore.php#L949) must be sent with an ID otherwise an error will be thrown.

Even though the postID is defaulted in `get_post_type()` when the CPT data store is loaded, we'll need to make sure this function doesn't break when the HPOS data store is loaded.

That means we can either:

1. Return early if store has HPOS enabled.
2. Look for and fetch an ID from `$GLOBALS['post']` if it exists. If no post/ID is found, then we return early from the function.

Personally I like option 2 because it's a bit more defensive, but I'm happy with either option :)

includes/class-wc-subscriptions-product.php
mattallan: For safety, we can revert this change and leave it as is since products are staying in the Posts table.

--


> WC_Data_Store::load( 'product' )->get_order_type()

To answer this question, the product data store doesn't have a `get_order_type()` function.
We need to use `get_product_type()` but this current `strpos` won't work becuase `WC_Data_Store::load( 'product' )->get_product_type()` changes the product type. For instance, for posts with type `product_variation` this function returns 'variation' so we can't look for product types that include 'product'

As I mentioned before, we can revert this change and leave it as is since it will require some bigger logic changes.

includes/class-wcs-post-meta-cache-manager.php
```diff
@@ -248,12 +248,12 @@ protected function trigger_delete_all_caches_hook( $meta_key ) {
...
 	 *
-	 * @param int $post_id Post ID or post object.
-	 * @return bool Whether the post type for the given post ID is the post type this instance manages.
+	 * @param int $id Subscription ID or subscription object itself.
+	 * @return bool Whether the order type for the given ID is the order type this instance manages.
 	 */
-	protected function is_managed_post_type( $post_id ) {
-		return $this->post_type === get_post_type( $post_id );
+	protected function is_managed_post_type( $id ) {
+		return WC_Data_Store::load( 'subscription' )->get_order_type( $id ) === $this->post_type;
```
mattallan: After thinking about this change a little more, I don't think we need to update this function because the `Post_Meta_Cache_Manager` class is predicated on the assumption that this class will be used for any posts.

Therefore loading the subscriptions data store and checking order type doesn't make sense here and might add confusion (even though the CPT `get_order_type()` is just a wrapper for `get_post_type()`.

For stores that have HPOS is enabled, we'll now use James' new `WCS_Object_Data_Cache_Manager` class being added in #304.

So I say let's leave this function using `$this->post_type === get_post_type( $post_id );` and I'll make a comment on James' PR to add a note to this class to specify it's only used for WP Posts and also make sure we're overriding `is_managed_post_type()` in the new `WCS_Object_Data_Cache_Manager` class (which extends this)

Let me know if that makes sense.

mattallan (commented):
Thanks @shendy-a8c for this PR! I've added a few comments as part of code review, let me know if you have any questions or need clarification!

includes/class-wc-subscriptions-product.php
shendy-a8c: Thank you for explaining all that. It makes so much sense. Reverted in [16b9bc5](https://github.com/Automattic/woocommerce-subscriptions-core/pull/318/commits/16b9bc566fff34aa6cc2a90240d22b506f91f1f4).

includes/class-wcs-post-meta-cache-manager.php
shendy-a8c: It makes sense. Reverted in [2b1d9d5](https://github.com/Automattic/woocommerce-subscriptions-core/pull/318/commits/2b1d9d5cc44188df80af84a62d25c762fa681ab7).

includes/admin/class-wcs-admin-post-types.php
shendy-a8c: That makes sense. Addressed option 2 in [92c767f](https://github.com/Automattic/woocommerce-subscriptions-core/pull/318/commits/92c767fbcc7ac2ade5efd2f4f19404a36085e5ab).

shendy-a8c:
@mattallan this PR is ready for another round of review.

includes/admin/class-wcs-admin-post-types.php
```diff
@@ -249,7 +249,12 @@ public function print_bulk_actions_script() {
 
 		$post_status = ( isset( $_GET['post_status'] ) ) ? $_GET['post_status'] : '';
 
-		if ( 'shop_subscription' !== get_post_type() || in_array( $post_status, array( 'cancelled', 'trash', 'wc-expired' ) ) ) {
+		$subscription_id = $GLOBALS['post'] ?? '';
```
mattallan: I haven't tested this but `$GLOBALS['post']` is the post object isn't it? whereas we want to send the ID to the data store function

Does that mean we want this line need to be something like:

```
$subscription_id = ! empty( $GLOBALS['post']->ID ) ? $GLOBALS['post']->ID : '';
```

shendy-a8c: Ah you're right. When I debugged, it did show up as a post object and it worked because `WC_Data_Store::load( 'subscription' )->get_order_type()` can take object just fine too but I named the variable `$subscription_id` so I should have got the post's id.

Addressed in [e9f34d1](https://github.com/Automattic/woocommerce-subscriptions-core/pull/318/commits/e9f34d1e6d4e20971ec5b36f11b192453f2d532f).

mattallan (approved):
Thanks @shendy-a8c ! These changes look good to me

Issue #325, opened 2022-11-29
Add support for php8.2


brucealdridge:
## Description

There are a number of features that are being deprecated in php 8.2 and then removed in php9 that we need to resolve.
Full detail of [php8.2 changes](https://stitcher.io/blog/new-in-php-82)

#### Deprecation of dynamic properties [RFC](https://wiki.php.net/rfc/deprecate_dynamic_properties)

Anywhere that class variables are used without being explicitly set (`__get` and `__set` will continue to work) will now generate a Deprecation warning.

https://github.com/Automattic/woocommerce-subscriptions-core/blob/d08ed6617ad73fb8dc3dd16bce23a4817ae5c78c/includes/legacy/class-wc-product-subscription-legacy.php#L39-L41

In this example, `setting_id` is not defined and will generate a warning.

#### Deprecation of `${}` string interpolation [RFC](https://wiki.php.net/rfc/deprecate_dollar_brace_string_interpolation)

Usage of `${}` is now deprecated inside strings will now generate a Deprecation warning.

I can only find one instance of this which should be okay as this is **not** inside a string
https://github.com/Automattic/woocommerce-subscriptions-core/blob/d08ed6617ad73fb8dc3dd16bce23a4817ae5c78c/includes/class-wcs-cart-renewal.php#L1171

#### Deprecation of partially supported callables [RFC](https://wiki.php.net/rfc/deprecate_partially_supported_callables)

The following callable types are now deprecated
```
"self::method"
"parent::method"
"static::method"
["self", "method"]
["parent", "method"]
["static", "method"]
["Foo", "Bar::method"]
[new Foo, "Bar::method"]
```

They can be replaced using the following pattern.
```
"self::method"       -> self::class . "::method"
"parent::method"     -> parent::class . "::method"
"static::method"     -> static::class . "::method"
["self", "method"]   -> [self::class, "method"]
["parent", "method"] -> [parent::class, "method"]
["static", "method"] -> [static::class, "method"]
```

I couldn't find any examples of usage in the codebase where updates will be required but further testing will be needed.

### Testing instructions

Run smoke tests and unit tests with php8.2 installed.

related to #324

bummytime:
Bumping this → I saw a few of these warnings in wp-admin after updating WooPayments:

```
[...]

Issue #156, opened 2022-04-25
CI: Database tables don't exist when running unit tests


brucealdridge:
## Describe the bug

When running the unit tests in github actions, the logs show that not all tables are setup correctly.
All tests show as completing successfully and is just a logged error.

**Log snippet**
```
[...]

PR #326, opened 2022-11-29
Add support for php8.2 in github action checks

### LINKED ISSUE #324: Add php 8.2 to Github actions checks
## Description

Currently, our tests are run against the following versions of php.
https://github.com/Automattic/woocommerce-subscriptions-core/blob/d08ed6617ad73fb8dc3dd16bce23a4817ae5c78c/.github/workflows/php-lint-test.yml#L54

With the upcoming release of php 8.2 on December 8 these checks should be updated to include the latest version. 

## Product impact

This is a dev only change and will not affect WooCommerce Payments or WooCommerce Subscriptions.

brucealdridge:
Fixes #324 

## Description

Adds support for php8.2 to the github action checks.
As php8.2 is not yet released, this is using a nightly build.

## How to test this PR

Review the output of github actions and verify that the unit tests are being run against php 8.2

## Product impact

No product impact, development only change.

### DIFF
--- a/.github/workflows/php-lint-test.yml
+++ b/.github/workflows/php-lint-test.yml
@@ -51,7 +51,7 @@ jobs:
       fail-fast:    false
       max-parallel: 10
       matrix:
-        php: [ '7.1', '7.2', '7.3', '7.4', '8.0', '8.1' ]
+        php: [ '7.1', '7.2', '7.3', '7.4', '8.0', '8.1', '8.2' ]
     steps:
       # clone the repository
       - uses: actions/checkout@v2


brucealdridge:
**Note:** The checks are currently failing while still being marked as passing. This is to be resolved in #156 and #325

shendy-a8c (approved):
Approving just by looking at the diff.

shendy-a8c:
> Note: The checks are currently failing while still being marked as passing. This is to be resolved in https://github.com/Automattic/woocommerce-subscriptions-core/issues/156 and https://github.com/Automattic/woocommerce-subscriptions-core/issues/325

I don't quite understand what you mean especially the "The checks are currently failing" part because I see all GitHub checks are green.

brucealdridge:
@shendy-a8c the checks are green in github, but the logs are actually showing some errors. Most of these appear to be due to an incorrectly setup database which IMO should cause the checks to fail #156

PR #331, opened 2022-12-05
Refactor `WC_Subscriptions_Renewal_Order::get_failed_order_replaced_by()` to support HPOS

### LINKED ISSUE #143: Refactor WC_Subscriptions_Renewal_Order::get_failed_order_replaced_by() to not query for orders using direct DB queries on the posts table
## Description

With Custom Order Tables being worked on by the WC Core team ([public announcement](https://developer.woocommerce.com/2022/01/17/the-plan-for-the-woocommerce-custom-order-table/)), we need to begin to think about updating the Subscriptions Core library to remove any code that makes direct DB queries to the post table when fetching Order related data.

In this issue, we're tackling the only uses of `$wpdb->prepare()` where an order is being queried. Subscriptions, Products, and Order Items are fine.

## Function that needs updating

- [ ] [`get_failed_order_replaced_by()`](https://github.com/Automattic/woocommerce-subscriptions-core/blob/2a0654bf0385cc3e5b097943aea5589e843b05d0/includes/class-wc-subscriptions-renewal-order.php#L60)

## How to update code?

Direct DB queries should be replaced with `wc_get_orders()`/`WC_Order_Query`, however I'm not exactly sure if every use case is as simple as that. If there's some complicated code/queries we may have to look at refactoring some code as part of tackling this issue.

## Impact

 - PayPal Standard payment request handling

Within out codebase, this function is only used in PayPal Standard payment request handling so I was say **Impact is low** however, because this is a public static function it's possible some 3rd party code might be using it.

## Product impact

- Does this feature affect WooCommerce Payments? no (potentially does impact WCPay if 3rd party subscriptions code is using this function)

## Additional context

Project thread: pdjTHR-Xp-p2

Jinksi:
Fixes #143

## Description

This PR replaces a call to `$wpdb->prepare` with `wcs_get_orders_with_meta_query` to ensure `WC_Subscriptions_Renewal_Order::get_failed_order_replaced_by()` is compatible with HPOS stores.

PayPal Standard payment request handling is the only use-case of this function internally. However, this is a public static function, so 3rd party code may be using it.

As mentioned in the [issue thread](https://github.com/Automattic/woocommerce-subscriptions-core/issues/143#issuecomment-1333164770), this change is relatively low-risk, and testing by directly calling the function should suffice.

## How to test this PR

1. Create a renewal order for a subscription. Note the renewal order ID (aka `$sub_id` below).
2. Manually add `_failed_order_replaced_by` to a previous order ID, via the `_postmeta` table (HPOS disabled) or `_wc_orders_meta` table.
3. Run `WC_Subscriptions_Renewal_Order::get_failed_order_replaced_by( $sub_id );` in [WP Console](https://wordpress.org/plugins/wp-console/).
5. Confirm the ID of the manually mutated order is returned.
6. Run `WC_Subscriptions_Renewal_Order::get_failed_order_replaced_by( $another_order_id );` in [WP Console](https://wordpress.org/plugins/wp-console/). `$another_order_id` should be an order that does not have the `_failed_order_replaced_by` meta.
7. Confirm it's returning `false`.

## Product impact

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -7,6 +7,7 @@
 * Fix - When a customer changes their address on their account or subscription, make sure the new address is saved when HPOS is enabled.
 * Update - Replace `get_post_type()` calls in `WC_Subscriptions_Manager` for HPOS support.
 * Update - Replace `get_post_type()` calls in `WC_Subscriptions_Synchroniser` for HPOS support.
+8 Fix - Refactor `WC_Subscriptions_Renewal_Order::get_failed_order_replaced_by()` to support HPOS stores.
 
 = 5.1.0 - 2022-11-24 =
 * Fix - Set payment tokens when copying data between orders and subscriptions in a CRUD compatible way. Fixes PHP notices during renewal order process.
--- a/changelog.txt
+++ b/changelog.txt
@@ -7,7 +7,7 @@
 * Fix - When a customer changes their address on their account or subscription, make sure the new address is saved when HPOS is enabled.
 * Update - Replace `get_post_type()` calls in `WC_Subscriptions_Manager` for HPOS support.
 * Update - Replace `get_post_type()` calls in `WC_Subscriptions_Synchroniser` for HPOS support.
-8 Fix - Refactor `WC_Subscriptions_Renewal_Order::get_failed_order_replaced_by()` to support HPOS stores.
+* Fix - Refactor `WC_Subscriptions_Renewal_Order::get_failed_order_replaced_by()` to support HPOS stores.
 
 = 5.1.0 - 2022-11-24 =
 * Fix - Set payment tokens when copying data between orders and subscriptions in a CRUD compatible way. Fixes PHP notices during renewal order process.
--- a/changelog.txt
+++ b/changelog.txt
@@ -6,8 +6,6 @@
 * Fix - Set the `download_permissions_granted` value when purchasing a downloadable subscription product when HPOS is enabled.
 * Fix - When a customer changes their address on their account or subscription, make sure the new address is saved when HPOS is enabled.
 * Fix - Removed the potential for an infinite loop when getting a subscription's related orders while the subscription is being loaded.
-* Update - Replace `get_post_type()` calls in `WC_Subscriptions_Manager` for HPOS support.
-* Update - Replace `get_post_type()` calls in `WC_Subscriptions_Synchroniser` for HPOS support.
 * Fix - Refactor `WC_Subscriptions_Renewal_Order::get_failed_order_replaced_by()` to support HPOS stores.
 * Dev - Replace code using get_post_type( $subscription_id ) with WC Data Store get_order_type().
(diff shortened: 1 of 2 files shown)


changelog.txt
```diff
@@ -7,6 +7,7 @@
 * Fix - When a customer changes their address on their account or subscription, make sure the new address is saved when HPOS is enabled.
 * Update - Replace `get_post_type()` calls in `WC_Subscriptions_Manager` for HPOS support.
 * Update - Replace `get_post_type()` calls in `WC_Subscriptions_Synchroniser` for HPOS support.
+8 Fix - Refactor `WC_Subscriptions_Renewal_Order::get_failed_order_replaced_by()` to support HPOS stores.
```
```suggestion
* Fix - Refactor `WC_Subscriptions_Renewal_Order::get_failed_order_replaced_by()` to support HPOS stores.
```

includes/class-wc-subscriptions-renewal-order.php
```diff
@@ -55,11 +55,23 @@ public static function trigger_renewal_payment_complete( $order_id ) {
 	 * @return mixed If the renewal order did replace a failed order, the ID of the fail order, else false
 	 */
 	public static function get_failed_order_replaced_by( $renewal_order_id ) {
-		global $wpdb;
 
-		$failed_order_id = $wpdb->get_var( $wpdb->prepare( "SELECT post_id FROM $wpdb->postmeta WHERE meta_key = '_failed_order_replaced_by' AND meta_value = %s", $renewal_order_id ) );
+		// Get orders where order meta '_failed_order_replaced_by' = $renewal_order_id
+		$failed_orders = wcs_get_orders_with_meta_query(
+			array(
```
mattallan: Small nit-pick but we're using the old `array()` syntax here.

mattallan (approved):
It's so satisfying being able to change these DB queries with the new `wcs_get_orders_with_meta_query()` function.

I've found a couple of small issues in code review. Apart from that this is good to go :)

includes/class-wc-subscriptions-renewal-order.php
Jinksi: Thanks! I've updated to short `[]` syntax, which is consistent with the rest of this codebase.

This confuses me a little since the [WP coding standards](https://developer.wordpress.org/coding-standards/wordpress-coding-standards/php/#declaring-arrays) say: "Arrays must be declared using long array syntax."

I'm assuming that WooCommerce standards allow it?

mattallan: > I'm assuming that WooCommerce standards allow it?

We exclude the rule in our `phpcs.xml.dist` file [here](https://github.com/Automattic/woocommerce-subscriptions-core/blob/trunk/phpcs.xml.dist#L26) which I believe we copied from WC Payments 😄

mattallan (approved):
Thanks for making those changes. This is good to merge! :shipit:

PR #341, opened 2022-12-06
Use try/catch when calling process_payment to avoid fatals.

### LINKED ISSUE #330: change_payment_method_via_pay_shortcode doesn't catch Exceptions when calling process_payment()
## Describe the bug
I'm not 100% on the exact steps to replicate as I've only come across this via logs but here's the error:

```
Uncaught Exception: We're not able to process this request. Please refresh the page and try again. 
in /public_html/wp-content/plugins/woocommerce-payments/woocommerce-payments-3.6.1/includes/class-wc-payment-gateway-wcpay.php:893
Stack trace:
#0 public_html/wp-content/plugins/woocommerce-subscriptions/woocommerce-subscriptions-3.1.4/includes/class-wc-subscriptions-change-payment-gateway.php(355): WC_Payment_Gateway_WCPay->process_payment(395)
#1 public_html/wp-includes/class-wp-hook.php(308): WC_Subscriptions_Change_Payment_Gateway::change_payment_method_via_pay_shortcode('')
#2 public_html/wp-includes/class-wp-hook.php(332): WP_Hook->apply_filters(NULL, Array)
#3 public_html/wp-includes/plugin.php(517): WP_Hook->do_action(Array)
#4 public_html/wp-settings.php(655): do_action('wp_loaded')
#5 public_html/wp-config.php(7): require('/home/wpcom/pub...')
#6 public_html/wp-load.php(50): require_once('/home/wpcom/pub...')
#7 public_html/wp-blog-header.php(13): require_once('/home/wpcom/pub...')
#8 public_html/index.php(17): require('/home/wpcom/pub...')
#9 {main}
  thrown in public_html/wp-content/plugins/woocommerce-payments/woocommerce-payments-3.6.1/includes/class-wc-payment-gateway-wcpay.php on line 893
```

## To Reproduce
I believe this could be reproduced by changing the associated payment method of a subscription and using a card that causes an error or perhaps blocking the network request to to update the card.

In any case, the problem appears to be that some gateways, notably WCPay, will throw an Exception in some cases [within the process_payment()](https://github.com/Automattic/woocommerce-payments/blob/c6154cddf73f49e486ee9316da1075c9cf982418/includes/class-wc-payment-gateway-wcpay.php#L748) method [which Subscriptions is calling](https://github.com/Automattic/woocommerce-subscriptions-core/blob/290be54fdf390108a2a27f4a90e96358a60ecf28/includes/class-wc-subscriptions-change-payment-gateway.php#L334), however, Subscriptions is not calling `process_payment()` within a `try {} catch{}` block. It looks like WCPay in this case, throws another Exception within a catch{} block. _I heard you like Exceptions_. 😏 

### Expected behavior
I would expect Subs to handle this Exception without causing a fatal.

## Product impact

tommyshellberg:
Fixes #330, please see for details.

Please also see discussion in p1670014386600359-slack-C029ME0FYRZ for how this affects processing subscription payment method changes with WooCommerce Payments in particular.

## Description

See #330 for a description.

## How to test this PR

Use the Change Payment Method instructions:

https://woocommerce.com/document/subscriptions/customers-view/#section-11

To simulate a failure, you could either block the network request w/ Developer Tools or use a test card that will decline.
Instead of a fatal error because of a failure, you should instead see a notice with the error.

## Product impact


includes/class-wc-subscriptions-change-payment-gateway.php
```diff
@@ -331,7 +331,12 @@ public static function change_payment_method_via_pay_shortcode() {
 			// Process payment for the new method (with a $0 order total)
 			if ( wc_notice_count( 'error' ) == 0 ) {
 
-				$result = $available_gateways[ $new_payment_method ]->process_payment( $subscription->get_id() );
+				try {
+					$result = $available_gateways[ $new_payment_method ]->process_payment( $subscription->get_id() );
+				} catch ( Exception $e ) {
+					wc_add_notice( $e->getMessage(), 'error' );
+					return;
```
Jinksi, line 338: I have a backwards-compatibility concern about returning before the `woocommerce_subscriptions_process_payment_for_change_method_via_pay_shortcode` filter.

https://github.com/Automattic/woocommerce-subscriptions-core/blob/1e980e5eac108bb551712bdf9952cb98ce079e5b/includes/class-wc-subscriptions-change-payment-gateway.php#L334-L344

If this filter is being used to override `$result['result']` to a `'success'` and we `return` before this is executed, we may introduce a bug.

Can I ask @Automattic/helix for a second opinion here?

shendy-a8c: I might not understand your concern correctly, @Jinksi, but I see that this PR handles the case where exception is thrown by `process_payment()`, which in that case, the code execution won't get to `woocommerce_subscriptions_process_payment_for_change_method_via_pay_shortcode` either, so I think there is no backward-compatibility issue here.

shendy-a8c:
I have tested by throwing an Exception from [WCPay's process_payment()](https://github.com/Automattic/woocommerce-payments/blob/cdac6f19ddfd60dd0462486a5c98695522a4df4c/includes/class-wc-payment-gateway-wcpay.php#L671) and confirm I got a fatal error with base branch and with this PR's head branch, I got the exception message displayed and not a fatal error.

I have a minor concern that the exception message got exposed and almost suggested to pass a generic error message to `wc_add_notice()` and log the thrown exception message but then I though with current situation, the exception message is already exposed anyway along with a fatal error, so I refrain myself from suggesting anything more.

shendy-a8c (approved):

PR #342, opened 2022-12-06
Add payment token when setting default payment method for all subscriptions

### LINKED ISSUE #128: Active Subscriptions not updated when changing default payment method
## Describe the bug

Updating the default payment method and choosing `Yes` to update active subscriptions to the new payment method does not update the subscription payment method as expected.

## To Reproduce

1. Create a new subscription order.
2. As a customer, add a new payment method.
3. Make the new payment method the default.
4. When prompted, select `Yes` to update subscriptions to the new payment method.
5. See payment method for the subscription is not updated.

## Product impact

- [x] Does this issue affect WooCommerce Payments? Yes https://github.com/Automattic/woocommerce-payments/issues/3729 
- [x] Does this issue affect WooCommerce Subscriptions? No

## Screenshots
Clicking `Make Default` on the Amex card ending in 0005. Presents the prompt.
Screen Shot 2022-03-23 at 12 59 15 PM
Clicking `Yes` does not update the payment method for the subscription
Screen Shot 2022-03-23 at 12 59 35 PM

shendy-a8c:
Fixes https://github.com/Automattic/woocommerce-subscriptions-core/issues/128

## How to test this PR

1. Go to …
2. Click on …

## Product impact

### DIFF
--- a/includes/class-wcs-payment-tokens.php
+++ b/includes/class-wcs-payment-tokens.php
@@ -45,6 +45,7 @@ class WCS_Payment_Tokens extends WC_Payment_Tokens {
 		}
 
 		$subscription->update_meta_data( $token_meta_key, $new_token->get_token() );
+		$subscription->add_payment_token( $new_token );
 		$subscription->save();
 
 		// Copy the new token to the last renewal order if it needs payment so the retry system will pick up the new method.


mattallan:
Hey @shendy-a8c thanks for opening this PR. At first, I thought this solution was pretty harmless but after thinking on this more, I realized that calling `add_payment_token()` adds the token ID to a list of payment tokens on the subscription which may not be the behaviour that every payment gateway wants.

Further down in this function we have the following docblock comment above the `woocommerce_subscription_token_changed` action hook.

```
Enable third-party plugins to run their own updates and filter whether the token was updated or not.
```

For that reason, I think we should fix this issue in WC Payments as this issue only affects WC Payments.

There are also other benefits to fixing this in WC Payments:
 - the issue will be fixed for stores using older versions of Subscriptions
 - this won't have the potential to break other payment gateways

I'm going to close this PR in favour of addressing the fixes in WC Payments: https://github.com/Automattic/woocommerce-payments/pull/5254

Feel free to re-open if you think this change should still go into `subscriptions-core`

PR #348, opened 2022-12-15
Check if the order actually exists before accessing order properties.

### LINKED ISSUE #344: Check if the order actually exists in wcs_order_contains_subscription()
## Describe the bug
The [function](https://github.com/Automattic/woocommerce-subscriptions-core/blob/trunk/includes/wcs-order-functions.php#L346) `wcs_order_contains_subscription()` contains a helper which retrieves the actual order in case a numeric value (or any non-order-type) is handled by the function. The function should at least check whether the order does actually exist before accessing getters as the function is used, e.g. in the Woo Subscriptions core plugin which uses it like this in `WC_Subscriptions_Payment_Gateways::get_available_payment_gateways( $available_gateways )`:

`if ( ! WC_Subscriptions_Cart::cart_contains_subscription() && ( ! isset( $_GET['order_id'] ) || ! wcs_order_contains_subscription( $_GET['order_id'] ) ) ) {
    return $available_gateways;
}`

This may lead to issues when the order does not exist.

### LINKED ISSUE #375: Error when completing pre-orders if subscriptions are also enabled
## Describe the bug

I'm experiencing an issue when using WooCommerce 7.2.2, WooCommerce Pre-Orders 1.8.0 and WooCommerce Subscriptions 4.7.0

Using the admin interface to complete pre-orders causes a fatal error in subscriptions-core.

Here is a screenshot of the error and stack trace
Screen Shot 2023-01-11 at 10 08 24 AM

## To Reproduce

1. Activate WooCommerce, WooCommerce Subscriptions, and WooCommerce Pre-Orders
2. Create a product available for pre-order
3. Place a pre-order
4. Use the admin panel to manually complete the pre-order

### Expected behavior

I would expect the pre-order to be completed, and not to interact with subscriptions at all.

### Actual behavior

> Fatal error: Uncaught Error: Call to a member function get_id() on bool in /woocommerce-subscriptions/vendor/woocommerce-subscriptions-core/includes/wcs-order-functions.php on line 352

WcPayWelcomePage from WooCommerce wants to know the available payment gateways.

The problem occurs when WC_Subscriptions_Payment_Gateways::get_available_payment_gateways() calls wcs_order_contains_subscription( $_GET['order_id'] ) without first verifying if $_GET['order_id'] is an integer.

https://github.com/Automattic/woocommerce-subscriptions-core/blob/5.1.0/includes/gateways/class-wc-subscriptions-core-payment-gateways.php#L101

When completing pre-orders, the order_id parameter is an array. For example in this specific case the parameters looked like: &order_id[]=300272&order_id[]=300244&order_id[]=300225

wcs_order_contains_subscription() expects to receive an integer or an order object. It eventually fails because it passes the same unverified order_id to wc_get_order() and never verifies if the returned value is an order.

https://github.com/Automattic/woocommerce-subscriptions-core/blob/5.1.0/includes/wcs-order-functions.php#L345-L352

## Product impact

- [ ] Does this issue affect WooCommerce Payments? I'm not using this so I'm not sure.

dennisnissle:
Fixes #344 

## Description
Check whether the order actually exists before accessing order properties in `wcs_order_contains_subscription()`.

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -8,6 +8,7 @@
 * Fix - When a customer changes their address on their account or subscription, make sure the new address is saved when HPOS is enabled.
 * Fix - Removed the potential for an infinite loop when getting a subscription's related orders while the subscription is being loaded.
 * Fix - Refactor `WC_Subscriptions_Renewal_Order::get_failed_order_replaced_by()` to support HPOS stores.
+* Fix - Check whether the order actually exists before accessing order properties in wcs_order_contains_subscription()
 * Dev - Replace code using get_post_type( $subscription_id ) with WC Data Store get_order_type().
 * Dev - Add subscriptions-core library version to the WooCommerce system status report.
 * Dev - Introduced a WCS_Object_Data_Cache_Manager and WCS_Object_Data_Cache_Manager_Many_To_One class as HPOS equivalents of the WCS_Post_Meta_Cache_Manager classes.
--- a/includes/wcs-order-functions.php
+++ b/includes/wcs-order-functions.php
@@ -344,6 +344,10 @@ function wcs_order_contains_subscription( $order, $order_type = array( 'parent',
 
 	if ( ! is_a( $order, 'WC_Abstract_Order' ) ) {
 		$order = wc_get_order( $order );
+
+		if ( ! $order ) {
+			return false;
+		}
 	}
 
 	$contains_subscription = false;


mattallan (approved):
Thanks @dennisnissle for this quick PR! Do you mind adding a changelog entry with the other v5.2 changes?
Once this is added it will be ready to merge.
Cheers

mattallan:
Thanks @dennisnissle . Merging this now :)

PR #349, opened 2022-12-19
Fixes updating all subscription payment methods when HPOS is enabled

### LINKED ISSUE #128: Active Subscriptions not updated when changing default payment method

mattallan:
Fixes #128 

>  **Note**
> To test this PR using WC Payments, you'll need to pull down these changes: https://github.com/Automattic/woocommerce-payments/pull/5254

## Description

While testing https://github.com/Automattic/woocommerce-payments/pull/5254 I noticed updating all payment methods for subscriptions was still broken on stores that have HPOS enabled.

After some digging I found we still have a `get_posts()` inside our `get_subscriptions_from_token()` function.
This PR updates this function to use `wcs_get_orders_with_meta_query()` so that this function works in both HPOS and non-HPOS. 

## How to test this PR

**Testing with both HPOS and non HPOS enabled:**

1. Purchase two active subscriptions with WC Payments
2. Go to My Account > Payment methods
3. Add a new payment method (i used card number`5555555555554444`)
4. After this card is added. Click "Make default"
5. You should see a notice to update your subscriptions to use this new payment method
6. Click "Yes"
7. Go to My Account > Subscriptions and confirm all of your active subscriptions have the new payment method

When testing on `trunk` and HPOS enabled, you won't even see the notice to update all of your subscriptions, this is because the `get_posts()` doesn't work in HPOS.

## Product impact

### DIFF
--- a/includes/class-wcs-payment-tokens.php
+++ b/includes/class-wcs-payment-tokens.php
@@ -102,28 +102,25 @@ class WCS_Payment_Tokens extends WC_Payment_Tokens {
 		$users_subscription_ids = WCS_Customer_Store::instance()->get_users_subscription_ids( $payment_token->get_user_id() );
 
 		if ( ! empty( $users_subscription_ids ) ) {
-			$meta_query = array(
-				array(
-					'key'   => '_payment_method',
-					'value' => $payment_token->get_gateway_id(),
-				),
-				array(
-					'key'   => '_requires_manual_renewal',
-					'value' => 'false',
-				),
-				array(
-					'value' => $payment_token->get_token(),
-				),
+			$subscription_ids = wcs_get_orders_with_meta_query(
+				[
+					'type'           => 'shop_subscription',
+					'status'         => [ 'wc-pending', 'wc-active', 'wc-on-hold' ],
+					'payment_method' => $payment_token->get_gateway_id(),
+					'meta_query'     => [ // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query
+						[
+							'key'   => '_requires_manual_renewal',
+							'value' => 'false',
+						],
+						[
+							'value' => $payment_token->get_token(),
+						],
+					],
+					'limit'          => -1,
+					'return'         => 'ids',
+					'post__in'       => $users_subscription_ids,
+				]
 			);
-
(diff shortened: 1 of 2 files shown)


includes/class-wcs-payment-tokens.php
```diff
@@ -102,29 +102,26 @@ public static function get_subscriptions_from_token( $payment_token ) {
...
+							'key'   => '_requires_manual_renewal',
+							'value' => 'false',
+						],
+						[
+							'value' => $payment_token->get_token(),
+						],
+					],
+					'limit'          => -1,
+					'return'         => 'ids',
+					'post__in'       => $users_subscription_ids,
```
shendy-a8c, line 121: Out of curiosity, can you explain why we have to use 'post' here? I tried to use `order__in` and it doesn't work.

shendy-a8c (approved):
Looks good!

includes/class-wcs-payment-tokens.php
mattallan: The `wc_get_orders()` function which we use in `wcs_get_orders_with_meta_query()` uses the `WP_Query` class for CPT and `OrdersTableQuery` when HPOS is enabled to query for orders by the given arguments. Both of these classes support the `post__in` argument.

mattallan:
Thank-you @shendy-a8c for the quick review on this one!

PR #372, opened 2023-01-13
[HPOS] Show Subscription status filters above list table when HPOS is enabled

mattallan:
Fixes #370 

## Description

With HPOS enabled, when you view the Subscriptions List Table, the status filters above the table only show WC Order statuses and not subscription statuses:

image

Issues in this screenshot:
 - The `all(1)` count is wrong (should be 3 not 1, isn't counting subscription statuses)
 - Status filters for `Active` and `Pending Cancellation` and all other subscription statuses are missing

To fix these issues, this PR uses the `views_woocommerce_page_wc-orders--shop_subscription` hook to attach the subscription status filters/views.

Changes in this PR:
 - New data store function (`get_subscriptions_count_by_status()`) to fetch subscriptions grouped by status along with the count
 - New function `filter_subscription_list_table_views()` which generates an array of [`{Subscription Status} (count)`]() links used to filter the subscriptions table

image

## How to test this PR

1. Enable HPOS 
2. Purchase a subscription and make sure the subscription has a non-order status like 'active'.
3. While on `trunk`, visit the **WooCommerce > Subscriptions** table and notice the list of statuses at the top of the table says "All (0)"
4. Check out this branch and refresh the page. You should see the status filters properly loading above the subscriptions list table.
5. Purchase more subscriptions and change the statuses to pending-cancelled and or on-hold.
6. Use the status (count) links to filter the subscriptions table by status.

## Product impact

- [x] Added changelog entry (or does not apply)
- [x] Will this PR affect WooCommerce Subscriptions? yes/no/tbc, add issue ref
- [x] Will this PR affect WooCommerce Payments? yes/no/tbc, add issue ref
- [ ]  Added deprecated functions, hooks or classes to the [spreadsheet](https://docs.google.com/spreadsheets/d/1xw9xszcPMnWsp4C8OKZMsLzZob7tOmWT7qMqmEIq314/edit#gid=0)

PR #374, opened 2023-01-13
Refactor WC_Subscriptions_Tracker to support HPOS

### LINKED ISSUE #147: Refactor the Subscriptions Tracker class to replace direct DB queries to fetch WC Order data from wp posts/postmeta tables
## Description

With Custom Order Tables being worked on by the WC Core team ([public announcement](https://developer.woocommerce.com/2022/01/17/the-plan-for-the-woocommerce-custom-order-table/)), we need to begin to think about updating the Subscriptions Core library to remove any code that makes direct DB queries to the post table when fetching Order related data.

In this issue, we're tackling the only uses of `$wpdb->get_col()` where the 'shop_order' post type is being queried.

## Code that needs updating
 - [x] [`WC_Subscriptions_Tracker::get_subscription_orders()`](https://github.com/Automattic/woocommerce-subscriptions-core/blob/2a0654bf0385cc3e5b097943aea5589e843b05d0/includes/class-wc-subscriptions-tracker.php#L127-L167)

## How to update code
Direct DB queries should be replaced with wc_get_orders()/WC_Order_Query, however I'm not exactly sure if every use case is as simple as that. If there's some complicated code/queries we may have to look at refactoring some code as part of tackling this issue.

## Product impact

## Dev notes

I've marked this as critical because if not implemented, we'll get incorrect/empty values sent to tracker for the number of switch, renewal and resubscribe orders on the store.

To call [`WC_Subscriptions_Tracker::add_subscriptions_tracking_data()`](https://github.com/Automattic/woocommerce-subscriptions-core/blob/9cc2b85f30c7ae18020fcff52a953fde4ca8b7a0/includes/class-wc-subscriptions-tracker.php#L32), run [`WC_Tracker::get_tracking_data()`](https://github.com/woocommerce/woocommerce/blob/2de5a8f08657a70987f48421dc43706890b96e4d/plugins/woocommerce/includes/class-wc-tracker.php#L117) with the WP Console plugin.  

## Additional context

Project thread: pdjTHR-Xp-p2

Jinksi:
Fixes #147

## Description

This PR replaces SQL queries and other non-CRUD WC functions present in the `WC_Subscriptions_Tracker` class with HPOS-compatible alternatives.

This class hooks into the WC `woocommerce_tracker_data` filter to calculate and send subscription-related data about a store (opted-in only) to [`tracking.woocommerce.com/v1/`](https://github.com/woocommerce/woocommerce/blob/2de5a8f08657a70987f48421dc43706890b96e4d/plugins/woocommerce/includes/class-wc-tracker.php#L28), [once per week by default](https://github.com/woocommerce/woocommerce/blob/2de5a8f08657a70987f48421dc43706890b96e4d/plugins/woocommerce/includes/class-wc-tracker.php#L48-L60).

You can prepare and view the data to be sent without sending by running `WC_Tracker::get_tracking_data()` in the [WP Console](https://wordpress.org/plugins/wp-console/) plugin or similar.

#### Note on data types

~The data type of some values in the output has changed from `str` to `float`/`int`.~

As we talked about below, all values will be returned as a string. https://github.com/Automattic/woocommerce-subscriptions-core/pull/374#issuecomment-1385320556

This data is processed by `wp_json_encode()` in [`WC_Tracker::send_tracking_data()`](https://github.com/woocommerce/woocommerce/blob/2de5a8f08657a70987f48421dc43706890b96e4d/plugins/woocommerce/includes/class-wc-tracker.php#L75) which keeps the dtypes intact (keeps an `int` an `int`).

In practice, the new approach outputting numbers as ~`int`/`float`~ `string` is more consistent and predictable. ~However, this _may_ be a concern for backwards compatibility purposes, depending on what the recipient of this data is expecting.~

The values are received by gh-woocommerce/woocommerce-tracking-api-service and converted to a string before storing (`WC_Tracker_API_V1::save_tracking_data()`).

## Todo

- [x] `get_subscription_orders()`
  - ~This is not working properly with HPOS enabled + sync disabled.~ Fixed in c957286.
  - TIL [`wcs_get_orders_with_meta_query()`](https://github.com/Automattic/woocommerce-subscriptions-core/blob/c95728673b2ed3d49a2d83e9085c445a6edc7b9b/includes/wcs-order-functions.php#L384). Refactor to use this helper function.
- [x] `get_subscription_dates()`
- [x] `get_subscription_counts()`. Replace `wp_count_posts()` with `get_subscriptions_count_by_status` from #372.

## How to test this PR

1. Enable WC Tracking in **WC Settings → Advanced → WooCommerce.com → Enable Tracking**.
2. Ensure your store has multiple subscriptions. Ideally, some are created in the past, including switch, renewal and resubscribes.
3. Start on `trunk` with HPOS disabled.
4. Run `WC_Tracker::get_tracking_data()` with the [WP Console](https://wordpress.org/plugins/wp-console/) plugin, noting what is output under `'extensions'=>'wc_subscriptions'`
5. Checkout this branch, enable HPOS + syncing (wait for the sync to complete).
6. Run `WC_Tracker::get_tracking_data()` again. Compare the output of `'extensions'=>'wc_subscriptions'` with what was output on `trunk`, ensuring they are identical.
   Note: run `WC_Tracker::get_tracking_data()['extensions']['wc_subscriptions']` for direct access to relevant data.

Example before (`trunk`):
```php
"subscription_orders" => [
    "switch_gross" => "0",
    "switch_count" => "3",
    "renewal_gross" => "200",
    "renewal_count" => "16",
    "resubscribe_gross" => "10",
    "resubscribe_count" => "1",
    "initial_gross" => 45,
    "initial_count" => 5,
  ],
```

After (this PR):

```php
"subscription_orders" => [
    "switch_gross" => 0.0,
    "switch_count" => 3,
    "renewal_gross" => 200.0,
    "renewal_count" => 16,
    "resubscribe_gross" => 10.0,
    "resubscribe_count" => 1,
    "initial_gross" => 45.0,
    "initial_count" => 5,
  ],
```

After `wp_json_encode()`:

```json
"subscription_orders": {
    "switch_gross": 0,
    "switch_count": 3,
    "renewal_gross": 200,
    "renewal_count": 16,
    "resubscribe_gross": 10,
    "resubscribe_count": 1,
    "initial_gross": 45,
    "initial_count": 5
  }
```

## Product impact

### DIFF
--- a/includes/class-wc-subscriptions-tracker.php
+++ b/includes/class-wc-subscriptions-tracker.php
@@ -156,8 +156,8 @@ class WC_Subscriptions_Tracker {
 				0
 			);
 
-			$order_totals[ $relation_type . '_count' ] = $count;
 			$order_totals[ $relation_type . '_gross' ] = $total;
+			$order_totals[ $relation_type . '_count' ] = $count;
 		}
 
 		// Finally, get the initial revenue and count.
--- a/includes/class-wc-subscriptions-tracker.php
+++ b/includes/class-wc-subscriptions-tracker.php
@@ -124,33 +124,20 @@ class WC_Subscriptions_Tracker {
 
 		// Get the subtotal and count for each subscription type.
 		foreach ( $relation_types as $relation_type ) {
-			$query_args = [
-				'type'       => 'shop_order',
-				'status'     => [ 'wc-completed', 'wc-processing', 'wc-refunded' ],
-				'limit'      => -1,
-				'meta_query' => [ // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query
-					[
-						'key'     => '_subscription_' . $relation_type,
-						'compare' => 'EXISTS',
+			// Get orders with the given relation type.
+			$relation_orders = wcs_get_orders_with_meta_query(
+				[
+					'type'       => 'shop_order',
+					'status'     => [ 'wc-completed', 'wc-processing', 'wc-refunded' ],
+					'limit'      => -1,
+					'meta_query' => [ // phpcs:ignore WordPress.DB.SlowDBQuery.slow_db_query_meta_query
+						[
+							'key'     => '_subscription_' . $relation_type,
+							'compare' => 'EXISTS',
+						],
 					],
-				],
-			];
-
-			// Prepare the handler for the query to get the orders for this relation type.
-			// Filter required when using CPT (HPOS disabled).
-			$relation_type_query_handler = function( $query ) use ( $relation_type ) {
-				$query['meta_query'][] = [
-					'key'     => '_subscription_' . $relation_type,
-					'compare' => 'EXISTS',
-				];
-				return $query;
-			};
-			add_filter( 'woocommerce_order_data_store_cpt_get_orders_query', $relation_type_query_handler, 10 );
-
-			// Fetch the orders.
-			$relation_orders = wc_get_orders( $query_args );
-
-			remove_filter( 'woocommerce_order_data_store_cpt_get_orders_query', $relation_type_query_handler, 10 );
+				]
+			);
 
 			// Sum the totals and count the orders.
 			$count = count( $relation_orders );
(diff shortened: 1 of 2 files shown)


includes/class-wc-subscriptions-tracker.php
```diff
@@ -170,28 +170,32 @@ private static function get_subscription_orders() {
...
-			"
-			SELECT
-				MIN( post_date_gmt ) as 'first', MAX( post_date_gmt ) as 'last'
-			FROM {$wpdb->prefix}posts
-			WHERE post_type = 'shop_subscription'
-			AND post_status NOT IN ( 'trash', 'auto-draft' )
-		",
-			ARRAY_A
+		// Ignore subscriptions with status 'trash'.
+		$first = wcs_get_subscriptions(
```
Jinksi, line 208: We're replacing a single SQL query with two queries using `wcs_get_subscriptions()`.

I'm unsure if this will have a noticeable impact on execution speed.

Is there a way to compress this into a single query using HPOS-compatible functions?

Jinksi: This is called [once per week by default](https://github.com/woocommerce/woocommerce/blob/2de5a8f08657a70987f48421dc43706890b96e4d/plugins/woocommerce/includes/class-wc-tracker.php#L48-L60), so it may not be a problem to run one extra query.

mattallan: @Jinksi I think these two queries are fine.

If we're concerned about performance, one option would be to do something similar to what we did in the `wcs_get_subscriptions_for_product()` function to support the direct DB queries in both CPT and HPOS environments.

For this particular query, we'd need to set variables `$date_column`, `$orders_tables_name`, `$order_type_column` and `$order_status_column` depending on whether HPOS is enabled.

I don't think there's a big concern around running these two queries versus one direct one. I kinda like using functions so that they're more future proof (hopefully 😄)

mattallan (commented):
> The data type of some values in the output has changed from str to float/int.

I've tested this PR with HPOS enabled and disabled and the main concern I have is the inconsistent data types when HPOS is enabled vs disabled.
For example, the subscription status count is returned as a string with HPOS disabled and as an integer when HPOS is enabled.

I'm not sure what impact this has with there this tracker data is sent but I'm thinking we should try keep it consistent.
Thoughts on casting some of these values to strings? 🙈 

Something I noticed is WooCommerce Core hasn't updated their tracker class to support HPOS yet, I wonder what approach they will use.

Jinksi:
Thanks @mattallan!

> Thoughts on casting some of these values to strings? 🙈

I agree, we should make these data types consistent. It is currently a bit of a mix of string/int on `trunk`, example below. 

<details>
<summary>
Example output showing mix of data types from `trunk`
</summary>

```php
  "subscriptions" => [
    "first" => "2022-02-17 05:28:05",
    "last" => "2023-01-11 06:27:48",
    "wc-pending" => 0,
    "wc-active" => "1",
    "wc-on-hold" => "13",
    "wc-cancelled" => "2",
    "wc-switched" => 0,
    "wc-expired" => 0,
    "wc-pending-cancel" => 0,
  ],
  "subscription_orders" => [
    "switch_gross" => "0",
    "switch_count" => "4",
    "renewal_gross" => "420.2",
    "renewal_count" => "21",
    "resubscribe_gross" => "22",
    "resubscribe_count" => "1",
    "initial_gross" => 223.2,
    "initial_count" => 13,
  ],
```
</details>

Since the current/`trunk` implementation prefers returning strings for 'truthy' values, I'd say we should send all numeric values as a string. Either way (string or int/float), it will be an improvement to be explicit and consistent.

Jinksi:
> Something I noticed is WooCommerce Core hasn't updated their tracker class to support HPOS yet, I wonder what approach they will use.

WC does not help guide us towards a consistent approach here either; here are the product counts from `WC_Tracker::get_tracking_data()`:

```php
"products" => [
    "total" => "11",
    "external" => 0,
    "grouped" => 0,
    "simple" => 7,
    "subscription" => 2,
    "variable" => 1,
    "Variable Subscription" => 1,
    "yith_bundle" => 0,
  ],
```

mattallan:
> WC does not help guide us towards a consistent approach here either; here are the product counts

Oh wow, it's so strange how inconsistent this data is 😭 I'm generally okay with it being a bit mix-and-match as long as it doesn't impact how the tracking data is used internally.

Happy with whatever you think is the right approach!

Jinksi:
@mattallan I think strings are a safe bet, enforced in a6e7012.

I've found the data recipient at gh-woocommerce/woocommerce-tracking-api-service, where it is converted to a string before storing (`WC_Tracker_API_V1::save_tracking_data()`).

mattallan (approved):
Thanks @Jinksi! The latest changes LGTM. I retested this PR with HPOS enabled and disabled and confirmed the values are casted to a string consistently . I also confirmed in gh-woocommerce/woocommerce-tracking-api-service they're also casting to a string so this should be fine.

haszari:
> I think strings are a safe bet, enforced in https://github.com/Automattic/woocommerce-subscriptions-core/commit/a6e7012004d59cbc6ed86b3de5f9c287775de8e3.

I agree, the data is ultimately sent as json (i.e. a string / text format) so if some numbers are rendered to string early I don't think that's a problem at all 😁

Jinksi:
> the data is ultimately sent as json

Although, a value's dtype before encoding to `json` will be recovered when decoding.

So if the recipient was assuming a value to be of type `float` or `int` _after decoding_ and we had changed it to a string, 🐛 😞

PR #381, opened 2023-01-17
[HPOS] Fixes the Cancel and remove personal data subscription bulk action on HPOS stores.

### LINKED ISSUE #376: Using the "Cancel and remove personal data" bulk action does not cancel the subscription on HPOS
## Describe the bug

In #367, we have addressed issues with bulk actions not working with HPOS enabled but while testing this PR I noticed the "Cancel and remove personal data" action wasn't cancelling the subscription.

With HPOS enabled, using this bulk action keeps the subscription status but still removes the data:

image

What's causing this is WooCommerce's new [`ListTable.php` class](https://github.com/woocommerce/woocommerce/blob/235709ebf7f61b8b53065048f637067d8856f558/plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php) is handling the `remove_personal_data` action [here](https://github.com/woocommerce/woocommerce/blob/235709ebf7f61b8b53065048f637067d8856f558/plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php#L1140) for both Orders and Subscriptions, however, we want our own [`WCS_Privacy`](https://github.com/Automattic/woocommerce-subscriptions-core/blob/trunk/includes/privacy/class-wcs-privacy.php) to handle 

## To Reproduce

1. With HPOS enabled, purchase a subscription
2. From the WP admin navigate to **WooCommerce > Subscriptions**
3. Select an active subscription and use the bulk action to "Cancel and remove personal data"
4. Notice the subscriptions data has been anonymized but the subscription is still active

## Product impact

mattallan:
Fixes #376 

## Description

image

While testing PR https://github.com/Automattic/woocommerce-subscriptions-core/pull/367 I found an issue with the **Cancel and remove personal data** not working on stores with HPOS enabled. This was caused by the new `ListTable` class handling this bulk action request for Orders as well as subscriptions.

To fix this, we need to introduce a new action name to prevent WooCommerce from handling it automatically. To do this, this PR adds a new action called `wcs_remove_personal_data` and then handles this action manually using the `handle_bulk_actions-woocommerce_page_wc-orders--shop_subscription` action hook.

## How to test this PR

1. Enable HPOS on your store.
2. Purchase multiple subscriptions on the checkout
3. Go to the Subscriptions list table
4. Select one or multiple subscriptions and use the `Cancel and remove personal data` action
5. Confirm the subscriptions are canceled with all their data removed.
6. Confirm the admin notice is shown
7. Run the same checks with HPOS disabled.

## Product impact

- [x] Added changelog entry. No changelog needed for this PR as we can use the same changelog used for #376 

### DIFF
--- a/assets/js/admin/admin.js
+++ b/assets/js/admin/admin.js
@@ -1370,4 +1370,16 @@ jQuery( function ( $ ) {
 		},
 	};
 	wcs_accounts_and_privacy_settings.init();
+
+	$( '#wpbody' ).on( 'click', '#doaction, #doaction2', function () {
+		var action = $( this ).is( '#doaction' )
+			? $( '#bulk-action-selector-top' ).val()
+			: $( '#bulk-action-selector-bottom' ).val();
+
+		if ( 'wcs_remove_personal_data' === action ) {
+			return window.confirm(
+				WCSubscriptions.i18n_remove_personal_data_notice
+			);
+		}
+	} );
 } );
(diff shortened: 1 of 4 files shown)


james-allan (approved):
LGTM. 👍 

I tested using this action on both HPOS and CPT stores. 

I confirmed that there's a confirm box and the subscriptions are cancelled and annoyimized in both scenarios.

PR #386, opened 2023-01-19
Fix "Subscriptions by Payment Gateway" in WC → Status when HPOS is enabled

Jinksi:
Fixes #384

## Description

This PR refactors the `WCS_Admin_System_Status::get_subscriptions_by_gateway()` to provide the correct subscription count by status & payment gateway when HPOS is enabled.

This function is used when rendering **Subscriptions by Payment Gateway** in **WooCommerce → Status** and the [WCS Rest API](https://github.com/woocommerce/woocommerce-subscriptions/blob/a2831c03c92bf4d28c87e0f01aeb09844f355193/includes/api/class-wc-rest-subscription-system-status-manager.php#L39).

## How to test this PR

1. Enable HPOS + no syncing.
2. Purchase a few subscriptions with 1 or more payment gateways.
3. Change the status of 1 or more subscriptions (e.g. move to trash, cancel, place on hold).
5. Go to **WooCommerce → Status** and find **Subscriptions by Payment Gateway**.
6. Ensure these values are correct by comparing with **WooCommerce → Subscriptions** and filtering by Payment Gateway.
7. Enable HPOS syncing and wait for the sync to complete.
8. Ensure the **WooCommerce → Status → Subscriptions by Payment Gateway** are correct.
7. Disable HPOS.
8. Ensure the **WooCommerce → Status → Subscriptions by Payment Gateway** are correct.

**WCS REST API**

I have tested the output provided to the `/wp-json/wc/v3/system_status` endpoint when WCS is enabled, and can confirm that the results are correct with both HPOS enabled/disabled. 

This can be tested (optional) by:
1. Enable the [WooCommerce Rest API](https://woocommerce.com/document/woocommerce-rest-api/).
1. Using Postman or similar, prepare a GET request, setting up Basic Auth as per guide above.
   1. If not using HTTPS, replace `is_ssl()` with a truthy value [on this line within WC](https://github.com/woocommerce/woocommerce/blob/2de5a8f08657a70987f48421dc43706890b96e4d/plugins/woocommerce/includes/class-wc-rest-authentication.php#L82).
1. Send a `GET` request to `/wp-json/wc/v3/system_status`.
2. The relevant values will be output under `subscriptions.subscriptions_by_payment_gateway`.

Example:

```json
"subscriptions_by_payment_gateway": {
    "square_credit_card": {
        "wc-active": "1",
        "wc-cancelled": "1",
        "wc-on-hold": "2"
    },
    "woocommerce_payments": {
        "trash": "1",
[...]

PR #387, opened 2023-01-19
Fix PHPCS violations in `WCS_Admin_System_Status` to improve code quality

Jinksi:
Based on #386. Please wait for it to be merged before reviewing/merging.

## Description

This PR only fixes PHPCS violations in the `WCS_Admin_System_Status` class to improve code quality. No changes to the underlying logic are expected.

This class is used to add details to **WooCommerce → Status**.

## How to test this PR

1. Run `./vendor/bin/phpcs includes/admin/class-wcs-admin-system-status.php` in your terminal. On `trunk` you should see errors/warnings. On this branch, you should not see any violations.
   Note: running phpcs directly from `./vendor/bin/` skips the "only changed LOC" rule that we use for PHPCS GH checks.

## Product impact

### DIFF
--- a/includes/admin/class-wcs-admin-system-status.php
+++ b/includes/admin/class-wcs-admin-system-status.php
@@ -260,7 +260,7 @@ class WCS_Admin_System_Status {
 
 		foreach ( $woocommerce_account_subscriptions as $subscription ) {
 			if ( isset( $subscription['product_id'] ) && self::WCS_PRODUCT_ID === $subscription['product_id'] ) {
-				$has_active_product_key = in_array( $site_id, $subscription['connections'], true );
+				$has_active_product_key = in_array( $site_id, $subscription['connections'], false ); // phpcs:ignore WordPress.PHP.StrictInArray.FoundNonStrictFalse
 				break;
 			}
 		}
--- a/includes/admin/class-wcs-admin-system-status.php
+++ b/includes/admin/class-wcs-admin-system-status.php
@@ -176,7 +176,7 @@ class WCS_Admin_System_Status {
 		$templates        = WC_Admin_Status::scan_template_files( $wcs_template_dir );
 
 		foreach ( $templates as $file ) {
-			$theme_file = $is_outdated = false;
+			$theme_file = false;
 			$locations  = array(
 				get_stylesheet_directory() . "/{$file}",
 				get_stylesheet_directory() . "/{$wc_template_path}{$file}",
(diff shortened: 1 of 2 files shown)


includes/admin/class-wcs-admin-system-status.php
```diff
@@ -32,8 +31,9 @@ public static function init() {
 	 * @since 1.0.0 - Migrated from WooCommerce Subscriptions v2.3.0
 	 */
 	public static function render_system_status_items() {
-
-		$store_data = $subscriptions_data = $subscriptions_by_payment_gateway_data = $payment_gateway_data = array();
```
Jinksi, line 36: ~The `$subscriptions_by_payment_gateway_data` variable was unused, so I've removed it.~

I was wrong here; it was intended to be used, so I've corrected this in de7494d.

Took me a while to realise the arg was passed by reference `( &$variable_name )`. This feature of PHP feels... obscure 😆

includes/admin/class-wcs-admin-system-status.php
```diff
@@ -177,7 +176,7 @@ private static function get_theme_overrides() {
 		$templates        = WC_Admin_Status::scan_template_files( $wcs_template_dir );
 
 		foreach ( $templates as $file ) {
-			$theme_file = $is_outdated = false;
+			$theme_file = false;
```
Jinksi, line 181: `$is_outdated` was unused, so I've removed it.

includes/admin/class-wcs-admin-system-status.php
```diff
@@ -261,7 +260,7 @@ private static function set_woocommerce_account_data( &$debug_data ) {
 
 		foreach ( $woocommerce_account_subscriptions as $subscription ) {
 			if ( isset( $subscription['product_id'] ) && self::WCS_PRODUCT_ID === $subscription['product_id'] ) {
-				$has_active_product_key = in_array( $site_id, $subscription['connections'] );
+				$has_active_product_key = in_array( $site_id, $subscription['connections'], false ); // phpcs:ignore WordPress.PHP.StrictInArray.FoundNonStrictFalse -- In case the value from $subscription['connections'] is a string.
```
Jinksi, line 265: I've left the non-strict comparison for this call to `in_array`, in case the value from $subscription['connections'] is a string.

includes/admin/class-wcs-admin-system-status.php
mattallan: Oh wow, we very rarely use this approach nowadays but good pick up on the incorrect variable name being created here and never used.

mattallan (approved):
Thanks @Jinksi these changes LGTM.

PR #388, opened 2023-01-19
Replace `wp_count_posts( 'shop_subscription' )` with new data store function to work with HPOS

### LINKED ISSUE #382: Replace wp_count_posts() uses with subscriptions data store equivalent to support HPOS
## Description

In https://github.com/Automattic/woocommerce-subscriptions-core/pull/372 we introduced a new subscriptions data store function to count subscriptions and group them by their statuses.

This issue is to look at replacing all instances of `wp_count_posts( 'shop_subscription' )` with the new `WC_Data_Store::load( 'subscription' )->get_subscriptions_count_by_status();`.

---

**Note:** There is one instance of `wp_count_posts( 'shop_order' )` used in the `wcs_is_large_site()` which won't work on stores with HPOS enabled, but there's no replacement function for this.

For now, I'm thinking we only worry about counting orders when the store has HPOS disabled since the idea behind this function is that we throttle the store based on how many orders and subscriptions they have, but performance issues might not be a think on HPOS sites.

So the logic in `wcs_is_large_site()` would change to something like:

```
// If an option has been set previously, convert it to a bool.
if ( false !== $is_large_site ) {
	$is_large_site = wc_string_to_bool( $is_large_site );
} elseif ( array_sum( (array) wp_count_posts( 'shop_subscription' ) ) > 3000 || ( ! wcs_is_custom_order_tables_usage_enabled() && array_sum( (array) wp_count_posts( 'shop_order' ) ) > 25000 ) ) {
	$is_large_site = true;
	update_option( 'wcs_is_large_site', wc_bool_to_string( $is_large_site ), false );
} else {
	$is_large_site = false;
}
```
## Product impact

mattallan:
Fixes #382 

## Description

In #372 we introduced a new data store function that returns an array of subscription statuses and the number of subscriptions on the store with that status:

```
WC_Data_Store::load( 'subscription' )->get_subscriptions_count_by_status()
```

In this PR, I'm updating all uses of `wp_count_posts()` with this new data store function.

**Notes**

- The `wcs_is_large_site()` was also counting the number of orders on a store but there's currently not HPOS equivalent function for this so for the time being I've opted into just leaving this count out of the condition when HPOS is enabled.
- There's a `wp_count_posts()` calls in the repair script files that I didn't bother updating in this PR because it's a repair script specifically for stores upgrading from 2.0.0 to 2.0.2

## How to test this PR

To test the Admin System Status changes:

1. Enable HPOS and purchase a handful of subscriptions
2. Change some statuses of the subscriptions
3. View **WooCommerce > Status**
4. Scroll down to Subscriptions section and look at the `Subscription Statuses:` row.
5. Disable HPOS and purchase some more subscriptions to test with HPOS disabled.

For the rest of the changes I ran the following snippet in WP Console and tested with HPOS enabled and disabled:

```
echo 'WC_Subscriptions::get_subscription_status_counts(): ' . var_export( WC_Subscriptions::get_subscription_status_counts(), true ) . "\n";

echo "\n\n";
delete_option( 'wcs_is_large_site' );

echo 'wcs_is_large_site(): ' . var_export( wcs_is_large_site(), true );

echo "\n\n";

echo 'WCS_Admin_System_Status::get_subscription_statuses(): ' . var_export( WCS_Admin_System_Status::get_subscription_statuses(), true );
```

## Product impact

### DIFF
--- a/includes/data-stores/class-wcs-subscription-data-store-cpt.php
+++ b/includes/data-stores/class-wcs-subscription-data-store-cpt.php
@@ -598,6 +598,6 @@ class WCS_Subscription_Data_Store_CPT extends WC_Order_Data_Store_CPT implements
 	 * @return array
 	 */
 	public function get_subscriptions_count_by_status() {
-		return WC_Data_Store::load( 'subscription' )->get_subscriptions_count_by_status();
+		return (array) wp_count_posts( 'shop_subscription' );
 	}
 }
--- a/includes/data-stores/class-wcs-subscription-data-store-cpt.php
+++ b/includes/data-stores/class-wcs-subscription-data-store-cpt.php
@@ -598,6 +598,6 @@ class WCS_Subscription_Data_Store_CPT extends WC_Order_Data_Store_CPT implements
 	 * @return array
 	 */
 	public function get_subscriptions_count_by_status() {
-		return (array) wp_count_posts( 'shop_subscription' );
+		return WC_Data_Store::load( 'subscription' )->get_subscriptions_count_by_status();
 	}
 }
--- a/changelog.txt
+++ b/changelog.txt
@@ -26,6 +26,7 @@
 * Fix - On HPOS environments, handle the admin subscriptions list table bulk actions and row actions in a HPOS compatible way.
 * Fix - On HPOS environments, admin notices might show incorrect subscription or order link as part of the admin notice.
 * Fix - On HPOS environments, parent order link on subscription edit page, in subscription details box, is incorrect.
+* Fix - Replace code using wp_count_posts( 'shop_subscription' ) with data store function to count subscriptions grouped by statuses on stores with HPOS enabled.
 * Dev - Replace code using get_post_type( $subscription_id ) with WC Data Store get_order_type().
 * Dev - Add subscriptions-core library version to the WooCommerce system status report.
 * Dev - Introduced a WCS_Object_Data_Cache_Manager and WCS_Object_Data_Cache_Manager_Many_To_One class as HPOS equivalents of the WCS_Post_Meta_Cache_Manager classes.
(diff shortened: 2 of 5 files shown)


includes/data-stores/class-wcs-subscription-data-store-cpt.php
```diff
@@ -598,6 +598,6 @@ public function delete_all_metadata_by_key( $meta_key ) {
 	 * @return array
 	 */
 	public function get_subscriptions_count_by_status() {
-		return (array) wp_count_posts( 'shop_subscription' );
+		return WC_Data_Store::load( 'subscription' )->get_subscriptions_count_by_status();
```
Jinksi: In testing, it looks like this function is calling itself recursively, resulting in the page endlessly loading.
This occurs when HPOS is disabled (using the WordPress posts table) and rendering the **WC → Status** page.

Jinksi (changes requested):
@mattallan These changes look good and test well with HPOS enabled 👍

But I've found an issue with HPOS disabled, as commented below.

includes/data-stores/class-wcs-subscription-data-store-cpt.php
mattallan: Biggest facepalm 🤦‍♂️ This definitely shouldn't have been updated

mattallan: Fixed in [7d87fae](https://github.com/Automattic/woocommerce-subscriptions-core/pull/388/commits/7d87fae5f0d8eb95447829202b9e1c69a53c6665)

Jinksi (approved):
LGTM 🙌 this tests well now with both HPOS enabled/disabled.

PR #414, opened 2023-03-03
Attach callback to the subscription_trial_end action scheduler hook to prevent action scheduler triggering a failed action

### LINKED ISSUE #255: Prevent noisy false positive errors when subscription trial period ends (due to recent change in Action Scheduler)
## Describe the bug
When a trial ends, WCS triggers action `woocommerce_scheduled_subscription_trial_end` which is empty by default. WCS doesn't need to do any processing, the action is there for extensibility.

In recent Action Scheduler, empty callbacks trigger an error. 

Our error handling code is showing this error to merchants in the dashboard, which is not necessary (as it's not causing any problems). See [WCS_Failed_Scheduled_Action_Manager::log_action_scheduler_failure()](https://github.com/Automattic/woocommerce-subscriptions-core/blob/trunk/includes/class-wcs-failed-scheduled-action-manager.php#L77).

#### Proposed fix
[WCS_Failed_Scheduled_Action_Manager::log_action_scheduler_failure()](https://github.com/Automattic/woocommerce-subscriptions-core/blob/trunk/includes/class-wcs-failed-scheduled-action-manager.php#L77) should have special handling to not log/store failed trial end action scheduler errors that contain the message "not be executed as no callbacks are registered.".

##### Original description

A change to Action Scheduler now throws an error if a scheduled action has no callbacks. https://github.com/woocommerce/action-scheduler/pull/806
WooCommerce Subscriptions adds the action `woocommerce_scheduled_subscription_trial_end` which doesn't do anything by itself. It's there so other extensions can add something to happen before a trial ends. 

Since the Action Scheduler update to 3.5.0, we've seen sporadic reports of failed scheduled actions. The trial is ending, but it's throwing an error in Subscriptions. This causes the `html-failed-scheduled-action-notice.php` to display on admin pages and logs an error. 

`10-19-2022 @ 10:05:18 - scheduled action 55915 (subscription trial end) failed to finish processing due to the following exception: Scheduled action for woocommerce_scheduled_subscription_trial_end will not be executed as no callbacks are registered.
`

I'm not entirely certain if the action should be changed to prevent the error or if the failed scheduled action notice should omit these or have a different error message. But we're getting more tickets in Happiness about failed actions related to `woocommerce_scheduled_subscription_trial_end`, and this change to AS appears to be the root. 

## To Reproduce

1. Create a subscription with a free trial
1. Wait until the trial ends (or manually run the action for `woocommerce_scheduled_subscription_trial_end`)
1. See the error as a failed scheduled action

### Expected behavior
Since the trial ended successfully, I don't expect to see an error displayed on admin pages and in the error logs. 

### Actual behavior

An error displays that causes the merchant to believe the trial end didn't process correctly. 

## Product impact

- [X] Does this issue affect WooCommerce Payments? yes/no/tbc, add issue ref

Confirmed that the action is in Subscriptions Core, so it will impact WooCommerce Subscriptions and WooCommerce Payments Subscriptions.

## Additional context

A few tickets: 
5680379-zen
5650349-zen
5672620-zen 
5692989-zen

felipevelzani:
> **Note** - From @mattallan
>
> While reviewing this PR we discovered that the error message coming Action Scheduler (AS) is translatable which means it can't reliably by used in a string comparison. To workaround the issue stemming from AS we decided to change the approach and attach a callback to Subscriptions that triggers a subscription-specific trial-ended hook.
>
> Testing instructions on this PR remain the same for the new changes.

---

Fixes #255

## Description

Some scheduled actions don't necessarily have hooks associated with them and have only been added for extensibility, such as `woocommerce_scheduled_subscription_trial_end`.

Action Scheduler 3.5.x introduced a new error message for empty callbacks, but we wouldn't want to log these as errors on WooCommerce Subscriptions Core, so we changed the code accordingly. 

For reference, followed @mattallan [suggestion here](https://github.com/Automattic/woocommerce-subscriptions-core/issues/255#issuecomment-1378034683):

> What I think we'll need to do is update our [WCS_Failed_Scheduled_Action_Manager::log_action_scheduler_failure()](https://github.com/Automattic/woocommerce-subscriptions-core/blob/trunk/includes/class-wcs-failed-scheduled-action-manager.php#L77) function to have special handling to not log/store failed trial end action scheduler errors that contain the message "not be executed as no callbacks are registered.". 

## How to test this PR

1. Create a subscription with a free trial
1. Wait until the trial ends (or manually run the action for `woocommerce_scheduled_subscription_trial_end`)
1. See the error as a failed scheduled action

## Product impact

### DIFF
--- a/includes/class-wcs-failed-scheduled-action-manager.php
+++ b/includes/class-wcs-failed-scheduled-action-manager.php
@@ -83,8 +83,8 @@ class WCS_Failed_Scheduled_Action_Manager {
 
 		$subscription_action = $this->get_action_hook_label( $action->get_hook() );
 
-		// Not log when there is not a hook associated with the scheduled action
-		if ( strpos( $error->getMessage(), 'not be executed as no callbacks are registered' ) ) {
+		// Don't log errors for Trial End and Payment Rety scheduled actions when there are no callbacks associated with the scheduled action.
+		if ( strpos( $error->getMessage(), 'not be executed as no callbacks are registered' ) && in_array( $action->get_hook(), [ 'woocommerce_scheduled_subscription_trial_end', 'woocommerce_scheduled_subscription_payment_retry' ], true ) ) {
 			return;
 		}
 
--- a/includes/class-wcs-failed-scheduled-action-manager.php
+++ b/includes/class-wcs-failed-scheduled-action-manager.php
@@ -85,7 +85,7 @@ class WCS_Failed_Scheduled_Action_Manager {
 
 		// Not log when there is not a hook associated with the scheduled action
 		if ( strpos( $error->getMessage(), 'not be executed as no callbacks are registered' ) ) {
-					return;
+			return;
 		}
 
 		switch ( current_filter() ) {
--- a/includes/class-wcs-failed-scheduled-action-manager.php
+++ b/includes/class-wcs-failed-scheduled-action-manager.php
@@ -83,6 +83,11 @@ class WCS_Failed_Scheduled_Action_Manager {
 
 		$subscription_action = $this->get_action_hook_label( $action->get_hook() );
 
+		// Not log when there is not a hook associated with the scheduled action
+		if ( strpos( $error->getMessage(), 'not be executed as no callbacks are registered' ) ) {
+					return;
+		}
+
 		switch ( current_filter() ) {
 			case 'action_scheduler_failed_action':
 				$this->log( sprintf( 'scheduled action %s (%s) failed to finish processing after %s seconds', $action_id, $subscription_action, absint( $error ) ) );
@@ -93,11 +98,6 @@ class WCS_Failed_Scheduled_Action_Manager {
 			case 'action_scheduler_unexpected_shutdown':
 				$this->log( sprintf( 'scheduled action %s (%s) failed to finish processing due to the following error: %s', $action_id, $subscription_action, $error['message'] ) );
 				break;
-		}
-
-		// Not log when there is not a hook associated with the scheduled action
-		if ( strpos( $error->getMessage(), 'not be executed as no callbacks are registered' ) ) {
-			return;
 		}
 
 		$this->log( sprintf( 'action args: %s', $this->get_action_args_string( $action->get_args() ) ) );
--- a/includes/class-wcs-failed-scheduled-action-manager.php
+++ b/includes/class-wcs-failed-scheduled-action-manager.php
@@ -95,6 +95,11 @@ class WCS_Failed_Scheduled_Action_Manager {
 				break;
 		}
 
+		// Not log when there is not a hook associated with the scheduled action
+		if ( strpos( $error->getMessage(), 'not be executed as no callbacks are registered' ) ) {
+			return;
+		}
+
 		$this->log( sprintf( 'action args: %s', $this->get_action_args_string( $action->get_args() ) ) );
 
 		// Store information about the scheduled action for displaying an admin notice
--- a/includes/class-wcs-failed-scheduled-action-manager.php
+++ b/includes/class-wcs-failed-scheduled-action-manager.php
@@ -83,11 +83,6 @@ class WCS_Failed_Scheduled_Action_Manager {
 
 		$subscription_action = $this->get_action_hook_label( $action->get_hook() );
 
-		// Don't log errors for Trial End and Payment Rety scheduled actions when there are no callbacks associated with the scheduled action.
-		if ( strpos( $error->getMessage(), 'not be executed as no callbacks are registered' ) && in_array( $action->get_hook(), [ 'woocommerce_scheduled_subscription_trial_end', 'woocommerce_scheduled_subscription_payment_retry' ], true ) ) {
-			return;
-		}
-
 		switch ( current_filter() ) {
 			case 'action_scheduler_failed_action':
 				$this->log( sprintf( 'scheduled action %s (%s) failed to finish processing after %s seconds', $action_id, $subscription_action, absint( $error ) ) );
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Subscriptions Core Changelog ***
 
+= 5.5.0 - 2023-xx-xx =
+* Fix - Prevent admin error notices being shown for the "subscription trial end" event that was caused by no callbacks being attached to this scheduled action. #414
+
 = 5.4.0 - 2023-02-21 =
 * Fix - Remove the recurring shipping method cache that caused bugs for third-party plugins like Conditional Shipping and Payments. #407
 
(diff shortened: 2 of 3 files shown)


includes/class-wcs-failed-scheduled-action-manager.php
```diff
@@ -83,8 +83,8 @@ public function log_action_scheduler_failure( $action_id, $error ) {
 
 		$subscription_action = $this->get_action_hook_label( $action->get_hook() );
 
-		// Not log when there is not a hook associated with the scheduled action
-		if ( strpos( $error->getMessage(), 'not be executed as no callbacks are registered' ) ) {
+		// Don't log errors for Trial End and Payment Rety scheduled actions when there are no callbacks associated with the scheduled action.
+		if ( strpos( $error->getMessage(), 'not be executed as no callbacks are registered' ) && in_array( $action->get_hook(), [ 'woocommerce_scheduled_subscription_trial_end', 'woocommerce_scheduled_subscription_payment_retry' ], true ) ) {
```
haszari: Is there a programmatic way to detect the "no callbacks" error? Ideally we wouldn't have to compare with an error message string fragment – fragile.

mattallan: @haszari Doesn't seem like there is as they just throw a standard exception. Here's a reference to the code in Action Scheduler: https://github.com/woocommerce/action-scheduler/blob/520790a8f770c900427e5bb350a2c8a4e86168e6/classes/actions/ActionScheduler_Action.php#L37

mattallan: We could add an additional check for `! has_action( $subscription_action_hook_name )` to confirm there's no callbacks but we would still need to do a string comparison on the error message to know what error we're checking against

I agree it's fragile but not sure of other solutions.

haszari: Thanks for checking @mattallan. Is it possible we're doing something wrong here? Or is this something we should feed back about – that exception looks harsh at face value, it's informational, there's no danger (no exception IMO!).

mattallan: > Is it possible we're doing something wrong here?

Do you mean: is Subscriptions using AS in an unintended way? Or is this change made to AS too harsh?

I don't think we're necessarily doing anything wrong in Subscriptions as the trial end period action is more for 3rd parties to hook onto i.e. the Memberships plugin or email services/plugins to send emails to customers

Another option to solve this would be for us to attach our own callback that just fires off a different hook name for the trial ending event that 3rd parties could use 🤔 

i.e. something like:

```
add_action( 'woocommerce_scheduled_subscription_trial_end', 'trigger_subscription_trial_end_event' );

function trigger_subscription_trial_end_event( $subscription_id ) {
    do_action( 'woocommerce_subscriptions_trial_end', $subscription_id );
}
```

Seems a bit silly to do this just to suppress the errors from AS but I'm open to ideas!!

> that exception looks harsh at face value, it's informational, there's no danger (no exception IMO!).

For more context, the exception is caught on Action Scheduler's side and then the error/exception is sent on the `action_scheduler_failed_execution` hook which we're hooked onto in Subscriptions to handle these error.

In the future there could be errors from AS that we want to ignore so it would be nice for AS to tell us the type of error via some key/ID so we don't have to check strings.

brucealdridge: It would be good to update the generic exception in Action Scheduler to a custom exception to allow for easier detection and handling. Something like `ActionScheduler_MissingHooksException`.

Then detection would be 
```php
if ( $error instanceof ActionScheduler_MissingHooksException ) { 
   ...
}
```

brucealdridge: However, we'd still have to check for the specific string for backwards compatibility.

haszari: > Do you mean: is Subscriptions using AS in an unintended way? Or is this change made to AS too harsh?

That's exactly what I mean. Are we doing things in a way that AS doesn't expect (e.g. logging all exceptions) or does AS need to better understand our use case? (or  a combination of both!)

> In the future there could be errors from AS that we want to ignore so it would be nice for AS to tell us the type of error via some key/ID so we don't have to check strings.

That would work, but IMO would be overengineering. Ideally we wouldn't need to selectively ignore any exceptions – that sounds like a patch on a bad API. Maybe the exceptions should have levels – i.e. the ones that are causing pain or signaling a problem vs. others that are informational or "something maybe fishy here".

Your approach of using a custom hook sounds elegant to me. I also think it would be reasonable to leave this as is and educate merchants/3PDs on what it means – i.e. it's informational. 

Anyway, up to y'all how we handle this, but it sounds like we should chat with AS squad to make sure we're aligned. We shouldn't have to patch or work around core – we're on the same team. Maybe we could ask for a review from that team on this issue?

mattallan: I just noticed the error message from Action Scheduler is translatable, so we won't be able to use this approach that we're using in this PR anyway 😢 

So I think we go with my other approach which is adding a callback for the trial end hook inside WC Subscriptions to _silence_ these "no callback errors"

> Something like ActionScheduler_MissingHooksException.

Yeah that would be nice if it were introduced at the beginning when they added this ! has_action check, but now we'd need to support the generic and non-generic exception which would be kind of messy

haszari: There's no particular urgency on solving this – so if we want an upstream fix, we can lobby for that and follow up later. 

Though @mattallan's idea sounds lean and elegant.

mattallan:
@felipelousantos thank you for opening this PR to improve compatibility with Subscriptions and the latest Action Scheduler changes.

During code review on this PR we found that the error message coming from action scheduler is translatable, and while it's extremely unlikely to be translated by anyone, we can't reliably use it in a string comparison. To workaround the issue of Action Scheduler throwing an exception if a hook doesn't have a callback, I have just pushed up some changes to this PR so that Subscriptions attaches a callback to the trial_end hook and fires our own hook.

I have tested this on the latest WC version and confirm it solves the issue.

To keep this moving, @brucealdridge would you mind giving this one a review so that we can get this in the next release? Thanks! :)

mattallan:
Just making a note that I've updated the title to better reflect the latest changes and will also add a note to the description.

brucealdridge (approved):
I think we should add a changelog entry for visibility of the fix.

```
 * Dev - Add woocommerce_subscription_trial_ended hook to prevent action scheduler errors.
```

mattallan:
Thanks @brucealdridge, oh I just added a changelog entry before seeing your latest comment. I was having difficulties writing this one so I'm curious to get your thoughts. Maybe @haszari can chime in as well! 😄

brucealdridge (approved):

haszari:
@brucealdridge changelog looks good to me. A good changelog captures info relating to these questions:

- What changed?
- Who needs to know? 
- If they need to do something, did we leave enough clues in changelog to get them stared?

Is `woocommerce_subscription_trial_ended` a replacement or new/better API for an existing thing – i.e. hooking `woocommerce_scheduled_subscription_trial_end` or `woocommerce_scheduled_subscription_payment_retry`?

PR #458, opened 2023-06-30
Fix PHP Lint issues currently causing automated checks to fail

prettyboymp:
## Description

Automated checks in this repo are failing because of existing lint issues.  This fixes those two issues by adding a phpcs:ignore for the specific problems.  In both cases, this is the correct fix.

`maybe_redirect_after_login()` already has a nonce check during the log-in process.

`validate_recurring_shipping_methods()` is comparing arrays which may not have the same key order.


mattallan (approved):
Thanks @prettyboymp, I've just added an inline comment above ignoring the nonce to explain why we can ignore it (for future reference).

If that's all good, this is good to merge :)

PR #468, opened 2023-07-11
When HPOS is disabled use the users subscription cache when fetching subscriptions by customer ID

### LINKED ISSUE #466: Use user's subscriptions cache data when querying customer IDs in wcs_get_subscription() on stores not using HPOS
## Describe the bug

In https://github.com/Automattic/woocommerce-subscriptions-core/pull/171 we removed this block of code inside our `wcs_get_subscriptions()` function as it was no longer necessary with HPOS enabled:

```
// Maybe filter to a specific user
if ( 0 != $args['customer_id'] && is_numeric( $args['customer_id'] ) ) {
	$users_subscription_ids = WCS_Customer_Store::instance()->get_users_subscription_ids( $args['customer_id'] );
	$query_args             = WCS_Admin_Post_Types::set_post__in_query_var( $query_args, $users_subscription_ids );
};
```

For large stores without HPOS this can be a significant hit to performance when querying for subscriptions by `customer_id` so it should still be used when HPOS is disabled.

#### Acceptance criteria

1. For stores with HPOS disabled, bring back the above approach of using the user's subscription cache data + `post__in` query args when querying by customer ID
2. For stores with HPOS enabled, continue to query using `customer_id` and remove the need for the users subscription cache.

mattallan:
Fixes #466 

## Description

In https://github.com/Automattic/woocommerce-subscriptions-core/pull/171 we removed a section of our `wcs_get_subscriptions()` function which used get subscriptions by customer ID using the user's subscription cache and then setting this list of IDs in `post__in` to improve performance of this query.

While working on HPOS changes we removed the cache/post__in approach as it was no longer needed due to improvements with the new order tables, however, this now means non-HPOS stores are now inefficiently getting subscriptions by customer ID.

The code we removed:
```
// Maybe filter to a specific user
if ( 0 != $args['customer_id'] && is_numeric( $args['customer_id'] ) ) {
	$users_subscription_ids = WCS_Customer_Store::instance()->get_users_subscription_ids( $args['customer_id'] );
	$query_args             = WCS_Admin_Post_Types::set_post__in_query_var( $query_args, $users_subscription_ids );
};
```

For large stores without HPOS, this can be a pretty significant hit to performance when using `wcs_get_subscriptions()` with the `customer_id` arg, so this PR brings back the user cache + `post__in` approach we had previously when HPOS is disabled.

## Product impact

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -2,6 +2,7 @@
 
 = 6.2.0 - 2023-xx-xx =
 * Fix - Ensure subscription checkout and cart block integrations are loaded on store environments where WooPayments is not enabled.
+* Update - When HPOS is disabled, fetch subscriptions by customer_id using the user's subscription cache to improve performance.
 
 = 6.1.0 - 2023-07-26 =
 * Fix - Resolved an issue that prevented the "Used for variations" checkbox to not be enabled on the edit product page load causing variations to be deleted erroneously.
--- a/wcs-functions.php
+++ b/wcs-functions.php
@@ -501,7 +501,13 @@ function wcs_get_subscriptions( $args ) {
 
 	// Maybe filter to a specific customer.
 	if ( 0 !== $args['customer_id'] && is_numeric( $args['customer_id'] ) ) {
-		$query_args['customer_id'] = $args['customer_id'];
+		// When HPOS is disabled, fetch subscriptions by customer_id using the user's subscription cache and query by post__in for improved performance.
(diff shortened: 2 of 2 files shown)


james-allan (approved):
Confirmed this returns the approach that was previously used prior to #171. 

I ran a basic test on a WP Posts and HPOS tables and confirmed the results returned are correct. I also note that there's a considerable number of unit tests that cover this function.

mattallan:
I've added a changelog entry. Thanks @james-allan for the review.

PR #474, opened 2023-07-17
[HPOS] Fix wcs_get_subscription_orders() returning empty list when querying parent orders

### LINKED ISSUE #422: End of the Month Subscription Issue for February 27, 2023 @ 11:00 pm
## Describe the bug

Have a site where we are using subscriptions to set up payment plans, and  while this isn't what it was inteded for it was the best option at the time.

We have a scenario that popped up where there is a variable subscription product with options of "Pay in Full" or Pay over 3 months"

The Pay in Full is a set to be the full amount paid every month but set to expire in 1 month

1. The customer originally purchased the **Pay in Full** on  **February 27, 2023 @ 11:00 pm EST**

1. ​That product was set correctly to pay $X.xx every month, but is also set to expire after 1 month.

1. ​​The renewal order happened on  **March 28, 2023 @ 12:05 am EDT** (daylight savings was on March 12, 2023)​​

1. ​The subscription is set to expire **March 30, 2023 @ 11:59 pm EDT** (the end of the month) ​

The issue appears to be with the WooCommerce Subscription plugin and how in this case it calculated when the subscription expires vs. schedules the renewal action
https://woocommerce.com/document/subscriptions/faq/#section-17

## To Reproduce

1. Create a subscription product that is $X.xx paid every month and expires in one month
4. Set you server/site time to be February 27, 2023 @ 11:00 pm when you purchase it
5. Review the Parent Order, Subscription, and Scheduled Action for the Renewal

### Expected behavior

Expected the subscription's expiration date and renewal to be the same causing the expiration to end and not renew

### Actual behavior

The subscription renewed 2 days prior to it being set to expire charging the customer again.

## Product impact

- [ yes] Does this issue affect WooCommerce Subscriptions? yes/no/tbc, add issue ref
- [ tbc] Does this issue affect WooCommerce Payments? yes/no/tbc, add issue ref

## Additional context
This seems like an edge case scenario, but the logic in setting the expiration date and scheduling the action for renewal seem to not be calculated the same.

mattallan:
Fixes #422

## Description

With HPOS enabled & data synchronization disabled, calling `wcs_get_subscription_orders( 'ids', 'parent' )` will return an empty array due to this function querying for `post_parent` from the posts table.

In this PR I'm replacing the following query with code that alters the query based on whether HPOS is enabled.

```
$order_ids = array_merge( $order_ids, $wpdb->get_col(
	"SELECT DISTINCT post_parent FROM {$wpdb->posts}
	 WHERE post_type = 'shop_subscription'
	 AND post_parent <> 0"
) );
```

> **Note** - This function not working as expected doesn't impact WC Subscriptions as we don't use this with HPOS enabled.

## How to test this PR

1. Enable HPOS on your store and turn off data syncing
2. Purchase a subscription
3. With WP Console run the following snippet:
```
echo var_export( wcs_get_subscription_orders( 'ids', 'parent' ), true );
```
4. While on `trunk`, this will return an empty array.
5. On this branch, the query will return all subscription parent orders.

## Product impact

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Subscriptions Core Changelog ***
 
+= 6.0.0 - 2023-xx-xx =
+* Dev - Fixed wcs_get_subscription_orders() returning an empty list when querying parent orders when HPOS is enabled with data syncing off.
+
 = 5.9.0 - 2023-07-14 =
 * Fix - Ensure when a customer changes the shipping method on cart and checkout that the recurring totals correctly reflect the chosen method.
 
--- a/includes/wcs-order-functions.php
+++ b/includes/wcs-order-functions.php
@@ -455,11 +455,18 @@ function wcs_get_subscription_orders( $return_fields = 'ids', $order_type = 'par
 	$order_ids = array();
 
 	if ( $any_order_type || in_array( 'parent', $order_type ) ) {
+		$is_hpos          = wcs_is_custom_order_tables_usage_enabled();
+		$table_name       = $is_hpos ? 'wc_orders' : 'posts';
+		$parent_order_col = $is_hpos ? 'parent_order_id' : 'post_parent';
+		$type_col         = $is_hpos ? 'type' : 'post_type';
+
+		// @codingStandardsIgnoreStart
 		$order_ids = array_merge( $order_ids, $wpdb->get_col(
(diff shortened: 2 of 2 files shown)


Mayisha (approved):
Tested the return value of `wcs_get_subscription_orders` function and the results are coming correctly in this branch :tada:

PR #520, opened 2023-10-05
Update trash_users_subscriptions() to properly delete subscriptions in HPOS environments.

mattallan:
## Description

When a user/subscriber is deleted, we hook onto the `delete_user` hook and delete all subscriptions that belong to this user. This function was still using `wp_post_delete()` which doesn't work in a HPOS environment resulting in subscriptions remaining without a customer attached.

This PR updates the `trash_users_subscriptions()` function and replaces the `wp_post_delete( $id )` method with `$subscription->delete( true )` so work in both HPOS & WP Posts environments.

## How to test this PR

1. Enable High-performance order storage on your store (found: WC > Settings > Advanced > Features)
2. Create a throw-away user by going to WP Admin > Users > Add New
3. Create the user with the minimum details i.e. customer1 customer1@example.com and default password
4. Go to WC > Subscriptions > Add New
5. Create a basic subscription for this new customer and use the actions drop-down to create a pending parent order.
6. Go to WP Admin > Users and delete the user.
7. On `trunk` you will notice the subscription still exists without a customer and the parent order will have a note saying the subscription was deleted.
8. On this branch the subscription will be removed.

## Product impact

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,11 +1,12 @@
 *** WooCommerce Subscriptions Core Changelog ***
 
 = 6.3.0 - xxxx-xx-xx =
+* Add - Introduce the "Subscription Relationship" column under the Orders list admin page when HPOS is enabled.
 * Fix - When HPOS is enabled, make the orders_by_type_query filter box work in the WooCommerce orders screen.
 * Fix - Ensure renewal orders paid via the Block Checkout are correctly linked to their subscription.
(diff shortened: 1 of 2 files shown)


james-allan (approved):
Changes look good to me and work as described.

I replicated the issue on `trunk` - the customer subscriptions remained after I deleted the user. 

I confirmed this also works while syncing is enabled (it deletes in both tables) with both HPOS and WP Posts tables.

- [x] HPOS - no syncing 
- [x] HPOS - with syncing
- [x] WP Posts - no syncing
- [x] WP Posts - with syncing

PR #539, opened 2023-11-08
After processing an early renewal, make sure the suspension count meta is reset to 0

mattallan:
## Description

In Woo Subscriptions we have a setting that limits the number of times a customer can suspend their subscription:
image

We store the number of times a subscription is suspended in meta `_suspension_count` and whenever we process a renewal order, we reset this count back to 0. While testing 4584-gh-woocommerce/woocommerce-subscriptions I noticed that early renewals do not reset the `_suspension_count` meta because even though inside `$subscription->payment_complete_for_order()` we're calling `$this->set_suspension_count( 0 );`, we never actually call `$subscription->save()`...

The issue is we relied on `$subscription->update_status( 'active' );` to call `save()` however this function only calls save if the status is actually updated and in the case of early renewals, the subscription status is not changed.

To fix this I've manually called save if we don't need to update the status of the subscription.

## How to test this PR

1. From the **WC > Settings > Subscriptions** page, enable early renewals and update the Customer Suspensions setting to any number greater than 1.
2. Purchase an active subscription, then from the My Account > Subscriptions > view subscription, click on the  "Suspend" action.
3. Open the DB and confirm `_suspension_count` meta is `1`.
4. Reactivate the subscription and click on "Renew now" action.
5. After renewing the subscription early, check in the DB to confirm the `_suspension_count` has been reset to 0.
6. On `trunk` this number will remain 1, but on this branch it'll be reset back to 0.

## Product impact

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -3,6 +3,7 @@
 = 6.5.0 - xxxx-xx-xx =
 * Add - When a customer toggles automatic renewals on or off via their My Account page, add a note to the subscription to record that event.
 * Fix - When a subscription is flagged as requiring manual payments, allow admin users to turn on automatic payments for a subscription via the Edit Subscription page by selecting a new payment method.
+* Fix - When processing an early renewal order, make sure the suspension count is reset back to 0 on payment complete.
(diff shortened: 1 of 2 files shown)


james-allan (approved):
Changes look good. The unit tests are failing, however they will be fixed once https://github.com/Automattic/woocommerce-subscriptions-core/pull/538 is merged.

PR #561, opened 2024-01-22
Added break statements in switch (#533)

### LINKED ISSUE #533: Deleting a subscription with validation error causes PHP error
## Describe the bug
I tried to delete (move to trash) a subscription, but because it had validation error, all I received as the critical error message.

The critical error message I received via email was:

```
An error of type E_ERROR was caused in line 2470 of the file /wp-content/plugins/woocommerce-subscriptions/vendor/woocommerce/subscriptions-core/includes/class-wc-subscription.php. 
Error message: Uncaught Exception: Subscription #58154: The cancelled date must occur after the last payment date. The end date must occur after the last payment date. in /wp-content/plugins/woocommerce-subscriptions/vendor/woocommerce/subscriptions-core/includes/class-wc-subscription.php:2470
Stack trace:
#0 /wp-content/plugins/woocommerce-subscriptions/vendor/woocommerce/subscriptions-core/includes/class-wc-subscription.php(1272): WC_Subscription->validate_date_updates()
#1 /wp-content/plugins/woocommerce-subscriptions/vendor/woocommerce/subscriptions-core/includes/class-wc-subscription.php(526): WC_Subscription->update_dates()
#2 /wp-content/plugins/woocommerce-subscriptions/vendor/woocommerce/subscriptions-core/includes/class-wc-subscriptions-manager.php(917): WC_Subscription->update_status()
...[remainder not sent]
```

## To Reproduce
I don't actually know how to get the subscription into the invalid state.

### Expected behavior
I would expect that you'd still be able to delete the subscription or at least receive an error message showing what's actually wrong.

### Actual behavior
Received the critical error message.

## Product impact
I believe this is only on woo subscriptions.

- [x] Does this issue affect WooCommerce Subscriptions? yes

10goC:
Fixes #533 

## Description

Missing `break` statements inside the `switch` ... `case` block in `WC_Subscription::validate_date_updates()` are causing an unexpected behavior when validating dates while changing subscription status.

According to PHP documentation, 

> when a case statement is found whose expression (...) matches the value of the switch expression (...) PHP continues to execute the statements until the end of the switch block.

So for example, when the case matches 'end' then all the following lines inside the block will be executed as if the expression would match 'cancelled', 'next_payment' and 'trial_end'.

- How can this code break?
This shouldn't break because it doesn't add any new code, it only enhances existing code, ensuring expected behavior.

subscription-bug-533

In this screenshot we can see how the debugger stops in the wrong case.

## How to test this PR

1. Generate a subscription.
2. After the subscription period, generate renewal.
3. Move the subscription status to "On hold".
4. Make a PUT request to /wc/v3/subscriptions/{ID} with {"status":"cancelled"}
5. Before the fix, the request will throw an error: `The next_payment date must occur after the start date.`
    After the fix, the request should change the subscription status to `cancelled` without errors.

## Product impact

- [x] Affects WooCommerce Subscriptions  #533

### DIFF
--- a/includes/class-wc-subscription.php
+++ b/includes/class-wc-subscription.php
@@ -2516,6 +2516,7 @@ class WC_Subscription extends WC_Order {
 						// translators: %s: date type (e.g. "end").
 						$messages[] = sprintf( __( 'The %s date must occur after the cancellation date.', 'woocommerce-subscriptions' ), $date_type );
 					}
+					break;
 
 				case 'cancelled':
 					if ( array_key_exists( 'last_order_date_created', $timestamps ) && $timestamp < $timestamps['last_order_date_created'] ) {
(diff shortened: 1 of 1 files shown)


lkraav:
So what happened here, the discovery was not correct?

PR #563, opened 2024-01-30
Verify order exists in WCS_Initial_Cart_Stock_Manager::get_order_from_query_vars() before attempting to use it.

prettyboymp:
Fixes #

## Description

Invalid order IDs passed to the order payment screen cause a fatal error because `WCS_Initial_Cart_Stock_Manager::has_handled_stock() ` attempts to use `$order` when the value of the parameter as passed in from `::get_order_from_query_vars()` is `false`

## How to test this PR

1. Attempt to go to a pay-for-order page for a non-existing order ID, e.g. `/checkout/order-pay/123456789/?pay_for_order=true&key=wc_order_123456789`

## Product impact


.phpunit.result.cache
wjrosa, line 1: Is this file intended to be committed? (probably not)

wjrosa (approved):
LGTM! Left a comment only

PR #573, opened 2024-02-22
Fix PHP 8.2 deprecated dynamic property errors in \WCS_PayPal_Reference_Transaction_API

lipemat:
PHP 8.2 required properties be defined before assigning them. The `\WCS_PayPal_Reference_Transaction_API` class has 4 properties assigned in __constructs which do not exist and return deprecated warnings.

## Description

Support using PHP 8.2 with the `\WCS_PayPal_Reference_Transaction_API` class without filling the logs with deprecated warnings or throwing exceptions during testing.

## How to test this PR

1. Set `error_reporting = E_ALL` in php.ini and tail the logs.
2. Use `convertDeprecationsToExceptions="true"` in <phpunit> config during unit testing.
3. Call any code which constructs the `\WCS_PayPal_Reference_Transaction_API` class. 

## Product impact

- [X] Will this PR affect WooCommerce Subscriptions? yes/no/tbc, add issue ref

### DIFF
--- a/includes/gateways/paypal/includes/class-wcs-paypal-reference-transaction-api-request.php
+++ b/includes/gateways/paypal/includes/class-wcs-paypal-reference-transaction-api-request.php
@@ -695,11 +695,11 @@ class WCS_PayPal_Reference_Transaction_API_Request {
 	 * Format prices.
 	 *
 	 * @since 1.0.0 - Migrated from WooCommerce Subscriptions v2.2.12
-	 * @param float|int $price
+	 * @param float|int|null $price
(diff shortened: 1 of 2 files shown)


mattallan (approved):
Thanks @lipemat for this PR and apologies for the delay in getting to this.

I reproduced the issue and confirmed the fix using the following steps:
1. Switch to PHP 8.2
2. Enable PayPal Standard using `add_action( 'woocommerce_should_load_paypal_standard', '__return_true' );`
3. Set up PayPal standard with a newer PayPal Account (with reference transactions enabled)
4. Purchase a subscription product

PR #574, opened 2024-02-24
Support block themes for add to cart button

bekarice:
Adds a CSS class needed for block themes to appropriately style buttons. 

Fixes #

## Description

Without the `wp-element-button` class applied, the add to cart buttons within the **Add to Cart with Options** block doesn't adopt any of the expected styles. Adding this class, when using add to cart blocks, resolves the styling issue. I tested this with the Twenty Twenty-Four theme.

## How to test this PR

- Use Twenty Twenty-Four
- Adjust your stylebook to set button styles / colors / etc
- Create the single-product template using blocks, ensuring that **Add to Cart with Options** is used
- Observe that the buttons are unstyled
- Apply this fix, and see they now adopt the theme styling

## Product impact

### DIFF
--- a/templates/single-product/add-to-cart/subscription.php
+++ b/templates/single-product/add-to-cart/subscription.php
@@ -48,7 +48,7 @@ if ( $product->is_in_stock() ) : ?>
 		do_action( 'woocommerce_after_add_to_cart_quantity' );
 		?>
 
-		<button type="submit" class="single_add_to_cart_button button alt" name="add-to-cart" value="<?php echo esc_attr( $product->get_id() ); ?>"><?php echo esc_html( $product->single_add_to_cart_text() ); ?></button>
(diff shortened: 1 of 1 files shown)


mattallan (approved):
Thanks @bekarice for this PR and also happy retirement! 😄

bekarice:
@mattallan  thanks! ♥️ didn't see any other instances of buttons that would need this fix (maybe the switching buttons?), but would be great to double check!

mattallan:
> didn't see any other instances of buttons that would need this fix (maybe the switching buttons?), but would be great to double check!

@bekarice yeah I had a look too and couldn't see any other `<button>` uses across subscriptions core and the main plugin so we should be good.

Looks like we use `<a href>` styled as buttons a lot

PR #586, opened 2024-03-26
Update the change payment method flow to reduce customers missing the checkbox to update all subscriptions

mattallan:
Fixes 4372-gh-woocommerce/woocommerce-subscriptions
Slack p1710775338910969-slack-C7U3Y3VMY

## Description

According to the linked issue, merchants are regularly contacting support confused as to why their customers are updating their payment details but their other subscriptions were not updating, and as reported in the issue, this is mostly due to the checkbox option being missed by customers:

image

This isn't a good merchant or customer experience so in the slack thread, I proposed the idea of simply changing this checkbox to be enabled by default, but that may have the opposite effect to those customers that wish to use different cards for different subscriptions and also don't see this checkbox.

In this PR I decided to experiment a little bit and propose moving the checkbox into a separate modal, prompting customers to take action before submitting the change payment method form.

image

I'm not convinced that the modal approach is the best thing here, mainly because it's introducing another step to the process. The other thing I'm not sure about is the text/content of the modal itself. There's not much we need to put here and it looks rather boring if you ask me, so I'm open to other options to help improve this.

Some other alternatives I considered:
 - Instead of a checkbox, use radio buttons and make it a required field so that customers have to choose. This could be a bit ugly 😅 
 - Have an option in subscription settings to let stores use the old checkbox over the modal approach if they don't like it.

## How to test this PR

1. Have a payment method that supports changing payment methods (i.e. Stripe)
2. Have multiple subscriptions purchased to prompt the update all my subscriptions flow
3. Visit **My Account > Subscriptions > View Subscription**
4. Click on Change payment action
5. Select a different payment method and press "Change payment method"
6. You should notice a new modal prompting whether you want to update all of your subscriptions.
7. Check the box and confirm all of your subscriptions have been updated.

## Product impact

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 *** WooCommerce Subscriptions Core Changelog ***
 
 = 6.9.0 - xxxx-xx-xx =
+* Add - New modal to replace the checkbox displayed on the change payment method page asking customers to update all subscriptions.
 * Fix - Resolved an issue where discounts, when reapplied to failed or manual subscription order payments, would incorrectly account for inclusive tax.
 * Fix - Resolved an issue that could cause an empty error notice to appear on the My Account > Payment methods page when a customer attempted to delete a token used by subscriptions.
 * Fix - Make sure we always clear the subscription object from cache after updating dates.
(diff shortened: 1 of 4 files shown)


mattallan:
Going to change this approach after some internal discussions and go with just setting the default to enabled. I'll open up a separate PR for this

PR #588, opened 2024-03-26
Make the checkbox on the change payment method page which updates all subscriptions enabled by default

mattallan:
Supersedes https://github.com/Automattic/woocommerce-subscriptions-core/pull/586
Fixes 4372-gh-woocommerce/woocommerce-subscriptions
Slack p1710775338910969-slack-C7U3Y3VMY

## Description

According to the linked issue, merchants are regularly contacting support confused as to why their customers are updating their payment details but their other subscriptions were not updating, and as reported in the issue, this is mostly due to the checkbox option being missed by customers:

image

This isn't a good merchant or customer experience so in the slack thread, I proposed the idea of simply changing this checkbox to be enabled by default which I've done in this PR:
image

Some other changes I've made:
- Make the field required. This is to remove the "(optional)" text which didn't make much sense if the field was enabled by default. But also puts a red * next to the field to make it a bit more noticeable by customers
- Change the text of the field slightly.

## How to test this PR

1. Have a payment method that supports changing payment methods (i.e. Stripe)
2. Have multiple subscriptions purchased to prompt the update all my subscriptions flow
3. Visit **My Account > Subscriptions > View Subscription**
4. Click on Change payment action
5. Notice the field is selected by default on page load

## Product impact


james-allan (approved):
Nice little improvement here. Changes look good to me. 

I've tested both checked and unchecked variants and confirmed that all subscription payment methods are updated or not based on the setting on this checkout box.

PR #607, opened 2024-04-17
Fix array related fatal error

### LINKED ISSUE #568: PHP Fatal error in cron when $subscription_ids is an empty array.
I ran into an issue when updating an old subscription where the $subscription_ids variable turned out to contain an empty array. Which triggers a PHP Fatal error that prevents wp cron from finishing the rest of the queue of subscriptions.

```PHP Fatal error:  Uncaught TypeError: rsort(): Argument #1 ($array) must be of type array, bool given in /mnt/persist/www/arbetaren.se/release/web/app/plugins/woocommerce-subscriptions/vendor/woocommerce/subscriptions-core/includes/data-stores/class-wcs-customer-store-cached-cpt.php:117```

Perhaps error handling with empty would be in order here to ensure wp cron doesn't abort?

```
if ( ! empty( $subscription_ids ) ) {
	rsort( $subscription_ids );
}
```

https://github.com/Automattic/woocommerce-subscriptions-core/blob/19d4f2acf246c7f8275438c9356eb79a141bdf4d/includes/data-stores/class-wcs-customer-store-cached-cpt.php#L117

### PR DESCRIPTION
**Pull request by Mayisha**
Fixes #568 

## Description
Added a couple of safety checks to prevent array related fatal errors.

### DIFF
--- a/includes/data-stores/class-wcs-customer-store-cached-cpt.php
+++ b/includes/data-stores/class-wcs-customer-store-cached-cpt.php
@@ -114,7 +114,9 @@ class WCS_Customer_Store_Cached_CPT extends WCS_Customer_Store_CPT implements WC
 		}
 
 		// Sort results in order to keep consistency between cached results and queried results.
-		rsort( $subscription_ids );
+		if ( ! empty( $subscription_ids ) ) {
+			rsort( $subscription_ids );
+		}
 
 		return $subscription_ids;
 	}
--- a/includes/data-stores/class-wcs-customer-store-cached-cpt.php
+++ b/includes/data-stores/class-wcs-customer-store-cached-cpt.php
@@ -113,10 +113,12 @@ class WCS_Customer_Store_Cached_CPT extends WCS_Customer_Store_CPT implements WC
 			$this->update_subscription_id_cache( $user_id, $subscription_ids );
 		}
 
-		// Sort results in order to keep consistency between cached results and queried results.
-		if ( ! empty( $subscription_ids ) ) {
-			rsort( $subscription_ids );
+		if ( ! is_array( $subscription_ids ) ) {
+			return [];
 		}
+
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: includes/data-stores/class-wcs-customer-store-cached-cpt.php [hunk 1]
```diff
@@ -114,7 +114,9 @@ public function get_users_subscription_ids( $user_id ) {
 		}
 
 		// Sort results in order to keep consistency between cached results and queried results.
-		rsort( $subscription_ids );
+		if ( ! empty( $subscription_ids ) ) {
```
mattallan:
```suggestion
		if ( is_array( $subscription_ids ) ) {
```

Thoughts on changing this to is_array()?

Or maybe return an empty array if it's not in the expected array format:

```
if ( ! is_array( $subscription_ids ) ) {
    return [];
}

// Sort results in order to keep consistency between cached results and queried results.
rsort( $subscription_ids );

return $subscription_ids;
```

Benefits to this is we're also not going to try call rsort on strings and other non-array types.
We're also validating the output of this function but returning an empty list if `$subscription_ids` is something other than array

mattallan (approved):
Thanks @Mayisha

I've just made a small suggestion but the rest of this looks good :)

It's difficult to understand what could by causing products that is taxable to not have a subtotal and for querying users orders to return a boolean but adding these additional safety checks prevents the errors so I'm happy to approve them.

Thanks!

includes/data-stores/class-wcs-customer-store-cached-cpt.php
Mayisha: Thanks @mattallan. I am going for the 2nd suggestion and returning an empty array. Addressed in [dd5c692](https://github.com/Automattic/woocommerce-subscriptions-core/pull/607/commits/dd5c6929cd368febc3b026a1465a937bbc11d9f9)

PR #609, opened 2024-04-26
Fix filtered order links in subscription reports not working correctly when HPOS is enabled

Mayisha:
Fixes WCS-4650

With HPOS enabled, the links highlighted in the screenshot below do not work as intended, resulting in a list of all orders instead of showing the filtered order for the reports being viewed.

reports

### Description
- The existing `posts_where` filter works on the `posts` table only. As this is not triggered for the HPOS order table, the orders are not filtered by expected ids when coming from the reports page.
- I have used the `woocommerce_orders_table_query_clauses` filter which is available to modify the query clauses of the HPOS order tables.

## How to test this PR

1. Enable HPOS from **WooCommerce > Settings > Advanced > Features** page.
2. Enable compatibility mode if not enabled already to view the subscription report when HPOS is enabled.
3. Go to **WooCommerce > Reporst > Subscriptions**
4. Click the link `subscription renewals`.
5. In `trunk` it should take you to the order table page and display all the orders.
6. In this branch, it should take you to the order table page and display the renewal orders only.

### DIFF
--- a/includes/admin/class-wc-subscriptions-admin.php
+++ b/includes/admin/class-wc-subscriptions-admin.php
@@ -1402,7 +1402,7 @@ class WC_Subscriptions_Admin {
 	public static function filter_orders_and_subscriptions_from_order_table( $clauses ) {
 		global $wpdb;
 
-		if ( ! ( isset( $_GET['page'] ) && 'wc-orders' === $_GET['page'] ) || ! isset( $_GET['_report'] ) ) {
+		if ( ! ( isset( $_GET['page'] ) && in_array( $_GET['page'], [ 'wc-orders', 'wc-orders--shop_subscription' ] ) ) || ! isset( $_GET['_report'] ) ) {
 			return $clauses;
 		}
 
--- a/includes/admin/class-wc-subscriptions-admin.php
+++ b/includes/admin/class-wc-subscriptions-admin.php
@@ -1402,21 +1402,38 @@ class WC_Subscriptions_Admin {
 	public static function filter_orders_and_subscriptions_from_order_table( $clauses ) {
 		global $wpdb;
 
-		if ( ! ( isset( $_GET['page'] ) && in_array( $_GET['page'], [ 'wc-orders', 'wc-orders--shop_subscription' ] ) ) || ! isset( $_GET['_report'] ) ) {
+		$query_vars = [
+			'page'                    => '',
+			'_report'                 => '',
+			'_orders_list_key'        => '',
+			'_subscriptions_list_key' => '',
+			'_data_key'               => '',
+		];
+
+		// phpcs:disable WordPress.Security.NonceVerification.Recommended
+		// Map and sanitize $_GET vars to $query_vars.
+		foreach ( array_keys( $query_vars ) as $var ) {
+			if ( isset( $_GET[ $var ] ) ) {
+				$query_vars[ $var ] = sanitize_text_field( wp_unslash( $_GET[ $var ] ) );
+			}
+		}
+		// phpcs:enable WordPress.Security.NonceVerification.Recommended
+
+		if ( ! ( ! empty( $query_vars['page'] ) && in_array( $query_vars['page'], [ 'wc-orders', 'wc-orders--shop_subscription' ], true ) ) || empty( $query_vars['_report'] ) ) {
 			return $clauses;
 		}
 
 		// Map the order or subscription type to their respective keys and type key.
-		$object_type = isset( $_GET['_orders_list_key'] ) ? 'order' : ( isset( $_GET['_subscriptions_list_key'] ) ? 'subscription' : '');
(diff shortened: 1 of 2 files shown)


includes/admin/class-wc-subscriptions-admin.php
```diff
@@ -1387,6 +1390,68 @@ public static function filter_orders_table_by_related_orders( $query_vars ) {
...
+	 * @since 7.1.0
+	 *
+	 * @param array $clauses The query clause.
+	 *
+	 * @return array $clauses The query clause with additional `where` clause .
+	 */
+	public static function filter_orders_and_subscriptions_from_order_table( $clauses ) {
+		global $wpdb;
+
+		if ( ! ( isset( $_GET['page'] ) && 'wc-orders' === $_GET['page'] ) || ! isset( $_GET['_report'] ) ) {
```
james-allan:
```suggestion
		if ( ! ( isset( $_GET['page'] ) && in_array( $_GET['page'], [ 'wc-orders', 'wc-orders--shop_subscription' ] ) ) || ! isset( $_GET['_report'] ) ) {
```

The way this logic is written at the moment this function won't work on the subscriptions list table page. 

If you click on any of these subscription specific ones, you'll be redirected to the subscriptions list table with no filtering applied. The suggestion I included above should fix that. 

Screenshot 2024-04-29 at 12 06 34 PM

includes/admin/class-wc-subscriptions-admin.php
```diff
@@ -1387,6 +1390,68 @@ public static function filter_orders_table_by_related_orders( $query_vars ) {
...
+
+		// The current subscriptions count report will include the specific result (the subscriptions active on the last day) that should be used to generate the subscription list.
+		if ( ! empty( $_GET['_data_key'] ) && isset( $results[ (int) $_GET['_data_key'] ] ) ) {
+			$results = array( $results[ (int) $_GET['_data_key'] ] );
+		}
+
+		$ids = explode( ',', implode( ',', wp_list_pluck( $results, "{$object_type}_ids", true ) ) );
+
+		$format = implode( ', ', array_fill( 0, count( $ids ), '%d' ) );
+		$clauses['where'] .= $wpdb->prepare( " AND {$wpdb->prefix}wc_orders.ID IN ($format)", $ids );
```
james-allan, line 1468: This line is throwing a couple of PHPCS errors because the prepare call has no placeholders but it does if you resolve the `$format` variable. 

WC core has similar code in a couple of places. For example the `sync_order_products()` function ([link](https://github.com/woocommerce/woocommerce/blob/1e225f701f2d21380b6e1a28b6a7cf88ad10b230/plugins/woocommerce/src/Admin/API/Reports/Products/DataStore.php#L490-L491)). ie you have to include to phpcs ignore.

james-allan (changes requested):
Hey @Mayisha I noticed there are a few PHPCS issues being flagged with this PR: 

Screenshot 2024-04-29 at 11 39 25 AM

I know this function you've introduced is based on the existing `filter_orders_and_subscriptions_from_list()` and I suspect that function has the same PHPCS concerns.

A couple of things about those PHPCS issues. 

**The Nonce violations** 

I'm not sure we need a nonce in this situation given the GET args in this situation. There must be a transient that matches the `_report` arg, and that transient value must have a key matching the value in `_subscriptions_list_key`. 

While that doesn't seem impossible, even if you got over that hurdle, the values inside that transient keyed value would all need to be integers otherwise they would be escaped by the `wpdb` `prepare`.

I think there might be better ways to protect this though. Ie double check that the `_report` arg is a known report type or something. 

We should also be cleaning these `$_GET` args etc.

james-allan (approved):

james-allan:
Changes look good. Thanks for picking this PR up @mattallan!

WebMarketingMatters:
Hi @mattallan @Mayisha @james-allan - Can I please confirm what version of WooCommerce subscriptions this fix is in? Looks to me like it's 7.1.0. Do you know when this will become available?
I found this bug and reported it to WooCommerce a few months ago, it would be good to finally have a fix!

james-allan:
Yep it should be 7.1.0 of `subscriptions-core`. That was included in 6.3.0 of the WooCommerce Subscriptions plugin which released back in early May.

WebMarketingMatters:
Hi @james-allan. Okay, I'm currently using 6.4.0 and the reports filtering doesn't work. It just brings up a question mark when you hover over the left-hand columns. (They should take you to the filtered/related orders for that period)

james-allan:
@WebMarketingMatters your screenshot probably didn't finish uploading before you submitted your comment so I cannot see it, however, I just had a look on my local site and it's working for me. 

Are you hovering over the link indicated by the ↗︎? Like this: 
Screenshot 2024-06-18 at 4 45 33 PM

WebMarketingMatters:
@james-allan Thanks for the reply. I have just cleared my cache and tried again and it's now working. Thanks for your help with this.

PR #614, opened 2024-05-09
Fixes bulk actions buttons (suspend, cancel) on Subscriptions list table not working while filtering the table

### LINKED ISSUE #553: Action buttons such as Cancel on Subscriptions page no longer work
## Describe the bug

Sometime in the last couple of months, hover Actions on the Subscriptions page no longer work, they return a 302 response page not found.

## To Reproduce

1. Go to /wp-admin/admin.php?page=wc-orders--shop_subscription&paged=1
2. Click on the Cancel hover action that appears when you hover near the Active indicator in the status column
3. The page will refresh but the Subscription cancel action does not occur
4. In Chrome console, you can see two requests, the first is a 302 reponse, the second is a 200 response:
https://www.example.com/wp-admin/admin.php?page=wc-orders--shop_subscription&paged=1&order%5B0%5D=63398&_wpnonce=xyz123&action=cancelled

https://www.example.com/wp-admin/admin.php?page=wc-orders--shop_subscription&paged=1

Note in the above, 63398 is an actual order ID, but I don't know why this is used: order%5B0%5D=63398
as I think this gets decoded as order[0]=63398  
IS THAT THE CORRECT FORMAT?

Otherwise, not sure why this is causing a 302 error.  Note that it works fine when I cancel from within the order, just not using the Cancel hover action.  

I have HPOS enabled the last couple of months, in case this is part of what triggered the issue.
All else is working fine with HPOS and I am no longer syncing orders.

### Expected behavior
The order should go into Pending Cancel status when you click the Cancel hover action

### Actual behavior
See 3. and 4. above

## Product impact
WooCommerce Subscriptions

- [yes ] Does this issue affect WooCommerce Subscriptions? yes/no/tbc, add issue ref

mattallan:
Fixes #553

## Description

When the subscriptions list table is being filtered by date, product, payment method or customer, the bulk action buttons under each list item (suspend, cancel etc) do not work:
image

This issue is caused by the `WP_List_Table::current_action()` function, which is called in WooCommerce's [`handle_bulk_actions()`](https://github.com/woocommerce/woocommerce/blob/f2cf6b56aa762cb36eb9496b878c5c295376f16d/plugins/woocommerce/src/Internal/Admin/Orders/ListTable.php#L1231-L1235), returning `false` when the `filter_action` query arg exists.

This means whenever the table is filtered, `handle_bulk_actions()` is not ran.
To fix this I've just removed the `filter_action` from the action URLs.

## How to test this PR

1. Go to the WooCommerce > Subscriptions and filter the table by customer "admin"
2. While on `trunk` try to use the bulk action buttons to cancel a subscription
3. Notice the page refreshes but the subscription is not cancelled :x: 
4. Check out this branch and run the same test.

## Product impact

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -11,6 +11,7 @@
 * Fix - Ensure the scheduled sale price for subscription products ends at the end of the "to" day set in product settings.
 * Fix - Subscription table is empty in mobile view when HPOS is enabled.
 * Fix - WooCommerce page header is hidden when HPOS is enabled.
+* Fix - Resolved an issue that prevented bulk actions from running on the Subscriptions list table when the table was filtered by date, payment method, product or customer.
 * Dev - Calling wcs_create_subscription() will no longer attempt to fetch a fresh instance of the subscription at the end. This is to prevent loading the subscription from the database potentially unnecessarily.
 
 = 7.0.0 - 2024-04-11 =
--- a/includes/admin/class-wcs-admin-post-types.php
+++ b/includes/admin/class-wcs-admin-post-types.php
@@ -1367,7 +1367,7 @@ class WCS_Admin_Post_Types {
 		}
 
 		$action_url   = add_query_arg( $action_url_args );
-		$action_url   = remove_query_arg( [ 'changed', 'ids' ], $action_url );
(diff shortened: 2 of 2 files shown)


james-allan (approved):
Changes look good. I've verified the issue on `trunk` and it's fixed on this PR.

PR #623, opened 2024-05-16
[WC Blocks] Show "FREE" instead of $0.00 in order summary if shipping is free

opr:
> [!NOTE]  
> This PR is a follow up to https://github.com/woocommerce/woocommerce/pull/47553 please wait for that to be merged (or check that PR out) before testing this.

## Description

This PR will show the text 'FREE' in bold and capital letters in the recurring totals area of the order summary on the Cart and Checkout blocks.

| Before | After |
| ------ | ----- |
| image | image |

## How to test this PR

1. Have at least two shipping options set up, one should be free shipping.
2. Go to WooCommerce -> Settings -> Shipping -> Shipping Settings and ensure `Enable the shipping calculator on the cart page` is checked.
3. Add a subscription item to your cart and go to the Cart block.
4. In the order summary, find the recurring totals section and the recurring packages section.
5. In the recurring packages section select the free shipping rate. The shipping price in the recurring totals area should now say 'FREE'.
<image src="https://github.com/Automattic/woocommerce-subscriptions-core/assets/5656702/b2d85423-9315-4135-8d6a-fc80e88b981b" width="400" />

6. Change the recurring package's shipping method to flat rate and ensure the correct price shows in the recurring totals section. It should not say 'FREE'.
7. Go to the Checkout block and do the same, ensure the correct value shows for free and paid shipping.

## Product impact

- [x] Will this PR affect WooCommerce Subscriptions? yes~/no/tbc~, add issue ref
- [ ] Will this PR affect WooCommerce Payments? ~yes~/no/~tbc~, add issue ref

### DIFF
--- a/assets/src/js/recurring-totals/index.js
+++ b/assets/src/js/recurring-totals/index.js
@@ -93,7 +93,11 @@ const ShippingTotal = ( {
 		: parseInt( values.total_shipping, 10 );
 
 	const valueToShow =
-		0 === shippingTotals ? <strong>Free</strong> : shippingTotals;
+		0 === shippingTotals ? (
+			<strong>{ __( 'Free', 'woocommerce-subscriptions' ) }</strong>
+		) : (
+			shippingTotals
+		);
 	return (
 		<TotalsItem
 			value={ valueToShow }
--- a/assets/src/js/recurring-totals/index.js
+++ b/assets/src/js/recurring-totals/index.js
@@ -91,9 +91,12 @@ const ShippingTotal = ( {
 		? parseInt( values.total_shipping, 10 ) +
 		  parseInt( values.total_shipping_tax, 10 )
 		: parseInt( values.total_shipping, 10 );
+
+	const valueToShow =
+		0 === shippingTotals ? <strong>Free</strong> : shippingTotals;
 	return (
 		<TotalsItem
-			value={ shippingTotals }
+			value={ valueToShow }
 			label={ __( 'Shipping', 'woocommerce-subscriptions' ) }
 			currency={ currency }
 			description={
@@ -257,7 +260,7 @@ const RecurringSubscription = ( {
 					<Subtotal currency={ currency } values={ totals } />
 					<DiscountTotals currency={ currency } values={ totals } />
 				</TotalsWrapper>
-				<TotalsWrapper>
+				<TotalsWrapper className="wc-block-components-totals-shipping">
 					<ShippingTotal
 						currency={ currency }
 						needsShipping={ needsShipping }
--- a/assets/src/js/recurring-totals/index.js
+++ b/assets/src/js/recurring-totals/index.js
@@ -10,7 +10,7 @@ import {
 	TotalsWrapper,
 } from '@woocommerce/blocks-checkout';
 import { getCurrencyFromPriceResponse } from '@woocommerce/price-format';
-import { getSetting } from '@woocommerce/settings';
+import { isWcVersion, getSetting } from '@woocommerce/settings';
 /**
  * Internal dependencies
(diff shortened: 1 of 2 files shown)


assets/src/js/recurring-totals/index.js
```diff
@@ -91,9 +91,12 @@ const ShippingTotal = ( {
 		? parseInt( values.total_shipping, 10 ) +
 		  parseInt( values.total_shipping_tax, 10 )
 		: parseInt( values.total_shipping, 10 );
+
+	const valueToShow =
+		0 === shippingTotals ? <strong>Free</strong> : shippingTotals;
```
opr: The styling to capitalise this is handled by CSS in WC Blocks, which is why an additional class was added to the `TotalsWrapper` below.

mattallan (approved):
Thanks @opr for the changes, I just pushed up a small commit to make sure the string is translatable but apart from that this is good to be merged :)

I'll wait for the PR to be merged into WooCommerce core first.

I had a question regarding styling... for stores running older versions of WooCommerce, do you think it's okay to just show the label as "Free" instead of "FREE" or should we include the CSS in subscriptions as well instead of relying on Woo?

Thanks!

opr:
@mattallan thanks a lot for your review and suggestions.

> I just pushed up a small commit to make sure the string is translatable

🤦🏼 Yeah that's my bad, thanks for catching that.

> should we include the CSS in subscriptions as well instead of relying on Woo

Great idea, actually, ~I'll take care of that.~ On second thought, how do you feel about the fact that in versions <9.0 "$0.00" will be shown in the main order summary, but FREE would be shown in WC Subscriptions? Seems like a lot of work to solve for something so minor (We'd need to port [this version checking function](https://github.com/woocommerce/woocommerce/blob/427fe33af303c0fd98f919c1d8a2af67cbc07b94/plugins/woocommerce-blocks/assets/js/settings/shared/utils.ts#L70) over to WC Subscriptions, unless similar exists already?) but would love to hear your thoughts on ensuring the consistency between core and WC Subscriptions.

mattallan:
> how do you feel about the fact that in versions <9.0 "$0.00" will be shown in the main order summary, but FREE would be shown in WC Subscriptions?

I feel like someone will notice the inconsistency so I;v just pushed up https://github.com/Automattic/woocommerce-subscriptions-core/pull/623/commits/9124f653af5f9eb666c53756f27b4aa03ec4752b which uses `isWcVersion()` to only show the "free" label when the store has 9.0+

@opr let me know if there will be problems with this.
Subscriptions requires a minimum of WC 7.9 and I've confirmed there's no issues on that version.

opr:
Thanks @mattallan this is great! I appreciate you using `isWcVersion`, I didn't realise that was exported!

I think this is ready to go then. Should we hold off on merging this until https://github.com/woocommerce/woocommerce/pull/47553 is merged? Just in case that doesn't make it into the 9.0 release for any reason?

changelog.txt
```diff
@@ -1,5 +1,8 @@
 *** WooCommerce Subscriptions Core Changelog ***
 
+= 7.1.2 - 2024-xx-xx =
+* Update - Show "FREE" instead of 0 when there is no shipping cost in the recurring totals section of the Cart and Checkout blocks.
```
```suggestion
* Update - Show "FREE" instead of 0 when there is no shipping cost in the recurring totals section of the Cart and Checkout blocks (requires WooCommerce 9.0+).
```

mattallan:
Great, thanks @opr! I see your PR into WooCommerce core was merged and tagged for the 9.0 milestone so we'll get this merged here as well.

PS. I added a small clarification in the changelog entry to include the WC version.

PR #644, opened 2024-07-31
Checking for admin context before calling the page screen ID check function

Mayisha:
Fixes 4682-gh-woocommerce/woocommerce-subscriptions
Ticket 8267031-zen

## Description

Some plugin partners reported that subscriptions-core is conflicting with their code. This is happening on more recent versions that call `wcs_get_page_screen_id`. This method calls `wc_get_page_screen_id` without checking for the required admin context. We are including that check here.

## How to test this PR

Code review should be enough, as the admin check function is always present.

## Product impact


wjrosa (approved):
Code change looks good and safe. I have also added an additional check to it.

Issue #645, opened 2024-07-31
[EPIC] Introduce Expiration and Renewal Notifications


xristos3490:
### Description

This epic is created to track the implementation of expiration and renewal notifications feature.

PRD: pfwcF5-Gn-p2
Project P2: pfwcF5-JS-p2

### Task List

- Create a scheduler class based on the `WCS_Scheduler` interface to manage hooks triggered by subscription changes.
- Develop a singleton controller to manage emails and create all necessary HTML and plain email templates.
- Copy the core `Automattic\WooCommerce\Internal\BatchProcessing\BatchProcessingController` into the plugin.
- Create a batch processor class using the batch processing interface to manage changes to the offset option (days prior to an event).
- Develop another batch processor based on the same interface to support a new debug tool found under WooCommerce > Status > Tools, which will fix all AS actions if necessary.
- Implement tracking data to monitor the usage of the new option.
- Create a dismissible admin notice to inform existing stores about the new feature.
- Ensure that everything is protected by unit tests.

#### Main Pull Request
- [x] https://github.com/Automattic/woocommerce-subscriptions-core/pull/648 (Main PR)

#### Improvements
- [x] https://github.com/Automattic/woocommerce-subscriptions-core/pull/706
- [x] https://github.com/Automattic/woocommerce-subscriptions-core/pull/694
- [x] https://github.com/Automattic/woocommerce-subscriptions-core/pull/703
- [x] https://github.com/Automattic/woocommerce-subscriptions-core/pull/677
- [x] https://github.com/Automattic/woocommerce-subscriptions-core/pull/708
- [x] https://github.com/Automattic/woocommerce-subscriptions-core/pull/713

#### Batch Processing
- [x] https://github.com/Automattic/woocommerce-subscriptions-core/pull/654
- [x] https://github.com/Automattic/woocommerce-subscriptions-core/pull/673
- [x] https://github.com/Automattic/woocommerce-subscriptions-core/pull/689
- [x] https://github.com/Automattic/woocommerce-subscriptions-core/pull/696
- [x] https://github.com/Automattic/woocommerce-subscriptions-core/pull/704

#### Testing
- [x] https://github.com/Automattic/woocommerce-subscriptions-core/pull/684

#### Tracking
- [x] https://github.com/Automattic/woocommerce-subscriptions-core/pull/695

PR #648, opened 2024-08-05
Added notification emails and basic scheduling logic

peterfabian:
First POC that contributes towards https://github.com/Automattic/woocommerce-subscriptions-core/issues/645

## Description

This is a WIP to add notifications to Subscriptions Core. 

Currently, it 
- adds simple email templates (placeholders until we [figure out the specifics](https://woomarketingstudio.wordpress.com/2024/07/10/studio-request-notify-users-about-new-minimum-required-php-woocommerce-and-wordpress-versions-for-all-somewherewarm-extensions/)), 
- adds email-related classes, 
- registers these classes, 
- adds simple scheduling logic for actions that should trigger the email with an offset (3 days)

Mostly, it's inspired by code that's already in Subscriptions Core. 

### Notification Types Added

- Auto Renewal
  - scheduled {{time_offset}} ahead of `$subscription->get_date( 'next_payment' )` for automatically renewing subscriptions (`$subscription->is_manual()` is `false`)
- Manual Renewal
  - scheduled {{time_offset}} ahead of `$subscription->get_date( 'next_payment' )` manual subscriptions (`$subscription->is_manual()` is `true`)
- Trial Expiry
  - scheduled {{time_offset}} ahead of `$subscription->get_date( 'trial_end' )` manual  subscriptions
- Subscription End
  - scheduled {{time_offset}} ahead of `$subscription->get_date( 'end' )` manual  subscriptions

### Open Questions

Some questions that crossed my mind while implementing this:
- We need to decide where to put filters/actions to allow extensibility
- I could have added the schedules to `WCS_Action_Scheduler`, but it seemed more appropriate to create a new class for notifications (`WCS_Action_Scheduler_Customer_Notifications `), otherwise the scheduling logic might be a bit complex. It should be quite easy to change this, though.
- The scheduling logic currently uses the state of the subscription object instead of deciding depending on which date/status is being updated. It was easier to grasp for me, given my inexperience with Subscriptions, but again, it should be possible to convert this to an approach similar to `WCS_Action_Scheduler` that relies entirely on Subscription status and which date is being updated (combining the logic of `get_scheduled_action_hook`, `update_status` and `update_date` methods).
- Which conditions we want to check at schedule time vs at the time of sending the notification?
- Do we want to unschedule notifications when store manager disables them? 
  - ~Most likely not, as it's a lot of work to schedule/unschedule all of them. If they're disabled, this will create clutter in the db, though~. Yes, we decided we will do this.

### Side Notes For Testing
- I had to downgrade to node 16 to build it
- I had problems committing as phpcs was giving me errors. I updated `ControlStructureSpacingSniff.php` (converted last argument to array for `ExtraSpaceBeforeCloseParenthesis` and `ExtraSpaceAfterCloseParenthesis` sniffs), otherwise `phpcs` was failing. Perhaps there's a better way to solve this, but I didn't want to spend too much time investigating what's going on and what array should be passed where it gets a string. Updating `squizlabs/php_codesniffer` to 3.10.2 didn't help either. 
  - it seems like [this has been fixed in the latest WP sniffs](https://github.com/WordPress/WordPress-Coding-Standards/blob/develop/WordPress/Sniffs/WhiteSpace/ControlStructureSpacingSniff.php#L169), so maybe updating some composer packages would resolve this (like the old WC sniffs?)

## How to test this PR

1. Create a couple of different types of subscriptions: simple, variable, with free trial, with defined end-date/length, with free trial and length
2. Buy these subscriptions, update them, renew them manually and automatically, cancel them. 
3. See if the notifications are being (un)scheduled as expected.


includes/class-wcs-core-autoloader.php
```diff
@@ -227,7 +227,7 @@ protected function get_relative_class_path( $class ) {
 			if ( false !== strpos( $class, 'handler' ) ) {
 				$path .= '/deprecation-handlers';
 			}
-		} elseif ( false !== strpos( $class, 'email' ) && 'wc_subscriptions_email' !== $class ) {
+		} elseif ( false !== strpos( $class, 'wcs_email' ) && 'wc_subscriptions_email' !== $class ) {
```
xristos3490: This seems somewhat risky due to potential backward compatibility issues 🤔

includes/class-wcs-action-scheduler-customer-notifications.php
```diff
@@ -0,0 +1,233 @@
...
+ * @class     WCS_Action_Scheduler
+ * @version   1.0.0 - Migrated from WooCommerce Subscriptions v2.0.0
+ * @package   WooCommerce Subscriptions/Classes
+ * @category  Class
+ */
+class WCS_Action_Scheduler_Customer_Notifications extends WCS_Scheduler {
+
+	protected int $hours_offset;
+
+	protected DateInterval $time_delta;
```
xristos3490: We probably don't want this. 

Also, I would consider refactoring the 'hours_offset' into a getter to encapsulate the 3PD filter to control offset based on a subscription object. Thoughts?

includes/class-wcs-action-scheduler-customer-notifications.php
```diff
@@ -0,0 +1,233 @@
...
+		$this->schedule_notification(
+			$subscription,
+			'woocommerce_scheduled_subscription_customer_notification_expiration',
+			$timestamp
+		);
+	}
+
+	public function schedule_payment_notification( $subscription ) {
+
+		//TODO: For end of trial, should we schedule payment notification? Let's say no.
```
xristos3490: +1

includes/class-wcs-action-scheduler-customer-notifications.php
```diff
@@ -0,0 +1,233 @@
...
+		$this->time_delta = new DateInterval( "PT{$this->hours_offset}H" );
+
+	}
+
+	public function set_date_types_to_schedule() {
+		$date_types_to_schedule = wcs_get_subscription_date_types();
+		unset( $date_types_to_schedule['start'] );
+
+		//TODO: filter?
+		$this->date_types_to_schedule = array_keys( $date_types_to_schedule );
```
xristos3490: Sounds good to me! An internal filter would be helpful for quickly managing active notifications "types"

If we do add one, we need to reuse it within each email's trigger function to handle existing records before applying that filter. It's a small "gap" but has a simple solution.

includes/class-wcs-action-scheduler-customer-notifications.php
```diff
@@ -0,0 +1,233 @@
...
+ * @category  Class
+ */
+class WCS_Action_Scheduler_Customer_Notifications extends WCS_Scheduler {
+
+	protected int $hours_offset;
+
+	protected DateInterval $time_delta;
+
+	protected array $notification_actions = array(
+		'woocommerce_scheduled_subscription_customer_notification_trial_expiring',
```
xristos3490: Nit: Perhaps keeping it consistent with the one below:
`woocommerce_scheduled_subscription_customer_notification_trial_expiration`?

includes/class-wcs-action-scheduler-customer-notifications.php
```diff
@@ -0,0 +1,233 @@
...
+	 * @param string $date_type Can be 'trial_end', 'next_payment', 'expiration', 'end_of_prepaid_term' or a custom date type
+	 * @param object $subscription An instance of WC_Subscription to get the hook for
+	 *
+	 * @return array Array of name => value pairs stored against the scheduled action.
+	 */
+	protected function get_action_args( $subscription ) {
+
+		$action_args = array( 'subscription_id' => $subscription->get_id() );
+
+		// TODO maybe filter
```
xristos3490: I think it might be a good idea to _avoid_ adding a filter for i1 here. 

Regardless of the decision, we should consider making `$date_type` the first argument to prevent potential future conflicts when adding a filter or implementing something else type-related in this context. WDYT?

includes/class-wcs-action-scheduler-customer-notifications.php
```diff
@@ -0,0 +1,233 @@
...
+				}
+				break;
+			case 'on-hold':
+			case 'cancelled':
+			case 'switched':
+			case 'expired':
+			case 'trash':
+				// Unschedule all
+				foreach ( $this->notification_actions as $action ) {
+					$this->unschedule_actions( $action, $this->get_action_args( $subscription ) );
```
xristos3490: If we introduce a filter in the `get_action_args()`, it may cause issues when trying to unschedule events registered without a potential hook. This is another reason to not add a filter there, in my opinion.

includes/class-wcs-action-scheduler-customer-notifications.php
xristos3490: This might also concern the global toggle of this feature. Since it is initially disabled, we need a reliable method to control its access. Adding a filter here could introduce some complexity, but there is a crucial difference between this class and the `WCS_Action_Scheduler`, in that this class requires a toggle to enable or disable the feature. 🤔 

A follow-up question here would be related to how we handle this "global toggle." Should that be a checkbox? Or should we consider an empty value in the "days input" to indicate that it is disabled? I lean more toward the latter option.

xristos3490 (commented):
Amazing work, @peterfabian! Loved how you incorporated the `WCS_Scheduler` structure and the date types! Everything seem to be falling into place 🙌 

Left some initial thoughts on the structure before I start doing some hands-on testing.

includes/class-wcs-core-autoloader.php
peterfabian: You're probably right. This change makes the check more specific, so I think it makes sense in the context of folder structure in WCSC, but we can just add one more exclusion like with `wc_subscriptions_email` and keep it safer.

includes/class-wcs-action-scheduler-customer-notifications.php
peterfabian: [7078d87](https://github.com/Automattic/woocommerce-subscriptions-core/pull/648/commits/7078d871b4efafdf985943aff93cd533cc575ec6) ✅

includes/class-wcs-action-scheduler-customer-notifications.php
peterfabian: As I'm looking at it, I can see the emails in Woo have their own toggle for enabling/disabling, maybe we can reuse that (and if we provide a global disable, we can just set this toggle for all notifications?)

xristos3490: Yeah, we probably don't need a filter here at this point. 

I also think that we could pass the date type into the days-getter-filter and allow developers to disable specific types (e.g., by returning `0`.) 

Related question: Should we schedule notifications for disabled emails (based on the UI toggle)? I think we still should. This is different from disabling by returning `0` from the filter above, as we cannot schedule "0 days before the event".

includes/class-wcs-action-scheduler-customer-notifications.php
peterfabian: Done in [b614a4e](https://github.com/Automattic/woocommerce-subscriptions-core/pull/648/commits/b614a4e9bb82c9d380edb9d81df549d966fb686b).

includes/class-wcs-action-scheduler-customer-notifications.php
peterfabian: 👍 Good thinking.

includes/class-wcs-core-autoloader.php
peterfabian: Changed it in [ab517ff](https://github.com/Automattic/woocommerce-subscriptions-core/pull/648/commits/ab517ffece9052d6a1c57e5e7fcf840022fac82a).

includes/class-wcs-action-scheduler-customer-notifications.php
peterfabian: Agree, removed todo in [41ea0c7](https://github.com/Automattic/woocommerce-subscriptions-core/pull/648/commits/41ea0c75f584703ac5f1ef2786ebd95924892475).

includes/class-wcs-action-scheduler-customer-notifications.php
```diff
@@ -0,0 +1,290 @@
...
+
+		//TODO: filter?
+		$this->date_types_to_schedule = array_keys( $date_types_to_schedule );
+	}
+
+	protected function schedule_notification( $subscription, $action, $timestamp ) {
+		if (
+			! (
+				$subscription->has_status( 'active' )
+				|| $subscription->has_status( 'pending-cancel' ) //TODO: do we want to create notifications when user cancelled the subscription?
```
james-allan: > do we want to create notifications when user cancelled the subscription?

I think we'll need to test that to confirm but I don't think we would.

For some background the `WCS_Scheduler` hooks into both subscription date and status updates. 

When a subscription date is updated, we schedule the related AS event. 
When a subscription status is changed, we unschedule any events that are no longer applicable to that status and schedule events that are. eg putting a subscription on-hold cancels the payment event. Reactivating a subscription reschedules the next payment date.

I think this notification class should follow a similar principle.

So to answer the question then I don't think we do need to create notifications on the cancelled event because when a customer cancels the subscriptions, we calculate the end date and schedule the end date. The scheduling of that date is what should trigger the notification to be scheduled, not the status transition itself.

includes/class-wcs-action-scheduler-customer-notifications.php
```diff
@@ -0,0 +1,290 @@
...
+	 * @param WC_Subscription $subscription
+	 *
+	 * @return int
+	 */
+	protected function sub_time_offset( $datetime, $subscription ) {
+		$dt = new DateTime( $datetime, new DateTimeZone( 'UTC' ) );
+		$dt->sub( new DateInterval( "PT{$this->get_hours_offset( $subscription )}H" ) );
+
+		return $dt->getTimestamp();
+	}
```
james-allan, line 205: I might be missing the rest of the context here but I'm wondering if we should just use the number of seconds and timestamps? 

I noticed that the setting (interval and period) is being converted into hours and I'm wondering if it would be easier to convert it into seconds and then just do a straight subtraction.

includes/class-wcs-action-scheduler-customer-notifications.php
```diff
@@ -0,0 +1,290 @@
...
+				$this->schedule_trial_ending_notification( $subscription );
+			}
+
+			if ( $subscription->get_date( 'end' ) ) {
+				$this->schedule_expiry_notification( $subscription );
+			}
+
+			if ( $subscription->get_date( 'next_payment' ) ) {
+				$this->schedule_payment_notification( $subscription );
+			}
```
james-allan: This could be right but wanted to double check. When a subscription's date is updated you reschedule all dates? 

eg if the next payment date is updated, it would pass the `in_array()` check and then all subscription dates would be rescheduled.

includes/class-wcs-action-scheduler-customer-notifications.php
```diff
@@ -0,0 +1,290 @@
...
+	}
+
+	/**
+	 * Delete a date from the action scheduler queue
+	 *
+	 * @param object $subscription An instance of a WC_Subscription object
+	 * @param string $date_type Can be 'trial_end', 'next_payment', 'end', 'end_of_prepaid_term' or a custom date type
+	 */
+	public function delete_date( $subscription, $date_type ) {
+		$this->update_date( $subscription, $date_type, '0' );
```
james-allan: Just noting that the 3rd param isn't used in `update_date()`. It might still work though. Just wanted to flag as I skimmed :)

includes/class-wcs-action-scheduler-customer-notifications.php
```diff
@@ -0,0 +1,290 @@
...
+					$this->schedule_expiry_notification( $subscription );
+				}
+
+				if ( $subscription->get_date( 'next_payment' ) ) {
+					$this->schedule_payment_notification( $subscription );
+				}
+
+				break;
+			case 'pending-cancel':
+				// Unschedule all notifications?
```
james-allan: I don't think you would want to unschedule all, I think the expiration one is still one you'd want.

james-allan (commented):
This looks pretty good @peterfabian. I haven't had a chance to test it today, however I've given it a skim in terms of reviewing the general approach taken. 

I left some initial comments. :)

Just a heads up. I noticed in the PR `array()`. We have switched to use the shorthand syntax (`[]`) however haven't updated all existing uses. Our current approach is to use `[]` in newer code.

includes/class-wcs-action-scheduler-customer-notifications.php
xristos3490: Seems like a good idea to me. Working with seconds allows us more granular control. 
Also, since both the subscription's `get_date()` and the AS work in UTC by default, we should be aligned. Right?

includes/class-wcs-action-scheduler-customer-notifications.php
peterfabian: Thanks for elaborating. My todo comment was added to clarify whether we want to notify the customer about the upcoming expiry if they manually canceled the subscription (and thus should be aware that it's expiring).

includes/class-wcs-action-scheduler-customer-notifications.php
peterfabian: > Should we schedule notifications for disabled emails (based on the UI toggle)? I think we still should.

Yes, I think we should and the code indeed does that.

includes/class-wcs-action-scheduler-customer-notifications.php
peterfabian: > When a subscription's date is updated you reschedule all dates?

Yes. If the scheduling is correct, nothing is written to the db, but the code currently checks whether the scheduling is correct for all 3 types of notifications (and fixes it if it isn't).

includes/class-wcs-action-scheduler-customer-notifications.php
peterfabian: What do you mean it might still work? 😄 
I'm old-school, so if a method is declared with 3 arguments, I provide 3 arguments. 

Jokes aside, I removed the call as I updated the logic when notifications get scheduled.

includes/class-wcs-action-scheduler-customer-notifications.php
peterfabian: I generally prefer to work with standard DateTime classes when available to avoid some of the pitfalls mentioned here https://gist.github.com/timvisee/fcda9bbdff88d45cc9061606b4b923ca, but I suppose using timestamps would be easier here and should be ok.

includes/class-wcs-action-scheduler-customer-notifications.php
```diff
@@ -0,0 +1,340 @@
...
+
+/**
+ * Scheduler for subscription events that uses the Action Scheduler
+ *
+ * @class     WCS_Action_Scheduler
+ * @version   1.0.0 - Migrated from WooCommerce Subscriptions v2.0.0
+ * @package   WooCommerce Subscriptions/Classes
+ * @category  Class
+ */
+class WCS_Action_Scheduler_Customer_Notifications extends WCS_Scheduler {
```
xristos3490, line 11: While testing this, I discovered that when a user switches the payment type (manual/auto) from their My Account page, the scheduler does not recognize the change. See:
Markup 2024-09-04 at 16 28 25

This is likely unnecessary for the subscription core since it uses a single action for both types.

I think we could modify the constructor here to hook into the `woocommerce_subscription_object_updated_props` and reschedule when the `requires_manual_renewal` prop changes.

What do you think?

peterfabian: How do you get the toggle? I can't see it in my My Account.

xristos3490: Just to close the loop, as we already discussed through Slack. Here's the toggle:

peterfabian: To really close the loop, I changed the notification scheduling to only schedule one payment notification that decides which email to send based on subscription type at the time of notification.

includes/admin/debug-tools/class-wcs-notifications-debug-tool-processor.php
```diff
@@ -0,0 +1,311 @@
...
+		$allowed_statuses = array(
+			'active',
+			'pending',
+			'on-hold',
+		);
+
+		return array_map( 'wcs_sanitize_subscription_status_key', $allowed_statuses );
+	}
+
+	public function get_total_pending_count(): int {
```
james-allan, line 83: **Minor:** This function is missing a function comment.

includes/admin/debug-tools/class-wcs-notifications-debug-tool-processor.php
```diff
@@ -0,0 +1,311 @@
...
+	 * to use a different size.
+	 *
+	 * @return int Default batch size.
+	 */
+	public function get_default_batch_size(): int {
+		return 20;
+	}
+
+	/**
+	 * Start the background process for coupon data conversion.
```
james-allan:
```suggestion
	 * Start the background process for batch processing subscription notifications updates.
```

Looks like a copy-paste.

includes/admin/debug-tools/class-wcs-notifications-debug-tool-processor.php
```diff
@@ -0,0 +1,311 @@
...
+		if ( $batch_processor->is_enqueued( self::class ) ) {
+			return __( 'Background process for updating subscription notifications already started, nothing done.', 'woocommerce-subscriptions' );
+		}
+
+		$batch_processor->enqueue_processor( self::class );
+		return __( 'Background process for updating subscription notifications started', 'woocommerce-subscriptions' );
+	}
+
+	/**
+	 * Stop the background process for coupon data conversion.
```
james-allan:
```suggestion
	 * Stop the background process for batch processing subscription notifications updates.
```

includes/class-wc-subscriptions-email-notifications.php
```diff
@@ -0,0 +1,297 @@
+<?php
+
+/**
+ * Subscriptions Email Notifications Class
+ *
+ * Some details to enlighten your exploration of this code.
```
james-allan:
```suggestion
```

Were there supposed to be details that enlighten my exploration ? 😄

includes/class-wc-subscriptions-email-notifications.php
```diff
@@ -0,0 +1,297 @@
...
+
+	public static function init() {
+
+		add_action( 'woocommerce_email_classes', __CLASS__ . '::add_emails', 10, 1 );
+
+		add_action( 'woocommerce_init', __CLASS__ . '::hook_notification_emails' );
+
+		add_filter( 'woocommerce_order_actions', [ __CLASS__, 'add_notification_actions' ], 10, 1 );
+
+		// TODO this is a bit ugly...
```
james-allan: Agreed. We generally don't like using closures/anonymous functions especially attached to callbacks but we can address this separately if you're eager to merge this big PR.

includes/class-wc-subscriptions-email-notifications.php
```diff
@@ -0,0 +1,297 @@
...
+			1
+		);
+
+		add_filter( 'woocommerce_subscription_settings', [ __CLASS__, 'add_settings' ], 20 );
+
+		add_action( 'update_option_' . WC_Subscriptions_Admin::$option_prefix . self::$offset_setting_string, [ 'WC_Subscriptions_Email_Notifications', 'update_update_time' ] );
+		add_action( 'update_option_' . WC_Subscriptions_Admin::$option_prefix . self::$switch_setting_string, [ 'WC_Subscriptions_Email_Notifications', 'update_update_time' ] );
+	}
+
+	public static function update_update_time() {
```
james-allan: **Minor:** This is missing a function block comment. There are a few more below too.

includes/class-wc-subscriptions-email-notifications.php
```diff
@@ -0,0 +1,297 @@
...
+	public static function add_notification_actions( $actions ) {
+		global $theorder;
+
+		if ( wcs_is_subscription( $theorder ) ) {
+			$subscription = $theorder;
+			//TODO: confirm if these statuses make sense.
+			$allowed_statuses = [
+				'active',
+				'on-hold',
+				'pending-cancellation',
```
james-allan:
```suggestion
				'pending-cancel',
```

includes/class-wc-subscriptions-email-notifications.php
```diff
@@ -0,0 +1,297 @@
...
+	 *
+	 * @param array $actions An array of available actions
+	 * @return array An array of updated actions
+	 */
+	public static function add_notification_actions( $actions ) {
+		global $theorder;
+
+		if ( wcs_is_subscription( $theorder ) ) {
+			$subscription = $theorder;
+			//TODO: confirm if these statuses make sense.
```
james-allan: These make sense to me. The full list of statuses is:

```
wc-pending
wc-active
wc-on-hold
wc-cancelled
wc-switched
wc-expired
wc-pending-cancel
```

Of that list, I think notifications should be sent for active, on-hold, pending cancelled subscriptions. The others represent either un-started subscriptions or subscriptions that have ended.

includes/class-wc-subscriptions-email-notifications.php
```diff
@@ -0,0 +1,297 @@
...
+			$allowed_statuses = [
+				'active',
+				'on-hold',
+				'pending-cancellation',
+			];
+
+			if ( ! in_array( $subscription->get_status(), $allowed_statuses, true ) ) {
```
james-allan: **Nit:** This could be `$subscription->has_status( $allowed_statuses )`. But this is non-blocking. 

```suggestion
			if ( ! $subscription->has_status( $allowed_statuses ) ) {
```

includes/class-wcs-action-scheduler-customer-notifications.php
james-allan: > My todo comment was added to clarify whether we want to notify the customer about the upcoming expiry if they manually canceled the subscription (and thus should be aware that it's expiring)

Oh right. I think the answer is yes, even if the customer initiated the cancellation, we'd still want to send the pre-cancel email. The reason is that they may have cancelled the subscription weeks or months earlier and so notification of it being finally cancelled.

includes/class-wcs-action-scheduler-customer-notifications.php
```diff
@@ -0,0 +1,354 @@
...
+		// Here, we need the 'old' update timestamp for comparison, so can't use get_date_modified() method.
+		$subscription_update_time_raw = array_key_exists( 'date_modified', $subscription->get_data() ) ? $subscription->get_data()['date_modified'] : $subscription->get_date_created();
+		if ( ! $subscription_update_time_raw ) {
+			$subscription_update_utc_timestamp = 0;
+		} else {
+			$subscription_update_time_raw->setTimezone( new DateTimeZone( 'UTC' ) );
+			$subscription_update_utc_timestamp = $subscription_update_time_raw->getTimestamp();
+		}
+
+		$notification_settings_update_utc_timestamp = get_option( 'wcs_notification_settings_update_time' );
```
james-allan: I've been trying to understand if there's a potential for error here. If the `wcs_notification_settings_update_time` option isn't set, it will be `false`. Doing a comparison between a timestamp and false `int < false` may cause strange results. 

I'm not sure if we need to defend against that.

includes/class-wcs-action-scheduler-customer-notifications.php
```diff
@@ -0,0 +1,354 @@
...
+	}
+
+	public function set_date_types_to_schedule() {
+		$date_types_to_schedule = wcs_get_subscription_date_types();
+		unset(
+			$date_types_to_schedule['start'],
+			$date_types_to_schedule['cancelled'], // prevent scheduling end date when reactivating subscription.
+		);
+
+		$this->date_types_to_schedule = array_keys( $date_types_to_schedule );
```
james-allan: I got a question about this function that's based on the fact that `wcs_get_subscription_date_types()` is filterable. 

By default it contains:
```php
'start'
'trial_end'
'next_payment'
'last_payment'
'cancelled'
'end'
```

`start` and `cancelled` have been excluded which leaves `trial_end`, `next_payment`, `last_payment`, and `end`. `last_payment` is not a scheduled date, its something generated based on the subscription's order history so probably has no impact here. 

However there's another date type that can be returned from `wcs_get_subscription_date_types()`. Like `payment_retry` [ref](https://github.com/woocommerce/woocommerce-subscriptions/blob/6.5.0/includes/payment-retry/class-wcs-retry-manager.php#L140). 


I think the effect of that that means is that if a subscription's `payment_retry` is updated, we'd reschedule all the notifications. I don't know if that causes any issue other than an unnecessary update.

includes/class-wcs-action-scheduler-customer-notifications.php
james-allan: I'm also reading my comment and wondering what I meant by "it might still work". 🤔 😆 

I think I took that a function called `delete_date()` calling a function `update_date()` with `0` as a param to mean that passing `0` would cause the `update_date()` call to behave in a sort of delete mode. 

I was flagging that the third param in this version of `update_date()` isn't used and therefore '0' couldn't cause it to work in a delete mode. 

I think I was noting that I would need to come back to this. After looking at it a little deeper I don't know if I can tell there is a problem here. When a date is deleted, it will eventually call `schedule_all_notifications()` and that function only schedules notifications if a date exists. eg 

```php
if ( $subscription->get_date( 'next_payment' ) ) {
   $this->schedule_payment_notification( $subscription );
}
``` 

Because the date is deleted, that `if` condition would fail and it wouldn't schedule an event. That might be correct, but wouldn't we need to unschedule a related event if a relevant date is deleted?

james-allan: The way to test this I guess would be to have a subscription with a next payment date and a pre-renewal event scheduled. Then call `$subscription->delete_date( 'next_payment' )` and see if it correctly unschedules the event.

includes/class-wcs-notifications-batch-processor.php
```diff
@@ -0,0 +1,227 @@
...
+			'active',
+			'pending',
+			'on-hold',
+		);
+
+		return array_map( 'wcs_sanitize_subscription_status_key', $allowed_statuses );
+	}
+
+	public function get_notification_settings_update_time() {
+		$notification_settings_update_timestamp = get_option( 'wcs_notification_settings_update_time' );
```
james-allan: Just noting as I do a review that if this get_option returns `false` (`wcs_notification_settings_update_time` is not set), then this function will fatal because trying to create a `DateTime` with `@false` will error.

includes/emails/class-wcs-email-customer-notification-auto-renewal.php
```diff
@@ -0,0 +1,40 @@
...
+
+		$this->plugin_id = 'woocommerce-subscriptions_';
+
+		$this->id          = 'customer_notification_auto_renewal';
+		$this->title       = __( 'Customer Notification: Automatic Renewal for Subscription', 'woocommerce-subscriptions' );
+		$this->description = __( 'Customer Notification: Automatic Renewal emails are sent when a customer\'s subscription is about to be renewed automatically.', 'woocommerce-subscriptions' );
+
+		$this->heading = __( 'Subscription will be renewed automatically', 'woocommerce-subscriptions' );
+		// translators: placeholder is {blogname}, a variable that will be substituted when email is sent out
+		$this->subject = sprintf( _x( '[%s] Subscription Will Be Auto Renewed', 'default email subject for cancelled emails sent to the admin', 'woocommerce-subscriptions' ), '{blogname}' );
```
james-allan:
```suggestion
		$this->subject = sprintf( _x( '[%s] Subscription Will Be Auto Renewed', 'default email subject for upcoming auto renewal emails sent to the customer', 'woocommerce-subscriptions' ), '{blogname}' );
```

From another email by the looks. :)

includes/emails/class-wcs-email-customer-notification-manual-renewal.php
```diff
@@ -0,0 +1,40 @@
...
+
+		$this->plugin_id = 'woocommerce-subscriptions_';
+
+		$this->id          = 'customer_notification_manual_renewal';
+		$this->title       = __( 'Customer Notification: Manual Renewal for Subscription', 'woocommerce-subscriptions' );
+		$this->description = __( 'Customer Notification: Manual Renewal emails are sent when a customer\'s subscription needs to be manually renewed.', 'woocommerce-subscriptions' );
+
+		$this->heading = __( 'Subscription needs to be renewed manually', 'woocommerce-subscriptions' );
+		// translators: placeholder is {blogname}, a variable that will be substituted when email is sent out
+		$this->subject = sprintf( _x( '[%s] Subscription Needs Manual Renewal', 'default email subject for cancelled emails sent to the admin', 'woocommerce-subscriptions' ), '{blogname}' );
```
james-allan:
```suggestion
		$this->subject = sprintf( _x( '[%s] Subscription Needs Manual Renewal', 'default email subject for upcoming manual renewal emails sent to the customer', 'woocommerce-subscriptions' ), '{blogname}' );
```

includes/emails/class-wcs-email-customer-notification-subscription-expiration.php
```diff
@@ -0,0 +1,40 @@
...
+
+		$this->plugin_id = 'woocommerce-subscriptions_';
+
+		$this->id          = 'customer_notification_subscription_expiry';
+		$this->title       = __( 'Customer Notification: Subscription Is About To Expire', 'woocommerce-subscriptions' );
+		$this->description = __( 'Subscription Expiting Notification emails are sent when a customer\'s subscription is about to expire.', 'woocommerce-subscriptions' );
+
+		$this->heading = __( 'Subscription Expiring Notification', 'woocommerce-subscriptions' );
+		// translators: placeholder is {blogname}, a variable that will be substituted when email is sent out
+		$this->subject = sprintf( _x( '[%s] Subscription is about to expire', 'default email subject for cancelled emails sent to the admin', 'woocommerce-subscriptions' ), '{blogname}' );
```
james-allan:
```suggestion
		$this->subject = sprintf( _x( '[%s] Subscription is about to expire', 'default email subject for upcoming subscription expiration emails sent to the customer', 'woocommerce-subscriptions' ), '{blogname}' );
```

includes/emails/class-wcs-email-customer-notification.php
```diff
@@ -0,0 +1,139 @@
...
+		}
+
+		$result = $this->send( $this->get_recipient(), $this->get_subject(), $this->get_content(), $this->get_headers(), $this->get_attachments() );
+
+		if ( $result ) {
+			/* translators: 1: Notification type, 2: customer's email. */
+			$order_note_msg = sprintf( __( '%1$s was successfully sent to %2$s.', 'woocommerce-subscriptions' ), $this->title, $this->recipient );
+		} else {
+			/* translators: 1: Notification type, 2: customer's email. */
+			$order_note_msg = sprintf( __( 'Attempt to send %1$s to %2$s failed successfully.', 'woocommerce-subscriptions' ), $this->title, $this->recipient );
```
james-allan:
```suggestion
			$order_note_msg = sprintf( __( 'Attempt to send %1$s to %2$s failed.', 'woocommerce-subscriptions' ), $this->title, $this->recipient );
```

A successful failure. 🤔

templates/emails/customer-notification-auto-renewal.php
```diff
@@ -0,0 +1,63 @@
...
+			?>
+	</p>
+
+
+	<p>
+		<?php
+		echo wp_kses(
+			sprintf(
+				// translators: %1$s: name of the blog.
+				_x( 'Your subscription for XYZ on %1$s will be renewed automatically. Thanks for being a loyal customer with us!', 'In customer renewal invoice email', 'woocommerce-subscriptions' ),
```
james-allan: I'm assuming xyz is a placeholder for now but is that going to be the product or something else? 

Just curious because for subscriptions with multiple items, it could get unwieldy.

templates/emails/customer-notification-expiring-subscription.php
```diff
@@ -0,0 +1,64 @@
...
+			?>
+	</p>
+
+
+	<p>
+		<?php
+		echo wp_kses(
+			sprintf(
+			// translators: %1$s: name of the blog, %2$s: link to checkout payment url, note: no full stop due to url at the end
+				_x( 'Your subscription for XYZ on %1$s is about to expire. To keep receiving the goodies, renew it over here: %2$s', 'In customer renewal invoice email', 'woocommerce-subscriptions' ),
```
james-allan: I apologise if I've missed it, but we may need to consider a different word other than "renew". 

WooCommerce Subscriptions uses renew to describe the regular scheduled payments for a subscription. Subscriptions renew every month for example, but they only expire at their predetermined end date. 

Subscriptions expire (end automatically after x periods). eg $14/month for 12 months. After the 12 months, the subscription will automatically expire and the customer would need to resubscribe - not renew. 

Also, linking the customer to the subscription's checkout payment URL wouldn't work. That would send them to a URL like this `/checkout/order-pay/9269/?pay_for_order=true&key=wc_order_U168PYBA4ixPl` which is an empty pay for order page. 

Screenshot 2024-09-19 at 5 10 29 pm

TBH I'm not sure what the CTA would be for an upcoming subscription expiration because until it's expired, there's not a lot they can do. Once it is expired, they can resubscribe.

templates/emails/customer-notification-manual-renewal.php
```diff
@@ -0,0 +1,64 @@
...
+			?>
+	</p>
+
+
+	<p>
+		<?php
+			echo wp_kses(
+				sprintf(
+					// translators: %1$s: name of the blog, %2$s: link to checkout payment url, note: no full stop due to url at the end
+					_x( 'Your subscription for XYZ on %1$s is about to expire. To keep receiving the goodies, renew it manually over here: %2$s', 'In customer renewal invoice email', 'woocommerce-subscriptions' ),
```
james-allan: Similar feedback to above. Expire in WooCommerce Subscriptions means end. We may prefer to use "expire" in this customer facing way because that makes sense to customers -- in which case this feedback can be ignored. 😅 

Perhaps something like

"You subscription for xyz on %1$s is nearly due for renewal. To keep receiving..."

The CTA is also not likely to work here either because the renewal order wouldn't have been created yet and so they would need to wait until the renewal is created in order to pay for the renewal. There's a renewal invoice email that would be sent for that purpose.

james-allan (commented):
Hey @peterfabian and the rest of the team working on this! 💯 

Really nice work here. It's a monster of a PR to review.

I left a few inline comments but I'm happy if you want to turn those into issues so we can move forward with smaller issues and bug fixes.

A couple of general things I noticed that I thought were worth raising. 

1. `$subscription->get_time()` returns a subscription date as a timestamp. I noticed in a couple of places where a datetime object is being created from the `get_date()` mysql formatted value, and then converting into a timestamp via `getTimestamp()`. I'm curious if that's by design or if the approach taken in `get_time()` isn't correct or there's a preference with it. I ask because in general it would be my preference to use the established helpers to get subscription dates as a timestamp format rather than have multiple places where date formats are converted just from a DRY principle. 
2. Function block comments. I noticed there's a few functions missing function block comments. Understandably so given the size of the PR. I left a few inline comments before realising it's more widespread. Just wanted to flag it incase you respond to those inline comments. But as I mentioned, I'm happy to move forward and address that separately rather than hold up this PR on such a minor technicality. :)

includes/class-wc-subscriptions-email-notifications.php
```diff
@@ -0,0 +1,297 @@
...
+				return $actions;
+			}
+
+			if ( $subscription->get_date( 'trial_end' ) ) {
+				$actions['wcs_customer_notification_free_trial_expiration'] = esc_html__( 'Send Free Trial Expiration notification', 'woocommerce-subscriptions' );
```
mattallan: Super small nit-pick comment incoming.
The capitalization of these order notes isn't consistent with other similar WooCommerce's order actions:

| Order | Subscription |
|--------|--------|
|  | image | 

While I agree with your capitalization, I'm thinking we should change it to be consistent :)

```suggestion
				$actions['wcs_customer_notification_free_trial_expiration'] = esc_html__( 'Send free trial expiration notification', 'woocommerce-subscriptions' );
```

templates/emails/customer-notification-expiring-subscription.php
```diff
@@ -0,0 +1,64 @@
...
+		?>
+	</p>
+
+<?php
+/**
+ * @hooked WC_Emails::order_details() Shows the order details table.
+ *
+ * @since 8.0.0
+ */
+do_action( 'woocommerce_subscriptions_email_order_details', $order, $sent_to_admin, $plain_text, $email );
```
james-allan:
```suggestion
do_action( 'woocommerce_subscriptions_email_order_details', $subscription, $sent_to_admin, $plain_text, $email );
```


`$order` isn't a declared variable. Based on the other emails, I believe it's supposed to be subscription.

includes/emails/class-wcs-email-customer-notification-subscription-expiration.php
```diff
@@ -0,0 +1,40 @@
...
+
+		$this->template_html  = 'emails/customer-expiring-subscription.php';
+		$this->template_plain = 'emails/plain/customer-expiring-subscription.php';
```
james-allan:
```suggestion
		$this->template_html  = 'emails/customer-notification-expiring-subscription.php';
		$this->template_plain = 'emails/plain/customer-notification-expiring-subscription.php';
```

These files were missing the `notification` part of the file name.

includes/class-wcs-action-scheduler-customer-notifications.php
james-allan: So I've all but confirmed the issue above.

If you purchase a subscription and then delete the next payment date, the notification remains scheduled where there is no next payment date scheduled. 

ie Using the code snippet below. Deletes the next payment date but there remains a scheduled action.

```php
$subscription = wcs_get_subscription( 9276 );
$subscription->delete_date( 'next_payment' );
```
<p align="center">
Screenshot 2024-09-20 at 4 11 59 pm
</br>
<sup>No next payment date set in Edit subscription screen schedule metabox</sup>
</p> 

<p align="center">
Screenshot 2024-09-20 at 4 15 56 pm
</br>
<sup>The only scheduled event for this subscription is the notification - no next payment scheduled</sup>
</p>

james-allan:
I've tested the following scenarios: 

**SCHEDULING**

- [x] Purchase a new subscription
    - pre-renewal email is scheduled
- [x] Changing subscription date via UI updates the scheduled pre-renewal notification
- [x] Changing the Customer Notifications settings updates the schedule (eventually)
   - Only relevant Subscriptions (not on-hold)
- [x] Changing a subscription's status to on-hold cancels the scheduled jobs
- [x] Changing a subscription's status to active reschedules the jobs
- [x] Subscriptions purchased with a free trial schedule a pre-trial email and not a pre-renewal email.
- [x] Disabling the feature cancels all schedule notifications
    - \* See question below about a potential performance improve here.

**EMAILS**
- [x] Emails are sent when the feature is enabled
- [x] Emails are not sent when the feature is disabled
   - 🗒️  I had to manually trigger the `do_action()` to confirm this. All the Scheduled Actions were cancelled when the feature was disabled.
- [x] Running the actions from the edit subscription page sends the email
    - The pre-renewal emails (manual & automatic) worked. 
    - The pre-expiration email failed to send.
    - ⚠️ Noting that these notification actions are still shown when the feature is disabled and trying to trigger them will silently do nothing because the feature is disabled. We can probably not show these actions when the feature is disabled. 

<p align="center">
Screenshot 2024-09-20 at 3 26 57 pm
</br>
<sup>Failed to send pre-expiration email Order note</sup>
</p> 

In my email log, I see that the email failed to send with the error: `Message body empty`. This is related to [this](https://github.com/Automattic/woocommerce-subscriptions-core/pull/648/files#r1768024093) feedback item.

| Automatic Pre-renewal | Manual Pre-renewal |
|--------|--------|
| Screenshot 2024-09-20 at 3 22 15 pm | Screenshot 2024-09-20 at 3 28 00 pm | 

---

### Questions

**DISABLING THE SETTING**
If you disable the **Customer notifications** setting, all the actions are eventually cancelled. I understand this occurs in batches via the `WCS_Notifications_Batch_Processor` which is enqueued whenever the setting is changed - including disabling it. This process occurs, subscription by subscription. I wonder if we could simplify that and just unschedule all related events in one go via `as_unschedule_all_actions( 'woocommerce_scheduled_subscription_customer_notification_renewal' )` rather than traversing every subscription? This isn't a functionality concern because even if a notification action slipped through, it wouldn't send an email based on the logic checks at the time of running the action - it's the code and potential performance of the batch processor queue.

**ON-HOLD STATUS**

After testing things, I'm starting to wonder if we need to worry about the on-hold status. On-hold subscriptions are effectively paused either by the customer, or if the latest order failed payment and so, they won't renew, end or really do anything without being reactivated first. So the batch process and `add_notification_actions()` probably doesn't need to query or allow of on-hold subscriptions. The customer would need to reactivate the subscription for the subscription to renew or end and at that point we would schedule the relevant notification events. In fact, we may also be able to ignore `pending` subscriptions too. ie perhaps just active subscriptions are relevant for batch processing. 🤔

includes/class-wcs-notifications-batch-processor.php
```diff
@@ -0,0 +1,227 @@
+<?php
+
+use Automattic\WooCommerce\Internal\BatchProcessing\BatchProcessingController;
+use Automattic\WooCommerce\Internal\BatchProcessing\BatchProcessorInterface;
```
mattsherman: Just noting that these are `Internal` classes from WC Core. In general, I don't think we should be using `Internal` classes outside of WC Core, since they could be changed in non-backward compatible ways without consideration that there might be other extensions using them (hence why they are `Internal`).

includes/class-wc-subscriptions-email-notifications.php
```diff
@@ -0,0 +1,297 @@
...
+			}
+
+			if ( $subscription->get_date( 'end' ) ) {
+				$actions['wcs_customer_notification_subscription_expiration'] = esc_html__( 'Send Subscription Expiration notification', 'woocommerce-subscriptions' );
+			}
+
+			if ( $subscription->get_date( 'next_payment' ) ) {
+				if ( $subscription->is_manual() ) {
+					$actions['wcs_customer_notification_renewal'] = esc_html__( 'Send Manual Renewal notification', 'woocommerce-subscriptions' );
```
mattallan: Could we combine these renewal order actions? I don't know if merchants pay attention to whether the subscription is manually renewing or automatic.

Maybe the order action could just say "Send upcoming renewal notification" for both manual and automatic subscriptions?

includes/class-wc-subscriptions-email-notifications.php
```diff
@@ -0,0 +1,297 @@
...
+			if ( ! in_array( $subscription->get_status(), $allowed_statuses, true ) ) {
+				return $actions;
+			}
+
+			if ( $subscription->get_date( 'trial_end' ) ) {
+				$actions['wcs_customer_notification_free_trial_expiration'] = esc_html__( 'Send Free Trial Expiration notification', 'woocommerce-subscriptions' );
+			}
+
+			if ( $subscription->get_date( 'end' ) ) {
+				$actions['wcs_customer_notification_subscription_expiration'] = esc_html__( 'Send Subscription Expiration notification', 'woocommerce-subscriptions' );
```
mattallan: This is more general feedback for these order actions.

This might be silly but I could see there being some confusion from merchants for what "Send subscription expiration notification" means.

To help clarify these actions, I was wondering if we could add "upcoming" to make it clear that these emails are for future/upcoming events and not "your subscription has expired" notifications.

Some examples:

- Send trial is ending notification
- Send upcoming subscription expiration notification
  - this one might be too long
- Send upcoming renewal notification

includes/class-wc-subscriptions-email-notifications.php
mattallan: Could we rename this function name to avoid the double wording `update_update_`? 😅 

Maybe `set_notification_settings_update_time()`?

mattallan (commented):
Thanks @peterfabian for all your work on this one! It's been incredible catching up on this project after getting back from sabbatical and seeing it all come together in this PR 🎉

I've left some in-line comments but I also had some other feedback to share:

**File naming**

I found it a little difficult to find the files within `/includes` that related to this email notification feature, so I was thinking we could help that a little bit by either

1. Sticking to a common file name structure i.e.:
   - `class-wc-subscriptions-email-notifications.php`
   - `class-wc-subscriptions-email-notifications-scheduler.php`
   - `class-wc-subscriptions-email-notifications-batch-processor.php`
2. Move related files into a `includes/email-notifications` directory

**New Settings**

This is very very minor and I would be happy to move this to another issue but I noticed the sizing of the settings are off:

image

includes/class-wcs-action-scheduler-customer-notifications.php
peterfabian: Agreed!

includes/class-wcs-action-scheduler-customer-notifications.php
peterfabian: Switched to timestamps, so resolving this.

includes/class-wcs-action-scheduler-customer-notifications.php
peterfabian: Agree, should be resolved now.

peterfabian:
Thanks @james-allan @mattallan @mattsherman for taking the time to review the PR!

A general note, and apologies for not mentioning this in the PR description, but the subjects/texts of the emails were just temporary placeholders as we were waiting for the copy from this Studio request: https://woomarketingstudio.wordpress.com/2024/07/10/studio-request-notify-users-about-new-minimum-required-php-woocommerce-and-wordpress-versions-for-all-somewherewarm-extensions/

The wording will be updated based on that. Perhaps it would be good for you to have a quick look at the proposals over there to make sure they make sense in the context of Subscriptions and other emails.

includes/class-wc-subscriptions-email-notifications.php
peterfabian: True. I thought it might make sense for merchants to know _exactly_ what kind of email will be sent, but perhaps it's not an important aspect here.

includes/class-wcs-action-scheduler-customer-notifications.php
peterfabian: Well spotted! Thanks for noticing, fixed in [17bf275](https://github.com/Automattic/woocommerce-subscriptions-core/pull/648/commits/17bf27562f285ce407a341865991e6961d014114).

includes/admin/debug-tools/class-wcs-notifications-debug-tool-processor.php
peterfabian: Will be adding those comments in the next day or two.

includes/class-wc-subscriptions-email-notifications.php
peterfabian: To be added soon to elevate the enlightenment :)

includes/class-wc-subscriptions-email-notifications.php
peterfabian: Updated in [b7ee762](https://github.com/Automattic/woocommerce-subscriptions-core/pull/648/commits/b7ee762bcd2a6e62a1863ad57ff24c2d3b6cfcbb).

includes/class-wc-subscriptions-email-notifications.php
peterfabian: I see my puns go unappreciated. 😅 Updated `update_update_time` in [622336a](https://github.com/Automattic/woocommerce-subscriptions-core/pull/648/commits/622336a7d04e0ba8214d6f441bccfbf9c29d86a9).

includes/class-wc-subscriptions-email-notifications.php
peterfabian: Thanks for the confirmation. 🙌

includes/class-wcs-action-scheduler-customer-notifications.php
peterfabian: Good point, added default = 0 in [ec4c979](https://github.com/Automattic/woocommerce-subscriptions-core/pull/648/commits/ec4c97931759806e727169231ed28991f22f82b2).

includes/class-wcs-action-scheduler-customer-notifications.php
peterfabian: Thanks, that makes sense. Even though we wouldn't reschedule notifications (only verify their scheduling and update if it's incorrect), I think it's safer to just enumerate the date types we're interested in. Updated in [61de47c](https://github.com/Automattic/woocommerce-subscriptions-core/pull/648/commits/61de47cdda71b595354e020ccc9b9f1be40b124a).

includes/class-wcs-notifications-batch-processor.php
peterfabian: Thanks for noticing. @xristos3490 is fixing this in https://github.com/Automattic/woocommerce-subscriptions-core/pull/689/files#diff-d9eda2054181e7a8de782ad735d85bb637f35ad5c67dee9262b6518578cf6eb0R51

includes/emails/class-wcs-email-customer-notification.php
peterfabian: Just a reference to the famous Windows message, as I knew we would be replacing the placeholder texts with copy from Studio.
iKzHLJY

includes/emails/class-wcs-email-customer-notification-subscription-expiration.php
peterfabian: Nice catch!

peterfabian:
> ⚠️ Noting that these notification actions are still shown when the feature is disabled and trying to trigger them will silently do nothing because the feature is disabled. We can probably not show these actions when the feature is disabled.

Makes sense, fixed in [78e18e5](https://github.com/Automattic/woocommerce-subscriptions-core/pull/648/commits/78e18e5356b14e79047a48cc80338daafac5c719).

includes/class-wcs-notifications-batch-processor.php
peterfabian: Thanks for raising this, Matt, I initially missed that. Perhaps even more importantly, this won't work in WC before 9.0 as the way core dependency injection container loaded the BatchProcessor wouldn't allow it to use our class (which is outside of WC core's Internal namespace). This problem was fixed by [this PR](https://github.com/woocommerce/woocommerce/pull/46975), but since WC Subscriptions need to support WC 7.7+, it's a no-go for us. @xristos3490 is working on pulling relevant bits of batch processing from core to subscriptions core instead.

includes/class-wc-subscriptions-email-notifications.php
peterfabian: Makes sense. Fixed in [868ff38](https://github.com/Automattic/woocommerce-subscriptions-core/pull/648/commits/868ff381e18aeea1480d61c61723932e16e43083)

includes/class-wcs-notifications-batch-processor.php
xristos3490: Yes, thanks, Matt! This might apply to your thinking around gift cards. Here's the PR: 
- https://github.com/Automattic/woocommerce-subscriptions-core/pull/689

includes/class-wc-subscriptions-email-notifications.php
peterfabian: Updated in [00b87da](https://github.com/Automattic/woocommerce-subscriptions-core/pull/648/commits/00b87da0689c52c2d890597c07ee330cdc6ee892).

includes/class-wc-subscriptions-email-notifications.php
peterfabian: Agreed. Fixed in [ef80366](https://github.com/Automattic/woocommerce-subscriptions-core/pull/648/commits/ef80366e86eef421e521832c6a50eb9bea61ee82).

peterfabian:
> New Settings
> 
> This is very very minor and I would be happy to move this to another issue but I noticed the sizing of the settings are off:

This is a component [reused from WC core](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/includes/admin/class-wc-admin-settings.php#L743), so perhaps I can address it over there.

peterfabian:
> File naming
>
> I found it a little difficult to find the files within /includes that related to this email notification feature, so I was thinking we could help that a little bit by either
>
> 1. Sticking to a common file name structure i.e.:
> - class-wc-subscriptions-email-notifications.php
> - class-wc-subscriptions-email-notifications-scheduler.php
> - class-wc-subscriptions-email-notifications-batch-processor.php
> 2. Move related files into a includes/email-notifications directory

I'm happy to go either way, what would be your preference?

I tried to blend the files in as I haven't seen other features being separated in the codebase. Since `WC_Subscriptions_Email_Notifications` was inspired by `WC_Subscriptions_Email`, I used similar naming:
- class-wc-subscriptions-email.php
- class-wc-subscriptions-email-notifications.php

Similarly, `WCS_Action_Scheduler_Customer_Notifications` was inspired by `WCS_Action_Scheduler`, so I used similar naming convention:
- class-wcs-action-scheduler.php
- class-wcs-action-scheduler-customer-notifications.php

etc

templates/emails/customer-notification-expiring-subscription.php
peterfabian: You're right, thanks for raising this. I suppose the best option for now (until WCS supports early resubscribe) is to just leave out the CTA. I did that in the PR where I applied all the copy updates (https://github.com/Automattic/woocommerce-subscriptions-core/pull/694)

templates/emails/customer-notification-manual-renewal.php
peterfabian: This is a reminder that a manual payment is coming up and it seems to me that renewal flow with this URL `wcs_get_early_renewal_url` should work if it's enabled, can you please confirm?

includes/class-wc-subscriptions-email-notifications.php
```diff
@@ -0,0 +1,309 @@
...
+		// Shortcut to unschedule all notifications more efficiently instead of processing them subscription by subscription.
+		if ( ! self::notifications_globally_enabled() ) {
+			\WC_Subscriptions_Core_Plugin::instance()->notifications_scheduler->unschedule_all_notifications();
+
+			$message = __( 'Unscheduling all notifications now.', 'woocommerce-subscriptions' );
+		} else {
+			$message = WCS_Notifications_Batch_Processor::enqueue();
+		}
+
+		//TODO: make sure this works first.
```
xristos3490: FYI I fixed that [here](https://github.com/Automattic/woocommerce-subscriptions-core/pull/696/files#diff-7c3173cd33df01d8ffe14586bacc988f8dc4df5ba26f34ad6415f69c21f0271dR94)

includes/class-wcs-action-scheduler-customer-notifications.php
```diff
@@ -0,0 +1,447 @@
...
+		parent::__construct();
+
+		$setting_option    = get_option(
+			WC_Subscriptions_Admin::$option_prefix . WC_Subscriptions_Email_Notifications::$offset_setting_string,
+			[
+				'number' => 3,
+				'unit'   => 'days',
+			]
+		);
+		$this->time_offset = self::convert_offset_to_seconds( $setting_option );
```
xristos3490: Should we use the `set_time_offset()` here?

includes/class-wcs-action-scheduler-customer-notifications.php
```diff
@@ -0,0 +1,447 @@
...
+	}
+
+	protected function maybe_schedule_notification( $subscription, $action, $timestamp ) {
+		if ( ! WC_Subscriptions_Email_Notifications::notifications_globally_enabled() ) {
+			return;
+		}
+
+		if (
+			! (
+				$subscription->has_status( 'active' )
```
xristos3490: Nit: This could be `if ( ! $subscription->has_status( [ 'active', 'pending-cancel' ] ) { ...`

includes/class-wcs-action-scheduler-customer-notifications.php
```diff
@@ -0,0 +1,447 @@
...
+
+	/*
+	 * Subtract time offset from given datetime based on the settings and subscription properties and return resulting timestamp.
+	 *
+	 * @param string $datetime
+	 * @param WC_Subscription $subscription
+	 *
+	 * @return int
+	 */
+	protected function sub_time_offset( $datetime, $subscription ) {
```
xristos3490: Nit: Maybe rename this to `get_sub_time_offset()`?

xristos3490 (commented):
Looking good, @peterfabian! And [the issue](https://github.com/Automattic/woocommerce-subscriptions-core/pull/684#pullrequestreview-2339417564) with the async date changes has been resolved! Left some nit comments mostly.

includes/class-wcs-action-scheduler-customer-notifications.php
james-allan: I don't think this is getting a time offset though, it's subtracting an offset. Based on that it should probably be something like `subtract_time_offset()`.

includes/class-wcs-batch-processing-controller.php
```diff
@@ -0,0 +1,572 @@
...
+			}
+		);
+
+		add_action(
+			self::PROCESS_SINGLE_BATCH_ACTION_NAME,
+			function ( $batch_process ) {
+				$this->process_next_batch_for_single_processor( $batch_process );
+			},
+			10,
+			2
```
james-allan, line 82: This function is only accepting 1 arg isn't it? `$batch_process`

james-allan (commented):
Hi @peterfabian. Sorry for how long it has taken to review this one. 😞 

I think next week @mattallan and I need to move quickly to get this approved and merged so you folks can be released back to your normal work load. From there we can work from it as a base.

Our next scheduled release isn't until Nov 19th and so that should be plenty of time to have it running locally and apply fixes to if they come up.

I just skimmed the changes again today, there's a few function block comments etc missing, however that's all minor stuff. I'm going to carve out some time on Monday to really put it through the wringer and hopefully get it approved early.

peterfabian:
Thanks @james-allan 🙌  All your input will be highly appreciated!

includes/class-wc-subscriptions-email-notifications.php
peterfabian: Thanks!

includes/admin/debug-tools/class-wcs-notifications-debug-tool-processor.php
```diff
@@ -0,0 +1,327 @@
...
+	/**
+	 * Add the tool to start or stop the background process that manages notification batch processing.
+	 *
+	 * @param array $tools Old tools array.
+	 * @return array Updated tools array.
+	 */
+	public function handle_woocommerce_debug_tools( array $tools ): array {
+
+		if ( ! WC_Subscriptions_Email_Notifications::notifications_globally_enabled() ) {
+			$tools['start_add_subscription_notifications'] = array(
```
james-allan, line 292: ❓ Just so I understand, this specific tool is added as a "stub" of sorts to the **WooCommerce → Status → Tools** even though the feature is disabled just for reference? It will always be disabled when the feature isn't enabled.

james-allan (approved):
This looks good to merge imo. As I mentioned in my previous comment, this PR is super large and so I've mostly focused on reviewing critical architecture stuff. All the little things I noticed while testing and reviewing can be done in smaller, follow up PRs. I'll fill an issue to note those.

I've tested the following scenarios:

- [x] Enabling the setting schedules jobs for all valid subscriptions.
- [x] Changing the offset updates (cancels and reschedules) all jobs in the background. 
- [x] Setting a long pre-event time like 2 months won't schedule any events for subscriptions with an event shorter than 2 months. 
- [x] Setting a pre-event time like 7 days and then purchasing a weekly subscription won't schedule a pre-renewal email
    - **🔖 Note:** I think this is fine given the alternative would be to send an email at the time of purchase and that would be spamming the customer. In this case, the **"Thank you for your order"** sent for the order includes the next payment date and so it serves the purpose of notifying the customer of the next payment date.
- [x] cancelling an subscriptions schedules the pre-expiration email. 
- [x] Fully cancelling a subscription cancels the all pre-notification events. 
- [x] Manual vs automatic subscriptions send different emails.
- [x] Renewing early pushes out the scheduled notification.
- [x] Updating the subscription status updates the scheduled notifications.
- [x] Updating the date programmatically updates the scheduled notifications. 
- [x] Deleting a subscription date updates the scheduled notifications.
    - **🔖 Note:** If you delete a trial end date, it doesn't schedule the pre-renewal email. I'll follow up with this later.  
- [x] Emails aren't sent while in staging mode. 
- [x] Running the actions from the Edit Subscription > Actions dropdown works as expected. 
   - Noting that the actions listed match the subscription setup. Pre-trial only if in the trial period. Pre-renewal only when there's a next payment date and pre-end if there's a scheduled end date. 
- [x] Synced subscriptions work as expected. The pre-renewal email is scheduled x days prior to the sync date. 
- [x] Switching 🙈 
   - Switching to a subscription on a different period updates the events. 👍

includes/class-wcs-action-scheduler-customer-notifications.php
peterfabian: Agree, fixed in [53bca5e](https://github.com/Automattic/woocommerce-subscriptions-core/pull/648/commits/53bca5e75bc8111aad5e8538c8a196fb8ab138c7).

includes/class-wcs-action-scheduler-customer-notifications.php
peterfabian: Fixed in [53bca5e](https://github.com/Automattic/woocommerce-subscriptions-core/pull/648/commits/53bca5e75bc8111aad5e8538c8a196fb8ab138c7).

includes/class-wcs-action-scheduler-customer-notifications.php
peterfabian: Changed to `subtract_time_offset ` in [53bca5e](https://github.com/Automattic/woocommerce-subscriptions-core/pull/648/commits/53bca5e75bc8111aad5e8538c8a196fb8ab138c7).

includes/admin/debug-tools/class-wcs-notifications-debug-tool-processor.php
xristos3490: Yes, that's the idea. Instead of completely hiding this, we keep the UI consistent and use friendly messages. 

Initially, before the mass unscheduling, this was (kinda) necessary in case the progressing unscheduling failed. So, essentially, the same handling was maintained after the improvement. 

Perhaps the copy could use some improvement? I added some info in [this PR](https://github.com/Automattic/woocommerce-subscriptions-core/pull/704) if you'd like to take a look. 

Screenshot 2024-10-21 at 5 09 21 PM

includes/class-wcs-batch-processing-controller.php
xristos3490: Yes, I believe Peter is using a closure here to prevent others from removing that reference by calling "remove_action" for any reason.

cc @peterfabian

peterfabian: I think it was copied almost verbatim [from WC core](https://github.com/woocommerce/woocommerce/blob/trunk/plugins/woocommerce/src/Internal/BatchProcessing/BatchProcessingController.php#L74-L81) by you, @xristos3490  :) 
And yes, you're right @james-allan , there is just one parameter.

xristos3490: True that :) So, the reasoning for the closure is to avoid removals :)

james-allan: Yeah that's ok. We generally don't like using closures like this but given this is a copy of the WC Core file, we can live with that. :)

xristos3490 (approved):
The feature seems all set! I ran another smoke test across all parts, and I'm excited to share that everything looks great!

I noted a few areas in the welcome admin notice that might benefit from a little copy tweaking. Also, once we have the documentation URL, we can add a "Learn More" action that will be helpful!

Other than that, I think we're all set to merge! 🎉

FYI, I've included all the associated PRs in this epic:
- https://github.com/Automattic/woocommerce-subscriptions-core/issues/645

PR #677, opened 2024-09-11
Handle Subscription Notifications Rollout

xristos3490:
## Description

This PR includes a default value for two options that manage customer notifications. These options will only be updated for new installations by verifying the WCS's option for the previous version. The default offset for sending notifications is set to 3 days before.

Please follow these steps to test this PR:

1. Make sure that the `woocommerce_subscriptions_customer_notifications_enabled` and `woocommerce_subscriptions_customer_notifications_offset` options do not exist. If they do, please delete them.
2. Use one of the following: (a) a new store, (b) a store that has never installed subscriptions, or (c) delete the `woocommerce_subscriptions_previous_version` option from your test store.
3. Refresh a page.
4. Verify that the option defaults are added as expected.

## Product impact

### DIFF
--- a/includes/class-wc-subscriptions-core-plugin.php
+++ b/includes/class-wc-subscriptions-core-plugin.php
@@ -516,6 +516,18 @@ class WC_Subscriptions_Core_Plugin {
 				update_option( WC_Subscriptions_admin::$option_prefix . '_paypal_debugging_default_set', 'true' );
 			}
 
+			// If this is the first time activating WooCommerce Subscription we want to enable the customer email notifications (default to 3 days before.)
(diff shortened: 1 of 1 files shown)


peterfabian (approved):
I think this should work fine.

xristos3490:
@peterfabian FYI, I will merge this into the main branch. We will need to update the unit tests since test sites will have the feature enabled by default after this.

xristos3490:
Hey @peterfabian, just realized that this all is handled automatically by the `WC_Subscription_Admin::add_default_settings()` method [here](https://github.com/Automattic/woocommerce-subscriptions-core/blob/03797c709765450dae6572719269a0879a166f84/includes/class-wc-subscriptions-core-plugin.php#L509)

So, I'll be removing this fix. 😅

PR #679, opened 2024-09-11
WC version from constant should be checked if defined before accessed

### LINKED ISSUE #678: `WCS_Dependent_Hook_Manager::attach_woocommerce_dependent_hooks()` should check if a global constant is defined before accessing it
## Describe the bug
 
When trying to implement WCS Core library in a plugin, and that plugin is being uninstalled (deleted from the WP install) it is possible that the load order in WP causes 'WC_VERSION' not to be available in `plugins_loaded` where `WCS_Dependent_Hook_Manager::attach_woocommerce_dependent_hooks()` hooks to. This causes an error that makes the plugin non-deletable.

## Product impact

unfulvio:
Fixes #678 

## Description

See #678 -- this PR merely introduces a sanity check around `WC_VERSION`.

As an alternative or fallback, you could also check for `get_option( 'woocommerce_version' )`.

## Product impact

### DIFF
--- a/includes/class-wcs-dependent-hook-manager.php
+++ b/includes/class-wcs-dependent-hook-manager.php
@@ -48,7 +48,7 @@ class WCS_Dependent_Hook_Manager {
 		foreach ( self::$dependent_callbacks['woocommerce'] as $wc_version => $operators ) {
 			foreach ( $operators as $operator => $callbacks ) {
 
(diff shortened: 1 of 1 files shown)


wjrosa (approved):
Thank you for fixing this, @unfulvio ! Code looks good!

PR #684, opened 2024-09-16
Add notifications scheduling tests

peterfabian:
Adds tests to the subscription notifications. This should cover most workflows with the feature and changes to subscriptions.

## How to test this PR

Check those green marks ✅ 

## Product impact


xristos3490 (commented):
@peterfabian Testing this manually I get into a flow that doesn't seem to work as expected.

The issue and the suggested flow for a unit test:
- When I change the next payment date while a free trial is ON, only the free-trial notification is added. If I then delete the free-trial date **(without changing the next payment)**, I end up with no notification AS actions.

peterfabian:
Note: the tests are failing because of some global state is leaking between the test files. Will solve it on Monday.

includes/class-wcs-action-scheduler-customer-notifications.php
```diff
@@ -337,14 +337,16 @@ public function update_date( $subscription, $date_type, $datetime ) {
...
+	 * @param WC_Subscription $subscription An instance of a WC_Subscription object
 	 * @param string $date_type Can be 'trial_end', 'next_payment', 'end', 'end_of_prepaid_term' or a custom date type
 	 */
 	public function delete_date( $subscription, $date_type ) {
 		$action = $this->get_action_from_date_type( $date_type );
 		if ( $action ) {
 			$this->unschedule_actions( $action, self::get_action_args( $subscription ) );
 		}
+		//TODO: double check, if this is necessary.
+		$this->schedule_all_notifications( $subscription );
```
xristos3490, line 349: Theoretically, this is useful when a free trial date is deleted and a subsequent payment date needs to be scheduled.

includes/emails/class-wcs-email-customer-notification.php
```diff
@@ -80,9 +80,7 @@ public function trigger( $subscription_id ) {
 			|| ! WC_Subscriptions_Email_Notifications::should_send_notification()
 			|| WCS_Action_Scheduler_Customer_Notifications::is_subscription_period_too_short( $subscription )
 		) {
-			if ( is_admin() ) {
-				// TODO: add admin notice here
-			}
+			// TODO: add admin notice here if in admin
```
xristos3490, line 83: I'm assuming this addresses the situation when a store manager uses the subscription publish-metabox actions, right?

tests/unit/test-wcs-notifications.php
```diff
@@ -3,180 +3,2184 @@
...
+	 * Load data from protected properties of the tested classes.
+	 */
+	protected function init_protected_data() {
+		$reflection    = new ReflectionClass( 'WCS_Action_Scheduler_Customer_Notifications' );
+		$as_group_name = $reflection->getProperty( 'notifications_as_group' );
+		$as_group_name->setAccessible( true );
+
+		$instance = new WCS_Action_Scheduler_Customer_Notifications();
+
+		// Unhook the hooks from this extra instance that would mess up the tests.
```
xristos3490, line 459: Nit: It seems we should ideally make that a singleton. Perhaps a low-priority follow-up? I believe the base `WCS_Scheduler` might need to implement this change in the first place. 

Sounds kinda risky. Just thinking aloud.

tests/unit/test-wcs-notifications.php
```diff
@@ -3,180 +3,2184 @@
...
+			'Deleting trial_end date schedules next_payment notification.' =>
+				[
+					'callback'          => [ self::class, 'delete_subscription_date' ], // this should update to pending-cancelled.
+					'params'            => [ $subscription, 'trial_end' ],
+					'assertions_config' => [
+						[
+							'message' => 'Check that exactly one notification was deleted and one added.',
+							'type'    => 'assertTrue',
+							'actual'  => function ( $subscription, $actions_diff ) {
+								return $this->verify_notification_count( [ $subscription ], $actions_diff, 1, 0, 1 );
```
xristos3490, line 1979: Great stuff 😻

tests/unit/test-wcs-notifications.php
```diff
@@ -3,180 +3,2184 @@
...
+			]
+		);
+	}
+
 	/**
-	 * Mock function to check if an email was sent
+	 * Perform a recursive array diff.
 	 *
-	 * @return bool
+	 * Objects are compared just by their values, not if they are the same instance.
```
xristos3490, line 48: This appears crucial for understanding the format of the return value. We might consider including an example of the return value's state?

xristos3490 (commented):
Awesome work, @peterfabian! Left some comments before approving this.

includes/class-wcs-action-scheduler-customer-notifications.php
xristos3490: Expanding a bit more here: 

I believe the process might run twice due to the hook at `woocommerce_before_subscription_object_save`, but if someone deletes a date _directly (e.g., calling `$subscription->delete_date()`)_, it should trigger the data store's `save_dates()` function, which will necessitate a rescheduling here.

This might also relate to what @james-allan described in his follow-up [list](https://github.com/woocommerce/woocommerce-subscriptions/issues/4737): _"Deleting a free trial date doesn’t schedule the pre-renewal email."_?

tests/unit/test-wcs-notifications.php
```diff
@@ -3,180 +3,2184 @@
...
+	 * Wrapper to update the subscription's billing period (and return the subscription back).
+	 *
+	 * @param $subscription
+	 * @param $new_period
+	 *
+	 * @return WC_Subscription
+	 */
+	protected static function update_billing_period( $subscription, $new_period = 'year' ) {
+		$subscription->set_billing_period( $new_period );
+		$subscription->save(); // Is this needed?
```
xristos3490: I think so, yes.

tests/unit/test-wcs-notifications.php
```diff
@@ -3,180 +3,2184 @@
...
-	 * @return void
+	 * @param $subscription
+	 * @param $date_type
+	 *
+	 * @return WC_Subscription
 	 */
-	public function test_enable_disable_notifications() {
+	protected static function delete_subscription_date( $subscription, $date_type ) {
+		$subscription->delete_date( $date_type );
+		$subscription->save(); // Is this needed?
```
xristos3490: I think not. Most probably removing that will surface [the discussion](https://github.com/Automattic/woocommerce-subscriptions-core/pull/684#discussion_r1812710238) in the `delete_date()` function of the scheduler controller.

tests/unit/test-wcs-notifications.php
peterfabian: Yes, ideally we should prevent this. Since the notifications depend entirely on the subscription state, even multiple hooks should ideally be harmless, but I think I saw PHPUnit instantiating the class for each test method it runs, so these were all hooked 10 times each...

tests/unit/test-wcs-notifications.php
peterfabian: Seeing that e.g. `update_status` calls a bunch of `update_dates` and `delete_date` and after that also `WC_Subscription::save`, it might be needed actually?

tests/unit/test-wcs-notifications.php
peterfabian: Yup.

tests/unit/test-wcs-notifications.php
peterfabian: 🙌

includes/emails/class-wcs-email-customer-notification.php
peterfabian: Correct. I didn't like the "Subscription updated" message. A bit low priority, but I already spent a couple of hours untangling how the message gets passed between the requests to get something printed... Didn't seem like a priority.

tests/unit/test-wcs-notifications.php
xristos3490: It seems that the `delete_date()` function directly writes to the database, and whether a `save()` or any other function runs depends on the context. For instance, there are [existing unit tests](https://github.com/Automattic/woocommerce-subscriptions-core/blob/6bfbf519df799dde8cd9f39e6c93e820e5170352/tests/unit/test-class-wc-subscriptions.php#L904) that only verify that this function successfully writes to the database.

I believe that bulletproofing this process requires eliminating the "convenient" call to `save()`. What do you think?

tests/unit/test-wcs-notifications.php
xristos3490: It's quite interesting! I'm intrigued to make at least the notifications scheduler a singleton, but we can follow up on this given the timeframe we're working with right now.

tests/unit/test-wcs-notifications.php
peterfabian: Good idea, added.

tests/unit/test-wcs-notifications.php
peterfabian: Sure, our tests should still work, so I removed it.

peterfabian:
Thanks for the review 🙌 , I think it's ready for another go.

xristos3490 (approved):
Once again, awesome work, @peterfabian!

PR #687, opened 2024-09-17
Correctly update a subscription status upon losing a dispute

wjrosa:
Fixes 4465-gh-woocommerce/woocommerce-subscriptions

## Description

This PR correctly changes a subscription status when the related order loses a dispute. This is done by checking the current status of the subscription before updating it. If the status is `pending-cancel` we just update it to `cancelled`. 

## How to test this PR

Code review should be enough. Check if the tests are still passing. I have included specific unit tests for this change.

## Product impact

### DIFF
--- a/includes/class-wc-subscription.php
+++ b/includes/class-wc-subscription.php
@@ -459,7 +459,7 @@ class WC_Subscription extends WC_Order {
 
 					case 'pending':
 						// Nothing to do here
-					break;
+						break;
 
 					case 'pending-cancel':
 						// Store the subscription's end date and trial end date before overriding/deleting them.
@@ -470,7 +470,7 @@ class WC_Subscription extends WC_Order {
 						$end_date = $this->calculate_date( 'end_of_prepaid_term' );
 
 						// If there is no future payment and no expiration date set, or the end date is before now, the customer has no prepaid term (this shouldn't be possible as only active subscriptions can be set to pending cancellation and an active subscription always has either an end date or next payment), so set the end date and cancellation date to now
-						if ( 0 == $end_date || wcs_date_to_time( $end_date ) < current_time( 'timestamp', true ) ) {
+						if ( 0 == $end_date || wcs_date_to_time( $end_date ) < time() ) {
 							$cancelled_date = $end_date = current_time( 'mysql', true );
 						} else {
 							// the cancellation date is now, and the end date is the end of prepaid term date
@@ -485,18 +485,19 @@ class WC_Subscription extends WC_Order {
 								'end'       => $end_date,
 							)
 						);
-					break;
+						break;
 
 					case 'completed': // core WC order status mapped internally to avoid exceptions
 					case 'active':
-
 						if ( 'pending-cancel' === $old_status ) {
-							$this->update_dates( array(
-								'cancelled'    => 0,
-								'end'          => $this->meta_exists( 'end_date_pre_cancellation' ) ? $this->get_meta( 'end_date_pre_cancellation' ) : 0,
-								'trial_end'    => $this->meta_exists( 'trial_end_pre_cancellation' ) ? $this->get_meta( 'trial_end_pre_cancellation' ) : 0,
-								'next_payment' => $this->get_date( 'end' ),
-							) );
+							$this->update_dates(
+								array(
+									'cancelled'    => 0,
+									'end'          => $this->meta_exists( 'end_date_pre_cancellation' ) ? $this->get_meta( 'end_date_pre_cancellation' ) : 0,
+									'trial_end'    => $this->meta_exists( 'trial_end_pre_cancellation' ) ? $this->get_meta( 'trial_end_pre_cancellation' ) : 0,
+									'next_payment' => $this->get_date( 'end' ),
+								)
+							);
 						} else {
 							// Recalculate and set next payment date
 							$stored_next_payment = $this->get_time( 'next_payment' );
@@ -519,13 +520,13 @@ class WC_Subscription extends WC_Order {
 
 						// Trial end date and end/expiration date don't change at all - they should be set when the subscription is first created
 						wcs_make_user_active( $this->get_user_id() );
-					break;
+						break;
 
 					case 'failed': // core WC order status mapped internally to avoid exceptions
 					case 'on-hold':
 						// Record date of suspension - 'post_modified' column?
 						$this->set_suspension_count( $this->get_suspension_count() + 1 );
-					break;
+						break;
 					case 'cancelled':
 					case 'switched':
 					case 'expired':
@@ -542,7 +543,7 @@ class WC_Subscription extends WC_Order {
 						}
 
 						$this->update_dates( $dates_to_update );
-					break;
+						break;
 				}
 
 				// Make sure status is saved when WC 3.0+ is active, similar to WC_Order::update_status() with WC 3.0+ - set_status() can be used to avoid saving.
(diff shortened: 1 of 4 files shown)


mattallan (changes requested):
@wjrosa thanks for opening this PR!

The `_dispute_closed_status` order meta is only set and used by the Stripe gateway and so we shouldn't be writing changes to WooCommerce Subscriptions to support this specific meta.

WooCommerce Subscriptions, needs to be as gateway agnostic as possible, otherwise we'll have to write 50-60+ gateway support into subscriptions 😅 

---

Looking at the original GH issue, my understanding is that the problem has to do with our extension trying to mark the subscription as `on-hold` when calling `$subscription->payment_failed()` when the subscription already has `pending-cancel`.

To address this, I think our approach should be to update our `failed_subscription_sign_ups_for_order()` to check the status of the subscription before calling `$subscription->payment_failed()` and if the status is already `pending-cancel`, mark it as `cancelled`.

Or maybe we should update the default status in `$subscription->payment_failed()` so that it uses cancelled if the subscription is pending cancelled.

We will also want to look at updating the `payment_failed()` function because currently it adds an order note saying the subscription was cancelled due to maximum failed payments reached, which isn't relevant to this and many other cases 🤔 

image

I think this order note should only be used if the `woocommerce_subscription_max_failed_payments_exceeded` filter hook was used.

wjrosa:
> @wjrosa thanks for opening this PR!
> 
> The `_dispute_closed_status` order meta is only set and used by the Stripe gateway and so we shouldn't be writing changes to WooCommerce Subscriptions to support this specific meta.
> 
> WooCommerce Subscriptions, needs to be as gateway agnostic as possible, otherwise we'll have to write 50-60+ gateway support into subscriptions 😅
> 
> Looking at the original GH issue, my understanding is that the problem has to do with our extension trying to mark the subscription as `on-hold` when calling `$subscription->payment_failed()` when the subscription already has `pending-cancel`.
> 
> To address this, I think our approach should be to update our `failed_subscription_sign_ups_for_order()` to check the status of the subscription before calling `$subscription->payment_failed()` and if the status is already `pending-cancel`, mark it as `cancelled`.
> 
> Or maybe we should update the default status in `$subscription->payment_failed()` so that it uses cancelled if the subscription is pending cancelled.
> 
> We will also want to look at updating the `payment_failed()` function because currently it adds an order note saying the subscription was cancelled due to maximum failed payments reached, which isn't relevant to this and many other cases 🤔
> 
> image
> 
> I think this order note should only be used if the `woocommerce_subscription_max_failed_payments_exceeded` filter hook was used.

My initial thought was to use a very neutral name for the meta and document this new meta somewhere for other gateways to use it. But yeah, your suggestion is much simpler and smarter! Thank you so much! I have updated the code here and I will close the Stripe extension PR.

Also, I have simplified the filter code you mentioned and changed it to add the note only when the filter is executed in https://github.com/Automattic/woocommerce-subscriptions-core/pull/687/commits/70276dbaee5ffda3593b27716ca499d6af7d5fd5

includes/class-wc-subscription.php
```diff
@@ -1913,12 +1913,13 @@ public function payment_failed( $new_status = 'on-hold' ) {
...
 		// Allow a short circuit for plugins & payment gateways to force max failed payments exceeded
-		if ( 'cancelled' == $new_status || apply_filters( 'woocommerce_subscription_max_failed_payments_exceeded', false, $this ) ) {
-			if ( $this->can_be_updated_to( 'cancelled' ) ) {
-				$this->update_status( 'cancelled', __( 'Subscription Cancelled: maximum number of failed payments reached.', 'woocommerce-subscriptions' ) );
-			}
-		} elseif ( $this->can_be_updated_to( $new_status ) ) {
-			$this->update_status( $new_status );
+		$status_note = '';
+		if ( apply_filters( 'woocommerce_subscription_max_failed_payments_exceeded', false, $this ) ) {
+			$status_note = __( 'Subscription Cancelled: maximum number of failed payments reached.', 'woocommerce-subscriptions' );
```
mattallan: We need to also set `$new_status = 'cancelled';` when this filter is used to preserve backwards compatibility.

mattallan (changes requested):
Thanks @wjrosa for the later changes.

I noticed a small backwards compatibility issue for any gateway using the `woocommerce_subscription_max_failed_payments_exceeded` filter.

includes/class-wc-subscription.php
wjrosa: Gotcha! I have updated the whole implementation again to make this small change easier to read in https://github.com/Automattic/woocommerce-subscriptions-core/pull/687/commits/b7a92f35657d25a5480fdf8e4f15c753506ffd22

mattallan (changes requested):
Hey @wjrosa, so sorry to be a pain but this latest commit doubles the size of the PR to include PHPCS adjustments that I think should be handled in a separate PR so that we can discuss and review those changes on their own 😅

It's generally fine to address a small whitespace/PHPCS tweaks in the function you're updating, however, this PR is now updating all instances of `current_time( 'timestamp', true )` in the main subscriptions class which makes it difficult to review the changes  in this PR and understand which code is relevant to the fix.

Also if there's any reason we need to revisit these `current_time()` changes if it's in a separate PR, we can easily find/review the changes and revert it without that impacting other fixes :)

With that in mind, could you please remove the fixes to PHPCS violation fixes and create an issue/PR for the PHPCS changes? i.e. create an issue to replace all instances of `current_time( 'timestamp', true )` with `time()`?

Sorry and thanks 😄

includes/class-wc-subscription.php
```diff
@@ -1913,12 +1913,18 @@ public function payment_failed( $new_status = 'on-hold' ) {
 		$this->add_order_note( __( 'Payment failed.', 'woocommerce-subscriptions' ) );
 
 		// Allow a short circuit for plugins & payment gateways to force max failed payments exceeded
-		if ( 'cancelled' == $new_status || apply_filters( 'woocommerce_subscription_max_failed_payments_exceeded', false, $this ) ) {
-			if ( $this->can_be_updated_to( 'cancelled' ) ) {
-				$this->update_status( 'cancelled', __( 'Subscription Cancelled: maximum number of failed payments reached.', 'woocommerce-subscriptions' ) );
+		// This also forces the new status to be 'cancelled' if the filter is applied or the subscription is pending-cancel
+		$status_note                                 = '';
+		$max_failed_payments_exceeded_filter_applied = apply_filters( 'woocommerce_subscription_max_failed_payments_exceeded', false, $this );
```
mattallan: Small nit-pick: the large amount of whitespace here to align `=` looks a bit funky 🥲 What are your thoughts on shortening this variable name to something like: `$max_failed_payments`?

I was thinking we could also remove the nested if statement needed to set the status note by doing something like:

```
$max_failed_payments = apply_filters( 'woocommerce_subscription_max_failed_payments_exceeded', false, $this );
$status_note         = $max_failed_payments ? __( 'Subscription Cancelled: maximum number of failed payments reached.', 'woocommerce-subscriptions' ) : '';

if ( $max_failed_payments || $this->has_status( 'pending-cancel' ) ) {
	$new_status = 'cancelled';
}
```

Not sure if this is more readable or worse though. Happy to run with what you think is best :)

wjrosa: Yeah, I agree. Usually, it is best practice to name variables according to exactly what they are in PHP (citing PHP here because in Go, for example, it would be OK to name this `mfp`), even if it gets long. So, personally, my issue is more with the attribution alignment rule than with the variable name itself 😅 , but I understand this is a global rule. Change applied ✅ 

I loved your suggestion! I think it is simpler to understand 👍 . Also applied

wjrosa:
> Hey @wjrosa, so sorry to be a pain but this latest commit doubles the size of the PR to include PHPCS adjustments that I think should be handled in a separate PR so that we can discuss and review those changes on their own 😅
> 
> It's generally fine to address a small whitespace/PHPCS tweaks in the function you're updating, however, this PR is now updating all instances of `current_time( 'timestamp', true )` in the main subscriptions class which makes it difficult to review the changes in this PR and understand which code is relevant to the fix.
> 
> Also if there's any reason we need to revisit these `current_time()` changes if it's in a separate PR, we can easily find/review the changes and revert it without that impacting other fixes :)
> 
> With that in mind, could you please remove the fixes to PHPCS violation fixes and create an issue/PR for the PHPCS changes? i.e. create an issue to replace all instances of `current_time( 'timestamp', true )` with `time()`?
> 
> Sorry and thanks 😄

There's no need to apologize! Thanks for disclosing this kind of issue. I don't know if it's because I am used to automatically fixing lint issues (using my editor) on the files I work on or if it was the git commit hook. Still, it was not fully intentional. I reverted those changes now.

I don't even know why the `current_time` calls were changed. Do you have more info on this? Then I can work on those on another PR.

mattallan (approved):
Thanks @wjrosa !

I've tested these changes by:
1. Purchase a subscription
2. Mark the new subscription as pending-cancellation
3. Go to the parent order and mark it as failed
4. On `trunk` the subscription remains pending cancellation
5. On this branch the subscription transitions to cancelled 👍

PR #689, opened 2024-09-23
Add hosted version of the batch processor controller and fix related unit tests

xristos3490:
## Description

This pull request brings a hosted version of core's [BatchProcessingController](https://github.com/woocommerce/woocommerce/blob/37bb36994b08d417e34b9043fea0bf92a3f9c236/plugins/woocommerce/src/Internal/BatchProcessing/BatchProcessingController.php#L29) as `WCS_Batch_Processing_Controller` together with a new `WCS_Batch_Processor` interface. It also addressed required changes for the unit tests, adds more coverage, and fixes the provisioning for new installations.

### Reason for moving these interfaces

Before [this](https://github.com/woocommerce/woocommerce/pull/46975) fix was introduced (which was implemented in WC v9.0), the batch processing controller in Core could only work with "namespaced" processors. To avoid using polyfill classes, we decided to transfer the functionality to WCS for better stability. Plus, using internal-namespaced classes is considered an antipattern, so moving the functionality was the most direct solution.

## How to test this PR

Unit Tests 
+
Exploratory testing. Some suggested steps:
- Change the offset or enable it if it's disabled.
- Keep an eye on AS actions in the "Tools > Scheduled Actions" table.
- Observe any changes made to AS actions.
- Now, delete any notification-related AS actions.
- Go to "WooCommerce > Status > Tools" and click on the "Add notifications" tool.
- Keep an eye on the AS actions.
- Confirm that the expected AS actions are restored after the batcher has run.

## Product impact

### DIFF
--- a/includes/admin/debug-tools/class-wcs-notifications-debug-tool-processor.php
+++ b/includes/admin/debug-tools/class-wcs-notifications-debug-tool-processor.php
@@ -210,7 +210,7 @@ class WCS_Notifications_Debug_Tool_Processor implements BatchProcessorInterface 
 	 * @param array $batch Batch to process, as returned by 'get_next_batch_to_process'.
 	 */
 	public function process_batch( array $batch ): void {
-		// This is a bit unnecessary. Perhaps convert `update_status` to static to avoid instantiating the class?
+
 		$subscriptions_notifications = new WCS_Action_Scheduler_Customer_Notifications();
 
 		foreach ( $batch as $subscription_id ) {
--- a/includes/admin/debug-tools/class-wcs-notifications-debug-tool-processor.php
+++ b/includes/admin/debug-tools/class-wcs-notifications-debug-tool-processor.php
@@ -5,7 +5,7 @@
  *
  * @package  WooCommerce Subscriptions
  * @category Class
- * @since    8.0.0
+ * @since    x.x.x
  */
 class WCS_Notifications_Debug_Tool_Processor implements WCS_Batch_Processor {
 
--- a/includes/admin/debug-tools/class-wcs-notifications-debug-tool-processor.php
+++ b/includes/admin/debug-tools/class-wcs-notifications-debug-tool-processor.php
@@ -215,11 +215,16 @@ class WCS_Notifications_Debug_Tool_Processor implements WCS_Batch_Processor {
 	 */
 	public function process_batch( array $batch ): void {
 
-		$subscriptions_notifications = new WCS_Action_Scheduler_Customer_Notifications();
+		$subscriptions_notifications = WC_Subscriptions_Core_Plugin::instance()->notifications_scheduler;
 
 		foreach ( $batch as $subscription_id ) {
 			$subscription = wcs_get_subscription( $subscription_id );
-			$subscriptions_notifications->update_status( $subscription, $subscription->get_status(), null );
+
+			if ( WC_Subscriptions_Email_Notifications::notifications_globally_enabled() ) {
+				$subscriptions_notifications->update_status( $subscription, $subscription->get_status(), null );
+			} else {
+				$subscriptions_notifications->unschedule_all_notifications( $subscription );
+			}
 
 			// Update the subscription's update time to mark it as updated.
 			$subscription->set_date_modified( time() );
--- a/includes/admin/debug-tools/class-wcs-notifications-debug-tool-processor.php
+++ b/includes/admin/debug-tools/class-wcs-notifications-debug-tool-processor.php
@@ -296,7 +296,7 @@ class WCS_Notifications_Debug_Tool_Processor implements WCS_Batch_Processor {
 				'name'     => __( 'Start adding subscription notifications', 'woocommerce-subscriptions' ),
 				'button'   => __( 'Add notifications', 'woocommerce-subscriptions' ),
 				'disabled' => true,
-				'desc'     => __( 'This tool will add notifications to pending, active, and on-hold subscriptions. These updates will occur gradually in the background using Action Scheduler. Currently, there are no entries to convert.', 'woocommerce-subscriptions' ),
+				'desc'     => __( 'This tool will add notifications to pending, active, and on-hold subscriptions. These updates will occur gradually in the background using Action Scheduler. Currently, there are no subscriptions to process.', 'woocommerce-subscriptions' ),
 			);
 		} elseif ( $batch_processor->is_enqueued( self::class ) ) {
 			$tools['stop_add_subscription_notifications'] = array(
@@ -304,7 +304,7 @@ class WCS_Notifications_Debug_Tool_Processor implements WCS_Batch_Processor {
 				'button'   => __( 'Stop adding notifications', 'woocommerce-subscriptions' ),
 				'desc'     =>
 				/* translators: %1$d=count of total entries needing conversion */
-					sprintf( __( 'Stopping this will halt the background process that adds notifications to pending, active, and on-hold subscriptions. %1$d of %2$d entries remain to be converted.', 'woocommerce-subscriptions' ), $pending_count, ( $pending_count + $total_done ) ),
+					sprintf( __( 'Stopping this will halt the background process that adds notifications to pending, active, and on-hold subscriptions. %1$d of %2$d subscriptions remain to be processed.', 'woocommerce-subscriptions' ), $pending_count, ( $pending_count + $total_done ) ),
 				'callback' => array( $this, 'dequeue' ),
 			);
 		} else {
@@ -313,7 +313,7 @@ class WCS_Notifications_Debug_Tool_Processor implements WCS_Batch_Processor {
 				'button'   => __( 'Add notifications', 'woocommerce-subscriptions' ),
 				'desc'     =>
 				/* translators: %d=count of entries pending conversion */
-					sprintf( __( 'This tool will add notifications to pending, active, and on-hold subscriptions. These updates will occur gradually in the background using Action Scheduler. Currently, there are %d entries available for conversion.', 'woocommerce-subscriptions' ), $pending_count ),
+					sprintf( __( 'This tool will add notifications to pending, active, and on-hold subscriptions. These updates will occur gradually in the background using Action Scheduler. Currently, there are %d subscriptions to process.', 'woocommerce-subscriptions' ), $pending_count ),
 				'callback' => array( $this, 'enqueue' ),
 			);
 		}
--- a/includes/admin/debug-tools/class-wcs-notifications-debug-tool-processor.php
+++ b/includes/admin/debug-tools/class-wcs-notifications-debug-tool-processor.php
@@ -1,9 +1,13 @@
 <?php
 
-use Automattic\WooCommerce\Internal\BatchProcessing\BatchProcessingController;
-use Automattic\WooCommerce\Internal\BatchProcessing\BatchProcessorInterface;
-
-class WCS_Notifications_Debug_Tool_Processor implements BatchProcessorInterface {
+/**
+ * WooCommerce Subscriptions Notifications Debug Tool Processor.
+ *
+ * @package  WooCommerce Subscriptions
+ * @category Class
+ * @since    8.0.0
+ */
+class WCS_Notifications_Debug_Tool_Processor implements WCS_Batch_Processor {
 
 	/**
 	 * Option name for the tool state.
@@ -245,12 +249,13 @@ class WCS_Notifications_Debug_Tool_Processor implements BatchProcessorInterface 
 	 * @return string Informative string to show after the tool is triggered in UI.
 	 */
 	public function enqueue(): string {
-		$batch_processor = wc_get_container()->get( BatchProcessingController::class );
+		$batch_processor = WCS_Batch_Processing_Controller::instance();
 		if ( $batch_processor->is_enqueued( self::class ) ) {
 			return __( 'Background process for updating subscription notifications already started, nothing done.', 'woocommerce-subscriptions' );
 		}
 
 		$batch_processor->enqueue_processor( self::class );
+
 		return __( 'Background process for updating subscription notifications started', 'woocommerce-subscriptions' );
 	}
 
@@ -260,7 +265,7 @@ class WCS_Notifications_Debug_Tool_Processor implements BatchProcessorInterface 
 	 * @return string Informative string to show after the tool is triggered in UI.
 	 */
 	public function dequeue(): string {
-		$batch_processor = wc_get_container()->get( BatchProcessingController::class );
+		$batch_processor = WCS_Batch_Processing_Controller::instance();
 		if ( ! $batch_processor->is_enqueued( self::class ) ) {
 			return __( 'Background process for updating subscription notifications not started, nothing done.', 'woocommerce-subscriptions' );
 		}
@@ -276,23 +281,25 @@ class WCS_Notifications_Debug_Tool_Processor implements BatchProcessorInterface 
 	 * @return array Updated tools array.
 	 */
 	public function handle_woocommerce_debug_tools( array $tools ): array {
-		$batch_processor = wc_get_container()->get( BatchProcessingController::class );
+		$batch_processor = WCS_Batch_Processing_Controller::instance();
 		$pending_count   = $this->get_total_pending_count();
+		$state           = $this->get_tool_state();
+		$total_done      = isset( $state['last_offset'] ) ? (int) $state['last_offset'] : 0;
 
 		if ( 0 === $pending_count ) {
 			$tools['start_add_subscription_notifications'] = array(
 				'name'     => __( 'Start adding subscription notifications', 'woocommerce-subscriptions' ),
 				'button'   => __( 'Add notifications', 'woocommerce-subscriptions' ),
 				'disabled' => true,
-				'desc'     => __( 'This tool will add notifications to pending, active and on-hold subscriptions. This will happen overtime in the background (via Action Scheduler). There are currently no entries to convert.', 'woocommerce-subscriptions' ),
+				'desc'     => __( 'This tool will add notifications to pending, active, and on-hold subscriptions. These updates will occur gradually in the background using Action Scheduler. Currently, there are no entries to convert.', 'woocommerce-subscriptions' ),
 			);
 		} elseif ( $batch_processor->is_enqueued( self::class ) ) {
 			$tools['stop_add_subscription_notifications'] = array(
 				'name'     => __( 'Stop adding subscription notifications', 'woocommerce-subscriptions' ),
 				'button'   => __( 'Stop adding notifications', 'woocommerce-subscriptions' ),
 				'desc'     =>
-				/* translators: %d=count of entries pending conversion */
-					sprintf( __( 'This will stop the background process that adds notifications to pending, active and on-hold subscriptions. There are currently %d entries that can be converted.', 'woocommerce-subscriptions' ), $pending_count ),
+				/* translators: %1$d=count of total entries needing conversion */
+					sprintf( __( 'Stopping this will halt the background process that adds notifications to pending, active, and on-hold subscriptions. %1$d of %2$d entries remain to be converted.', 'woocommerce-subscriptions' ), $pending_count, ( $pending_count + $total_done ) ),
 				'callback' => array( $this, 'dequeue' ),
 			);
 		} else {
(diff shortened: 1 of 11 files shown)


includes/class-wc-subscriptions-email-notifications.php
```diff
@@ -20,23 +20,30 @@ class WC_Subscriptions_Email_Notifications {
...
 		add_action( 'woocommerce_order_action_wcs_customer_notification_free_trial_expiration', [ __CLASS__, 'forward_action' ], 10, 1 );
 		add_action( 'woocommerce_order_action_wcs_customer_notification_subscription_expiration', [ __CLASS__, 'forward_action' ], 10, 1 );
 		add_action( 'woocommerce_order_action_wcs_customer_notification_renewal', [ __CLASS__, 'forward_action' ], 10, 1 );
 
+		// Add settings UI.
 		add_filter( 'woocommerce_subscription_settings', [ __CLASS__, 'add_settings' ], 20 );
 
 		add_action( 'update_option_' . WC_Subscriptions_Admin::$option_prefix . self::$offset_setting_string, [ 'WC_Subscriptions_Email_Notifications', 'set_notification_settings_update_time' ] );
 		add_action( 'update_option_' . WC_Subscriptions_Admin::$option_prefix . self::$switch_setting_string, [ 'WC_Subscriptions_Email_Notifications', 'set_notification_settings_update_time' ] );
+		add_action( 'add_option_' . WC_Subscriptions_Admin::$option_prefix . self::$offset_setting_string, [ 'WC_Subscriptions_Email_Notifications', 'set_notification_settings_update_time' ] );
```
xristos3490, line 45: This addresses the issue where the callback is missed on the first write for existing installations.

includes/class-wcs-batch-processing-controller.php
```diff
@@ -0,0 +1,572 @@
...
+ * be dequeued once they notify that no more items are left to process (or when `force_clear_all_processes` is invoked).
+ * Failed batches will be retried after a while.
+ *
+ * This is heavily inspired by core's version at Automattic\WooCommerce\Internal\BatchProcessing\BatchProcessingController.
+ *
+ * @package  WooCommerce Subscriptions
+ * @category Class
+ * @since    8.0.0
+ */
+class WCS_Batch_Processing_Controller {
```
xristos3490, line 23: I tried to simplify this, but it seems pretty abstracted, already!

includes/admin/debug-tools/class-wcs-notifications-debug-tool-processor.php
```diff
@@ -210,7 +214,7 @@ public function get_next_batch_to_process( int $size ): array {
 	 * @param array $batch Batch to process, as returned by 'get_next_batch_to_process'.
 	 */
 	public function process_batch( array $batch ): void {
-		// This is a bit unnecessary. Perhaps convert `update_status` to static to avoid instantiating the class?
+
 		$subscriptions_notifications = new WCS_Action_Scheduler_Customer_Notifications();
```
peterfabian: Just a note here that I removed the instantiation in WCS_Notifications_Batch_Processor because `WCS_Action_Scheduler_Customer_Notifications` hooks `update_status`, `update_date` and `delete_date` via its parent `WCS_Scheduler` and wanted to avoid having double/triple hooks executed (even though they were implemented to be idempotent). Please see https://github.com/Automattic/woocommerce-subscriptions-core/pull/648/files#diff-d9eda2054181e7a8de782ad735d85bb637f35ad5c67dee9262b6518578cf6eb0R172

I think we should do the same here (not necessarily in this PR, but let's not forget)

includes/admin/debug-tools/class-wcs-notifications-debug-tool-processor.php
```diff
@@ -210,7 +214,7 @@ public function get_next_batch_to_process( int $size ): array {
...
```
peterfabian, line 219: Also, I suppose to be aligned with the approach taken in the ad-hoc batch processor, we should unschedule all notifications if they're disabled? 
https://github.com/Automattic/woocommerce-subscriptions-core/pull/648/files#diff-d9eda2054181e7a8de782ad735d85bb637f35ad5c67dee9262b6518578cf6eb0R176-R180

includes/admin/debug-tools/class-wcs-notifications-debug-tool-processor.php
```diff
@@ -301,7 +308,7 @@ public function handle_woocommerce_debug_tools( array $tools ): array {
 				'button'   => __( 'Add notifications', 'woocommerce-subscriptions' ),
 				'desc'     =>
 				/* translators: %d=count of entries pending conversion */
-					sprintf( __( 'This tool will add notifications to pending, active and on-hold subscriptions. This will happen overtime in the background (via Action Scheduler). There are currently %d entries that can be converted.', 'woocommerce-subscriptions' ), $pending_count ),
+					sprintf( __( 'This tool will add notifications to pending, active, and on-hold subscriptions. These updates will occur gradually in the background using Action Scheduler. Currently, there are %d entries available for conversion.', 'woocommerce-subscriptions' ), $pending_count ),
```
peterfabian: I don't think we should call it conversion. Maybe sth like "Currently, there are %d subscriptions to process."

includes/class-wc-subscriptions-core-plugin.php
```diff
@@ -164,10 +164,14 @@ public function init() {
...
 		add_action( 'plugins_loaded', 'WCS_Customer_Store::instance' );
 
+		// Initialise the batch processing controller.
+		add_action( 'init', 'WCS_Batch_Processing_Controller::instance' );
+
 		// Initialise the scheduler.
 		$scheduler_class = apply_filters( 'woocommerce_subscriptions_scheduler', 'WCS_Action_Scheduler' );
 		$this->scheduler = new $scheduler_class();
 
+		// Custom notifications scheduler.
```
peterfabian: Custom or Customer? :) Or Custom customer notifications scheduler? :)

includes/class-wc-subscriptions-core-plugin.php
```diff
@@ -529,16 +533,9 @@ public function activate_plugin() {
 				update_option( WC_Subscriptions_admin::$option_prefix . '_paypal_debugging_default_set', 'true' );
 			}
 
-			// If this is the first time activating WooCommerce Subscription we want to enable the customer email notifications (default to 3 days before.)
-			if ( '0' === get_option( WC_Subscriptions_Admin::$option_prefix . '_previous_version', '0' ) && false === get_option( WC_Subscriptions_Admin::$option_prefix . WC_Subscriptions_Email_Notifications::$offset_setting_string, false ) ) {
+			// Enable customer notifications by default.
```
peterfabian:
```suggestion
			// Enable customer notifications by default for new stores.
```
This is only for new stores, right?

includes/class-wc-subscriptions-email-notifications.php
peterfabian: Well spotted!

includes/class-wc-subscriptions-email-notifications.php
```diff
@@ -189,16 +208,17 @@ public static function should_send_notification() {
...
+	 * @param $subscription
 	 * @return bool
 	 */
 	public static function subscription_period_too_short( $subscription ) {
 		$period   = $subscription->get_billing_period();
 		$interval = $subscription->get_billing_interval();
 
 		// By default, there are no shorter periods than days in WCS, so we ignore hours, minutes, etc.
-		if ( $period <= 2 && 'day' === $interval ) {
+		if ( $interval <= 2 && 'day' === $period ) {
```
peterfabian, line 221: Nice catch! 🙌

includes/class-wcs-batch-processing-controller.php
```diff
@@ -0,0 +1,572 @@
...
+	 * Identifier of the action that will do the actual batch processing.
+	 * There's one action per enqueued processor that will keep rescheduling itself
+	 * as long as there are still pending items to process
+	 * (except if there's an error that caused no items to be processed at all).
+	 */
+	const PROCESS_SINGLE_BATCH_ACTION_NAME = 'wcs_run_batch_process';
+
+	const ENQUEUED_PROCESSORS_OPTION_NAME = 'wcs_pending_batch_processes';
+	const ACTION_GROUP                    = 'wcs_batch_processes';
+	const LOGS_CONTEXT                    = 'wcs-batch-processing';
```
peterfabian, line 42: Nice touch!

includes/class-wcs-notifications-batch-processor.php
```diff
@@ -33,9 +42,18 @@ protected function get_subscription_statuses() {
...
+	 */
 	public function get_notification_settings_update_time() {
-		$notification_settings_update_timestamp = get_option( 'wcs_notification_settings_update_time' );
-		$notification_settings_update_time      = new DateTime( "@$notification_settings_update_timestamp", new DateTimeZone( 'UTC' ) );
+		$notification_settings_update_timestamp = get_option( 'wcs_notification_settings_update_time', 0 );
+		if ( 0 === $notification_settings_update_timestamp ) {
+			return '';
+		}
+
+		$notification_settings_update_time = new DateTime( "@$notification_settings_update_timestamp", new DateTimeZone( 'UTC' ) );
```
peterfabian, line 56: Thanks for fixing this!

includes/interfaces/interface-wcs-batch-processor.php
```diff
@@ -0,0 +1,81 @@
+<?php
+/**
+ * WCS_Batch_Processor Interface
+ *
+ * Interface for batch data processors. See the WCS_Batch_Processing_Controller class for usage details.
+ *
+ * @package  WooCommerce Subscriptions
+ * @version  8.0.0
+ * @since    8.0.0
```
peterfabian: Side note: do we already know the version? Perhaps we should add it to more of the new code..

tests/unit/test-wcs-notifications-debug-tool.php
```diff
@@ -1,9 +1,34 @@
...
+
+		$this->assertTrue( $batch_processor->is_enqueued( WCS_Notifications_Debug_Tool_Processor::class ) );
+
+		// Process.
+		$reflection                              = new ReflectionClass( $batch_processor );
+		$process_next_batch_for_single_processor = $reflection->getMethod( 'process_next_batch_for_single_processor' );
+		$process_next_batch_for_single_processor->setAccessible( true );
+		$process_next_batch_for_single_processor->invoke( $batch_processor, WCS_Notifications_Debug_Tool_Processor::class );
+
+		$this->assertFalse( $batch_processor->is_enqueued( WCS_Notifications_Debug_Tool_Processor::class ) );
```
peterfabian, line 25: I suppose this means the processing has finished and the BatchProcessor dequeued itself, right? Might be good to add a comment here on why we assume this.

tests/unit/test-wcs-notifications-debug-tool.php
```diff
@@ -1,9 +1,34 @@
...
 
+	/**
+	 * Sanity check the controller first.
+	 *
+	 * @covers WCS_Notifications_Debug_Tool_Processor::process_batch
+	 *
+	 * @param array $data
+	 * @return bool
+	 */
+	public function test_batch_processesing_controller() {
```
peterfabian:
```suggestion
	public function test_batch_processing_controller() {
```

tests/unit/test-wcs-notifications-debug-tool.php
```diff
@@ -1,9 +1,34 @@
...
-use PHPUnit\Framework\TestCase;
-
 class WCS_Subscription_Notifications_Debug_Tool_Test extends WP_UnitTestCase {
 
+	/**
+	 * Sanity check the controller first.
+	 *
+	 * @covers WCS_Notifications_Debug_Tool_Processor::process_batch
+	 *
+	 * @param array $data
```
peterfabian:
```suggestion

```

tests/unit/test-wcs-notifications-debug-tool.php
```diff
@@ -1,9 +1,34 @@
...
+	 * @return bool
```
peterfabian:
```suggestion
	 * @return void
```

tests/unit/test-wcs-notifications-debug-tool.php
```diff
@@ -212,4 +155,78 @@ public function test_tool_state_while_processing() {
...
+				'billing_interval' => 1,
+			]
+		);
+
+		$simple_subscription->update_status( 'active' );
+		$simple_subscription->save();
+
+		$simple_subscription->update_dates(
+			[
+				'next_payment' => '2034-09-20 08:08:08',
```
peterfabian: Is this supposed to be "always" in the future? Why not use a relative date, then?

peterfabian: Also, is it a problem that 1-month subscription has next payment in 10 year's time? :)

tests/unit/test-wcs-notifications-debug-tool.php
```diff
@@ -14,17 +39,8 @@ class WCS_Subscription_Notifications_Debug_Tool_Test extends WP_UnitTestCase {
...
-			WC_Subscriptions_Admin::$option_prefix . WC_Subscriptions_Email_Notifications::$offset_setting_string,
-			[
-				'number' => '3',
-				'unit'   => 'days',
-			]
-		);
-
-		$batches = $this->notification_subscription_provider();
+		$batches   = $this->notification_subscription_data_provider();
+		$processor = new WCS_Notifications_Debug_Tool_Processor();
```
peterfabian, line 34: Might be better to call `\WCS_Action_Scheduler_Customer_Notifications::get_action_args` in the `$action_args` below.

tests/unit/test-wcs-notifications-emails.php
```diff
@@ -0,0 +1,118 @@
...
+				'billing_interval' => 3,
+			]
+		);
+
+		$too_short3 = WC_Subscriptions_Email_Notifications::subscription_period_too_short( $subscription3 );
+		$this->assertFalse( $too_short3 );
+	}
+
+	/**
+	 * Test subscription period too long.
```
peterfabian:
```suggestion
	 * Test global subscription notification switch.
```

tests/unit/test-wcs-notifications-emails.php
```diff
@@ -0,0 +1,118 @@
...
+		$this->assertTrue( $enabled );
+
+		$this->disable_notifications_globally();
+		$enabled = WC_Subscriptions_Email_Notifications::notifications_globally_enabled();
+		$this->assertFalse( $enabled );
+	}
+
+	/**
+	 * Helper to enable notifications globally.
+	 * TODO: We should create global helpers?
```
peterfabian, line 93: Yes :)

peterfabian (commented):
Overall, this looks nice! 

How do you see the robustness of the batch processing?

I'm thinking about scenario when one of the subscriptions will fail to get processed because it's in some invalid state or something (and an exception is thrown in the foreach loop). 

Then `process_batch` could theoretically never update the `last_offset` and total pending count will never go down, leaving the debug tool spinning forever.

I suppose we might have the same problem with the `WCS_Notifications_Batch_Processor` class if it wouldn't manage to update the subscription's update time...

tests/unit/test-wcs-notifications-emails.php
xristos3490: Haha! AI FTW! Nice catch; funny one!

tests/unit/test-wcs-notifications-debug-tool.php
xristos3490: > Is this supposed to be "always" in the future? Why not use a relative date, then?

Yes, I suppose I can make these dates dynamic. I wouldn't want someone 10 years from now cursing me while digging into a nasty unit test bug. 😅 

> Also, is it a problem that 1-month subscription has next payment in 10 year's time? :)

Yes, it seems odd. I believe it's related to our recent conversation about the differences between a "checkout" flow and the freestyle data model, which I actually don't mind.

includes/admin/debug-tools/class-wcs-notifications-debug-tool-processor.php
xristos3490: Done in https://github.com/Automattic/woocommerce-subscriptions-core/pull/689/commits/2332026e77506e4a944ae51dc3d92102704e6f18

includes/admin/debug-tools/class-wcs-notifications-debug-tool-processor.php
xristos3490: Done https://github.com/Automattic/woocommerce-subscriptions-core/pull/689/commits/2332026e77506e4a944ae51dc3d92102704e6f18

includes/admin/debug-tools/class-wcs-notifications-debug-tool-processor.php
xristos3490: Thanks! Done in https://github.com/Automattic/woocommerce-subscriptions-core/pull/689/commits/b564854949cfbe6694287c9c3e7f9a835b2591f0

We might want to revisit during polishing!

tests/unit/test-wcs-notifications-emails.php
xristos3490: Done https://github.com/Automattic/woocommerce-subscriptions-core/pull/689/commits/f9bee572a8cdd2f32271bfa2a4a05b743ca7ba54

tests/unit/test-wcs-notifications-debug-tool.php
xristos3490: Done https://github.com/Automattic/woocommerce-subscriptions-core/pull/689/commits/f9bee572a8cdd2f32271bfa2a4a05b743ca7ba54

tests/unit/test-wcs-notifications-debug-tool.php
xristos3490: Done https://github.com/Automattic/woocommerce-subscriptions-core/pull/689/commits/f9bee572a8cdd2f32271bfa2a4a05b743ca7ba54 and https://github.com/Automattic/woocommerce-subscriptions-core/pull/689/commits/725f13334e230eec6cfc0352dd123c5b111ba232

includes/interfaces/interface-wcs-batch-processor.php
xristos3490: Not sure tbh. Reverted to x.x.x in https://github.com/Automattic/woocommerce-subscriptions-core/pull/689/commits/38dd921e19b8f47452dea5d8e7d34bcd9d0f1ab5

tests/unit/test-wcs-notifications-debug-tool.php
xristos3490: Good idea! Done https://github.com/Automattic/woocommerce-subscriptions-core/pull/689/commits/e9e46076399926af6d0188cb41ccd8b3577b785f

xristos3490:
Thank you for the review, @peterfabian! I have addressed all of your feedback. Could you please take another look? 🙌 

> I'm thinking about scenario when one of the subscriptions will fail to get processed because it's in some invalid state or something (and an exception is thrown in the foreach loop).

Assuming that the `WCS_Batch_Processing_Controller` error handling catches the errors, each processor should eventually pass the `is_consistently_failing()` check and dequeue itself with a log message, right?

Given that there is no condition to be met for our "flags" to be modified and thus prevent an infinite loop, the only reason would be a fatal error, resulting in the system being in a constantly-failing mode.

tests/unit/test-wcs-notifications-emails.php
xristos3490: Perhaps we can do that alongside your tests! I'll follow up on a PR after this gets merged?

includes/admin/debug-tools/class-wcs-notifications-debug-tool-processor.php
xristos3490: @peterfabian I might be missing something, but could we avoid using a processor altogether and instead just remove every AS action for a given group via `as_unschedule_all_actions()`. Essentially, calling the `ActionScheduler_Store::instance()->cancel_actions_by_group()` behind the scenes? 

Same for the batch processor obviously.

peterfabian: No, I don't think you're missing anything. You're right that fewer abstractions and function calls are better. We can just call `as_unschedule_all_actions( WCS_Action_Scheduler_Customer_Notifications::$notifications_as_group )`. 

Essentially I just used it to encapsulate the group so that it's not a public property, but we can probably just make it public and call the AS function directly.

tests/unit/test-wcs-notifications-emails.php
peterfabian: Sounds good.

peterfabian (approved):
All the points I raised have been addressed, so let's merge this! 
Great work 💪 !

Extra kudos for making the tests pass 👌

PR #690, opened 2024-09-25
Tests for possible fatal errors when getting order

strutalan:
Submitting PR per support request

## Description

Encountered in the wild, I found an issue where fatal PHP errors could be thrown from class-wcs-cart-renewal on line 1314 and wcs-renewal-functions on line 112. 
This resulted in one customer in particular encountering a fatal error on all page requests immediately after logging in.

PHP Fatal error:  Uncaught Error: Call to a member function get_fees() on bool in
/wp-content/plugins/woocommerce-subscriptions/vendor/woocommerce/subscriptions-core/includes/class-wcs-cart-renewal.php:1314

PHP Fatal error:  Uncaught Error: Call to a member function has_status() on bool in
/wp-content/plugins/woocommerce-subscriptions/vendor/woocommerce/subscriptions-core/includes/wcs-renewal-functions.php:112

The submitted change checks that an order is preset before attempting to call methods that would result in a 500 error if the order is false.

## How to test this PR

How to replicate the apparent issue with a past order isn't known but suspect it was due to a manual deletion or cancellation of a past subscription via the admin. 

Tested the patch live on the site which had this problem by signing in as that user, used the site as normal, no other issues found.

### DIFF
--- a/includes/class-wcs-cart-renewal.php
+++ b/includes/class-wcs-cart-renewal.php
@@ -1307,15 +1307,19 @@ class WCS_Cart_Renewal {
 			return;
 		}
 
-		$cart_fees = $cart->get_fees();
+		$renewal_order = $this->get_order();
+
+		if ( ! $renewal_order ) {
+			return;
+		}
 
 		// Fees are naturally recurring if they have been applied to the renewal order. Generate a key (name + amount) for each fee applied to the order.
 		$renewal_order_fees = array();
-		if ($this->get_order()){ // fatal error casued by the foreach, if get_order returns false
-			foreach ( $this->get_order()->get_fees() as $item_id => $fee_line_item ) {
-				$renewal_order_fees[ $item_id ] = $fee_line_item->get_name() . wc_format_decimal( $fee_line_item->get_total() );
-			}
-	        }
+		$cart_fees          = $cart->get_fees();
+
+		foreach ( $renewal_order->get_fees() as $item_id => $fee_line_item ) {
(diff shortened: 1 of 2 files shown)


mattallan (approved):
Thanks @strutalan for the PR. I pushed a couple of commits to make some small adjustments and fix whitespace issues.

I tried replicating the issue but didn't have much luck. I thought maybe that it was caused by adding a pending renewal order to the cart and then deleting that order from the DB and refreshing the checkout page. I thought this would result in the cart item data for the renewal order still existing but fetching the order would return `false`... but that didn't reproduce it.

The changes here are making our code more defensive so I believe it's good to merge. I do have concerns that we might be suppressing a larger issue where existing orders aren't being fetched by `wc_get_order()` but this prevents a fatal under these rare occurrences.

strutalan:
> Thanks @strutalan for the PR. I pushed a couple of commits to make some small adjustments and fix whitespace issues.
> 
> I tried replicating the issue but didn't have much luck. I thought maybe that it was caused by adding a pending renewal order to the cart and then deleting that order from the DB and refreshing the checkout page. I thought this would result in the cart item data for the renewal order still existing but fetching the order would return `false`... but that didn't reproduce it.
> 
> The changes here are making our code more defensive so I believe it's good to merge. I do have concerns that we might be suppressing a larger issue where existing orders aren't being fetched by `wc_get_order()` but this prevents a fatal under these rare occurrences.

Thanks @mattallan that's a good point about wc_get_order(), though I don't know WC well enough to anticipate what could happen here specifically.

Yeah I wasn't able to replicate either, still don't know exactly how it transpired but appreciate the consideration!

Thanks
Alan

PR #691, opened 2024-09-25
Subscription status constants

wjrosa:
## Description

This PR introduces constants for the subscription's statuses. This should make it easier to track all the statuses, group them, create specific logic, and create new code with all the possible values.

## How to test this PR

Code review should be enough. Check if the tests are still passing. No behavior should change with this PR. You are encouraged to perform some smoke testing.

## Product impact


mattallan (commented):
Hey @wjrosa, thanks for opening this PR and starting the discussion.

I’m curious about the decision behind introducing these constants, as I’m not entirely clear on the benefits over using the status strings directly.

In my opinion, these changes makes our code less easier to read and understand, and potentially reduces flexibility for making future changes.

For me, I like to think of WooCommerce Subscriptions development in alignment with WooCommerce core practices. Therefore, I’d prefer to see a similar change introduced in WooCommerce before we adopt this approach in Subscriptions.

Additionally, when using these constants, it's unclear whether you're getting the 'wc-active' or 'active' version of the status. This adds potential confusion.

It feels strange to use the constant in some places but still rely on hardcoded values in others (e.g., 'wc-active' needed to register the post type and do direct DB queries, and 'active' for other logic), which somewhat undermines the purpose of using constants 😅.

---

1. Increased Code Fragmentation

Introducing constants moves the status values away from relevant logic, meaning developers have to jump between files to understand the code fully. This is typically fine in smart code editors, but when working outside an IDE or for non-technical users reviewing code in GitHub, it's less clear what these constants represent. 

This has potential slow down development and debugging.

2. Familiarity
Using constants for post or order statuses isn't a standard practice across WooCommerce or WordPress to my knowledge.

Developers across A8C and outside of our company interact with our subscriptions code and are accustomed to working directly with status strings, so I think we should be consistent with other similar code.

---

There are certainly benefits to using constants and they're especially helpful in cases where a value is used across multiple files and can be easily updated in one location or i.e.

```
const API_URL = 'https://api.example.com/v1/';
```

But I'm sad to admit that I don’t see the benefits to making the changes in this PR that are worth the overhead of refactoring the entire codebase to use status constants, as this could introduce more complexity and potential new issues.

wjrosa:
Hi Matt! Thanks for the detailed opinion!

> I’m curious about the decision behind introducing these constants, as I’m not entirely clear on the benefits over using the status strings directly.

As the definition says, constants should be used to store values that won't change. Statuses are good examples since they will always be the same and have minimal chances of changing. When they change, updating the constant value instead of all occurrences across the codebase is easier.

> In my opinion, these changes makes our code less easier to read and understand, and potentially reduces flexibility for making future changes.

Why do you think it is harder to read? You are right about the flexibility, but that's one of the goals: make the code stricter so it is easier to catch bugs and more guaranteed to be safe. Here's a [good article](https://lucidmodules.com/blog/php/why-should-you-use-strict-types/) about it.

> For me, I like to think of WooCommerce Subscriptions development in alignment with WooCommerce core practices. Therefore, I’d prefer to see a similar change introduced in WooCommerce before we adopt this approach in Subscriptions.

My idea was to introduce it there first, but I'm not sure if we can 😅 . I will do some research on it.

> Additionally, when using these constants, it's unclear whether you're getting the 'wc-active' or 'active' version of the status. This adds potential confusion.

> It feels strange to use the constant in some places but still rely on hardcoded values in others (e.g., 'wc-active' needed to register the post type and do direct DB queries, and 'active' for other logic), which somewhat undermines the purpose of using constants 😅.

They should be different constants. I was going to draft another PR with constants for the legacy statuses.

> This is typically fine in smart code editors, but when working outside an IDE or for non-technical users reviewing code in GitHub, it's less clear what these constants represent.

Why would a professional developer code without IDEs in 2024? I don't see any good reasons for it, except for someone who likes to do things in a "hardcode way." That would be a personal choice, an exception, and should not govern how everyone else develops code. We should aim for the best and not adapt ourselves to the worst (meaning: it is extremely unlikely that someone using notepad would do a better job in coding and debugging without all the tools IDEs provide).

The same goes for "non-technical users reviewing code." Shouldn't we expect only technical users to do it? How could someone who is not familiar with code give better advice to us who work with it the entire day? Maybe they could advise on business rules, but that should not be done in Github, as it is totally code-centered.

> Using constants for post or order statuses isn't a standard practice across WooCommerce or WordPress to my knowledge.

You're right, but I believe they should be. This is a basic coding standard that increases the quality of code. [Another good article about it](https://medium.com/@bluznierca1/constants-vs-string-literals-crafting-cleaner-php-code-63f4c4fee176).

> Developers across A8C and outside of our company interact with our subscriptions code and are accustomed to working directly with status strings, so I think we should be consistent with other similar code.

But shouldn't the code always evolve following the evolution of the language and best practices? Just like we introduced ternaries and coalesce operators instead of IFs, typed properties, arrays simplification, use of try...catch, or even class constants (just comparing random PHP 5 x 7 stuff).

I will close both PRs for now and understand why we don't have this yet in WC. Maybe I come back with it again if this scenario changes.

PR #692, opened 2024-09-26
Introducing order status constants

wjrosa:
## Description

This PR introduces constants for the WC order statuses. This should make it easier to track all the statuses, group them, create specific logic, and create new code with all the possible values.

## How to test this PR

Code review should be enough. Check if the tests are still passing. No behavior should change with this PR. You are encouraged to perform some smoke testing.

## Product impact


mattallan (commented):
Hey @wjrosa, I just left a comment on your other PR that is relevant to this one as well. See comment: https://github.com/Automattic/woocommerce-subscriptions-core/pull/691#pullrequestreview-2341909150

Since this PR is focused on Order statuses, I think I'd prefer to see this change made to WooCommerce core and not something we should develop into subscriptions.

Another way to think of this is that it doesn't make sense for Stripe, WooPayments, PayPal and every other extension to also introduce their own order status constants as that's a lot of work to maintain. 😅 

FWIW I had a look at github.com/woocommerce/woocommerce and couldn't find any issues or PRs for storing statuses in constants but there must also be a reason this hasn't been introduced yet. 🤔

wjrosa:
Thanks, Matt! I will be closing this and [the other PR](https://github.com/Automattic/woocommerce-subscriptions-core/pull/691) until I have more arguments to do it.

> Another way to think of this is that it doesn't make sense for Stripe, WooPayments, PayPal and every other extension to also introduce their own order status constants as that's a lot of work to maintain. 😅

Actually, one of the reasons why I did this work was because WCPay [did the same](https://github.com/Automattic/woocommerce-payments/blob/develop/includes/constants/class-order-status.php) (for multiple contexts). PayPal also uses it for [their own statuses](https://github.com/woocommerce/woocommerce-paypal-payments/blob/1f3034136e38cf6d00bf5dd8378db170e725a138/modules/ppcp-api-client/src/Entity/OrderStatus.php#L24). 

> FWIW I had a look at github.com/woocommerce/woocommerce and couldn't find any issues or PRs for storing statuses in constants but there must also be a reason this hasn't been introduced yet. 🤔

I will find out why 👀

PR #694, opened 2024-09-27
Updated the copy for subscription notification emails.

peterfabian:
Updates the copy of emails based on the doc https://docs.google.com/document/d/1CKfmxlvzoXbAOkAwsLI5HTkdNI6m7FOwkwZMsVxIGNQ/edit

## Notes 

For free trial expiration with manual payment and expiring subscription, I removed the call to action buttons/link, as:
- free trial can't be renewed early before the trial is up, as per https://woocommerce.com/document/subscriptions/early-renewal/#early-renewal-requirements
- a subscription that's about to expire can't be resubscribed until it expires, as per: https://woocommerce.com/document/subscriptions/customers-extend-subscription-expiration-date/ (there's a theoretical option that the store has subscriptions set up in the way that allows switching them to longer durations, but that seems like a pretty long shot)

Manual renewal reminder: early renewals might be disabled, so I slightly adjusted the wording and CTA in case the feature is disabled (or not allowed for the subscription).


xristos3490 (approved):
This looks great, @peterfabian! Nice work, everyone! cc @manospsyx @garethallison

I've tested all the emails, and they are looking as expected.

|Case|Email|
| --- | --- |
| **Renewal: Automatic Payment** | re-auto |
| **Renewal: Manual Payment**  | re-manual |
|**Renewal: Manual Payment** _(with no accepted early renewals; the URL goes to the my account page instead of the early renewal page)_ | re-manual-no-cta|
| **Free Trial: Automatic Payment** | free-auto |
| **Free Trial: Manual Payment** _(is missing the "Subscribe Now" CTA since it doesn't seem to be a way to create that flow yet.)_ | free-manual |
|**Expiration**|expiration|

PR #696, opened 2024-10-02
Attempt to bulk unschedule notification actions when the feature gets disabled

xristos3490:
## Description

This PR:
- Uses the `as_unschedule_all_actions( null, [], 'wcs_customer_notifications' );` function to unschedule all actions with the same group in bulk. It also prevents the batch processor from being enqueued when the feature is disabled.
- Disables the debug tool if the feature is disabled.

## How to test this PR

TBA

## Product impact


includes/class-wc-subscriptions-email-notifications.php
```diff
@@ -88,14 +88,13 @@ public static function set_notification_settings_update_time() {
 
 		// Shortcut to unschedule all notifications more efficiently instead of processing them subscription by subscription.
 		if ( ! self::notifications_globally_enabled() ) {
-			\WC_Subscriptions_Core_Plugin::instance()->notifications_scheduler->unschedule_all_notifications();
-
-			$message = __( 'Unscheduling all notifications now.', 'woocommerce-subscriptions' );
+			as_unschedule_all_actions( null, [], 'wcs_customer_notifications' );
```
xristos3490, line 91: This entire logic looks like it needs to be transferred to the `WCS_Action_Scheduler_Customer_Notifications` controller.

peterfabian, line 91: I think it would be better if the group name weren't hardcoded here as a string. That was one of the reasons why I decided to use the method instead of calling the `as_...` function directly. Other option would be to make the group public. Not strongly against that either. What do you think?

includes/class-wcs-notifications-batch-processor.php
```diff
@@ -230,11 +230,11 @@ public function get_default_batch_size(): int {
 	public static function enqueue(): string {
 		$batch_processor = WCS_Batch_Processing_Controller::instance();
 		if ( $batch_processor->is_enqueued( self::class ) ) {
-			return __( 'Background process for updating subscritpion notifications already started, nothing done.', 'woocommerce-subscriptions' );
+			return __( 'Background process for updating subscription notifications already started, nothing done.', 'woocommerce-subscriptions' );
```
peterfabian, line 233: Well spotted, thanks!

peterfabian (approved):
Looks good, thanks! I'm going to merge this and we can discuss/tweak the remaining points in the main PR.

PR #704, opened 2024-10-21
Add debug tool copy when notifications are disabled

xristos3490:
## Description

This PR adds a more descriptive copy to the debug tool's UI when the notifications are disabled. 

Expected UI:
Screenshot 2024-10-21 at 5 09 21 PM

## How to test this PR

1. Ensure that notifications are disabled.
2. Visit the "WooCommerce > Status" page and navigate to the "Tools" tab. 
3. Notice the new text and the CTA to edit the settings.

## Product impact

### DIFF
--- a/includes/admin/debug-tools/class-wcs-notifications-debug-tool-processor.php
+++ b/includes/admin/debug-tools/class-wcs-notifications-debug-tool-processor.php
@@ -297,7 +297,7 @@ class WCS_Notifications_Debug_Tool_Processor implements WCS_Batch_Processor {
 					'%1$s<br/><strong class="red">%2$s</strong> %3$s <a href="%4$s">%5$s</a>',
 					__( 'This tool will add notifications to pending, active, and on-hold subscriptions. These updates will occur gradually in the background using Action Scheduler.', 'woocommerce-subscriptions' ),
 					__( 'Note:', 'woocommerce-subscriptions' ),
-					__( 'Notifications are currently turned off. To activate them, check the "Enable customer renewal reminder notification emails." option (via Settings > Subscriptions > Customer Notifications).', 'woocommerce-subscriptions' ),
+					__( 'Notifications are currently turned off. To activate them, check the "Enable customer renewal reminder notification emails." option (via WooCommerce > Settings > Subscriptions > Customer Notifications).', 'woocommerce-subscriptions' ),
 					esc_url( admin_url( 'admin.php?page=wc-settings&tab=subscriptions' ) ),
 					__( 'Manage settings.', 'woocommerce-subscriptions' ),
 				),
--- a/includes/admin/debug-tools/class-wcs-notifications-debug-tool-processor.php
+++ b/includes/admin/debug-tools/class-wcs-notifications-debug-tool-processor.php
@@ -290,10 +290,10 @@ class WCS_Notifications_Debug_Tool_Processor implements WCS_Batch_Processor {
 		if ( ! WC_Subscriptions_Email_Notifications::notifications_globally_enabled() ) {
 
 			$tools['start_add_subscription_notifications'] = array(
-				'name'     => __( 'Start adding subscription notifications', 'woocommerce-subscriptions' ),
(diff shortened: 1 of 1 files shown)


peterfabian (commented):
This change looks fine, but since we're looking at this messaging, I noticed that when I enabled the global switch and all the subscriptions were processed quickly by the batch processor that is triggered by the setting update hook, the tool says

> This tool will add notifications to pending, active, and on-hold subscriptions. These updates will occur gradually in the background using Action Scheduler. Currently, there are 29 subscriptions to process.

That, to me, sounds pretty confusing. I would think there are still notifications that are missing, while in reality, everything is fine and only if we decided to run the debug tool, it would process a full set of 29 subscriptions again. 

My suggestion would be to reformulate it to sth like this:

> This tool will regenerate notifications to pending, active, and on-hold subscriptions. These updates will occur gradually in the background using Action Scheduler. 

Then, only if this background processor is enqueued, we can add:
> Currently, there are 29 subscriptions left to process.

What do you think?

xristos3490:
Thanks for the review, @peterfabian! Your suggestion is on point! 

Here's how it looks now:

**Enabled State**
Screenshot 2024-10-22 at 4 53 49 PM

**Enqueued/Running State**
Screenshot 2024-10-22 at 4 53 39 PM

peterfabian (commented):
This looks much better, thanks! Sorry for another nitpick, but when I click on `Add notifications`, the page reloads and still shows `Add notifications` button. Only the next reload will show `Stop adding notifications`

Also, when I click on the `Add notifications` after the reload, it will complain that the "Tool doesn't exist".

Screenshot 2024-10-23 at 11 02 38

Edit: and while we're at it, maybe the buttons should say `Regenerate notifications` and `Stop regenerating notifications`, what do you think?

xristos3490:
Thanks for catching this, @peterfabian! It definitely needs refinement. I've [updated](https://github.com/Automattic/woocommerce-subscriptions-core/pull/704/commits/4b2f4705ed622c3c0232adfdf03639d40e67f61f) the configuration to state that each tool requires a refresh of the list after being clicked. This solves the main issue, and you should immediately see the "Stop" tool.

There is a small catch right now.

The "Tool does not exist" notification still appears if a user refreshes the page right after clicking on this tool. This happens cause the `action` argument is still on the URL. While I could manage this issue using the [`woocommerce_system_status_tool_executed`](https://github.com/woocommerce/woocommerce/blob/ed7d8c850534d7ab81bd7cce2e861c1851047fa4/plugins/woocommerce/includes/admin/class-wc-admin-status.php#L73) hook to perform a redirect, unfortunately, handling the notices becomes challenging, as they get lost after the redirect. 

Moreover, the core hook does not provide the `$response` argument, which would be necessary for a smoother handling process. Thoughts? I think it's OK for now, we could follow up with a PR in core to add that argument.

peterfabian:
I think it behaves fine now. ✅ 
Previously, my test went as follows:

1. Click `Add notifications`, page refreshes
2. Observe the `Add notification` button is there again, so I clicked it again
3. Got `Tool doesn't exist` error.

which was confusing, because it looked like the action wasn't enqueued/executed successfully.

Now it behaves nicely:
1. Click `Add notifications`, page refreshes
2. I see `Stop adding notifications` button and the remaining count of subscriptions that needs to be processed.
3. World makes sense again. :)

Just the last bit from my previous comment remains:

> maybe the buttons should say `Regenerate notifications` and `Stop regenerating notifications` (instead of Add notifications/Stop adding notifications), what do you think?

xristos3490:
Apologies for missing this. I agree, the word "add" doesn't help here. I've [updated](https://github.com/Automattic/woocommerce-subscriptions-core/pull/704/commits/8196b7f0a84ccbe47da6fc422e2026ec3db3fd50) the tool name and the button texts. Here's how it looks now:
Screenshot 2024-10-23 at 2 48 40 PM
Screenshot 2024-10-23 at 2 48 55 PM
Screenshot 2024-10-23 at 2 49 29 PM

peterfabian (approved):
Awesome, thanks for all the updates 🙌

PR #708, opened 2024-10-24
Include the date type when calculating notification offset

xristos3490:
## Description

This pull request adds a date type for calculating the time offset when sending notifications. 

The goal is to allow merchants to define different offsets based on subscription properties and the type of current event (such as renewal, expiration, or end of free trial).

Here’s an example snippet:
```php
add_filter(
	'woocommerce_subscription_customer_notification_time_offset', 
	function( $offset_in_seconds, $subscription, $date_type ) {

		// If it's not a renewal notification, return the default offset.
		if ( 'next_payment' !== $date_type ) {
			return $offset_in_seconds;
		}

		$period   = $subscription->get_billing_period();
		$interval = $subscription->get_billing_interval();

		switch ( $period ) {
			case 'week':
				// Default.
				break;
			case 'month':
				// Default for less than 6 months.
				if ( $interval >= 6 ) {
					$offset_in_seconds = 1 * WEEK_IN_SECONDS;
				}
				break;
			case 'year':
				$offset_in_seconds = 1 * MONTH_IN_SECONDS;
				break;
		}

		return $offset_in_seconds;
	},
	10,
	3
);
```

Additionally, it renames some hooks for consistency.

## How to test this PR

Unit Tests.

## Product impact

### DIFF
--- a/includes/class-wc-subscriptions-email-notifications.php
+++ b/includes/class-wc-subscriptions-email-notifications.php
@@ -184,7 +184,7 @@ class WC_Subscriptions_Email_Notifications {
 	 *
 	 * @return bool
 	 */
-	public static function should_send_notification( $subscription ) {
+	public static function should_send_notification() {
 		$notification_enabled = true;
 
 		if ( WCS_Staging::is_duplicate_site() ) {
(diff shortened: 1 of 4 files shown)


peterfabian (approved):
Looking good, thanks! Always easier to add new filter than to take it away once it's out. 🙌

PR #710, opened 2024-10-28
Clean up and polishing of Subscriptions Notifications

peterfabian:
Fixes https://github.com/woocommerce/woocommerce-subscriptions/issues/4737

## Description

This should be a change that doesn't change the functionality at all. 

The following has changed:
- ~Added admin message after notification wasn't sent---maybe the approach is too butcher-y; Is there a better way to do this?~
  - Added order note when skipping sending the notification via manual trigger
- Improved docs for several classes and methods
- Renamed one method
- Refactored one method with early exits to make it more readable
- Removed accidentally committed PHPCS helper file
- Improved messaging to the user when they execute an action
- Fixed translations of singular/plural strings

## How to test this PR

Mostly just check the unit tests and check the tasks mentioned in the original issue have been fixed.

## Product impact


templates/emails/customer-notification-expiring-subscription.php
```diff
@@ -34,7 +34,12 @@
 		echo wp_kses(
 			sprintf(
 				// translators: %1$s: number of days until expiry, %2$s: date in local format.
-				__( 'Your subscription expires in %1$s days — that’s <strong>%2$s</strong>.', 'woocommerce-subscriptions' ),
+				_n(
+					'Your subscription expires in %1$s day — that’s <strong>%2$s</strong>.',
+					'Your subscription expires in %1$s days — that’s <strong>%2$s</strong>.',
+					(int) $subscription_days_til_event,
+					'woocommerce-subscriptions'
```
james-allan, line 39: Hey @peterfabian do you know if it would be possible to also fix this in the email Subject? 

Screenshot 2024-10-29 at 12 12 39 pm

The problem with the email subject is that it's editable via the email settings so it can't really be variable.

james-allan: I might have a fix for this. I'll submit a draft PR in a minute.

james-allan: I've submitted a draft PR here with a potential solution to this. 
https://github.com/Automattic/woocommerce-subscriptions-core/pull/711

includes/class-wc-subscriptions-email-notifications.php
```diff
@@ -278,7 +281,7 @@ public static function add_settings( $settings ) {
 				'autoload' => false,
 			],
 			[
-				'name'        => __( 'Renewal Reminder Timing', 'woocommerce-subscriptions' ),
+				'name'        => __( 'Renewal Timing', 'woocommerce-subscriptions' ),
```
mattallan: Is this change correct? It looks like you got rid of the wrong word so I wanted to just double-check here 😅 

Should this be "Reminder Timing"?

includes/emails/class-wcs-email-customer-notification.php
```diff
@@ -65,6 +74,31 @@ public function init_form_fields() {
...
+		}
+
+		// Replace 'message=1' with 'message=11' to display correct info to the user.
+		if ( strpos( $redirect_to, 'message=1' ) !== false ) {
+			$redirect_to = remove_query_arg( 'message', $redirect_to );
+			$redirect_to = add_query_arg( 'message', '11', $redirect_to );
+		}
+
+		return $redirect_to;
+	}
```
mattallan: This approach sadly won't work for non-HPOS stores.

Thinking more on this, did we want to add an order note on the subscription for these manual attempts that failed? Similar to the one that we add when the scheduled attempt fails?

Adding some sort of order note for manually triggered actions is common within Subscriptions so it might just make more sense to add a notice and not worry about this "Subscription updated." notice that is shown.

---

If we want to keep exploring other ideas, instead of using this approach with modifying the message param and creating a new messsage type when registering the subscription type, another approach could be to add a notice ourselves and handle the redirect yourself. For example, inside the `trigger()` you could do something like:

```
if ( ! $this->is_enabled()
	|| ! $this->get_recipient()
	|| ! WC_Subscriptions_Email_Notifications::should_send_notification()
	|| WCS_Action_Scheduler_Customer_Notifications::is_subscription_period_too_short( $subscription )
) {
	if ( WC_Subscriptions_Admin::is_edit_subscription_page() ) {
		wcs_add_admin_notice( sprintf( __( 'Attempt to send "%1$s" to %2$s failed.', 'woocommerce-subscriptions' ), $this->title, $this->recipient ), 'error' );
		wp_safe_redirect( $subscription->get_edit_order_url() );
		exit;
	}
	return;
}
```

> [!NOTE]
> `WC_Subscriptions_Admin::is_edit_subscription_page()` is currently a private static function, but we can make this public.

mattallan (commented):
Awesome work @peterfabian! I just pushed up a small commit fixing some typos and left some small feedback 🙌

includes/class-wc-subscriptions-email-notifications.php
peterfabian: Yes, that's deliberate. It's based on this Slack convo: https://a8c.slack.com/archives/C07CBV7SKPH/p1729863747146679

peterfabian: Though looking at it again now, I feel it's ambiguous, since it's the timing of the reminder, not of the renewal. I'll update.

includes/emails/class-wcs-email-customer-notification.php
peterfabian: Thanks for catching that, I wasn't aware. Converted it to an order note in [97b840e](https://github.com/Automattic/woocommerce-subscriptions-core/pull/710/commits/97b840e511be7b7b75b1f80f49e9b951661f7a0e).
Screenshot 2024-11-05 at 12 10 36

includes/emails/class-wcs-email-customer-notification.php
```diff
@@ -80,15 +89,42 @@ public function trigger( $subscription_id ) {
 			|| ! WC_Subscriptions_Email_Notifications::should_send_notification()
 			|| WCS_Action_Scheduler_Customer_Notifications::is_subscription_period_too_short( $subscription )
 		) {
-			// TODO: add admin notice here if in admin
+			// Add order/subscription note.
+			if ( is_admin() ) {
```
mattallan: What are your thoughts on removing this `is_admin()` check?

I was thinking it would be helpful for debugging purposes and for merchants to know whenever a reminder email was scheduled but not sent.

Is the idea that if we ended up inside this condition, the scheduled action should've been deleted but wasn't? And so we want to silently not send the email unless it was requested by the merchant?

includes/emails/class-wcs-email-customer-notification.php
```diff
@@ -114,10 +89,34 @@ public function trigger( $subscription_id ) {
...
+					$order_note_msg .= "\n";
+					$order_note_msg .= __( 'Not a production site.', 'woocommerce-subscriptions' );
+				}
+
+				if ( WCS_Action_Scheduler_Customer_Notifications::is_subscription_period_too_short( $subscription ) ) {
+					$order_note_msg .= "\n";
+					$order_note_msg .= __( 'Subscription billing cycle is too short to send notifications.', 'woocommerce-subscriptions' );
+				}
 			}
+			$subscription->add_order_note( $order_note_msg );
```
mattallan: `$order_note_msg` only exists within the `is_admin()` condition so this could throw an error on non-admin requests.

I changed up this code a little and fixed this up in d84d136749b458cdf25c956c4fa8c149416d04dc

mattallan (approved):
Thanks @peterfabian for fixing that. I like the new approach! I just had a question related to the `is_admin()`: https://github.com/Automattic/woocommerce-subscriptions-core/pull/710/files#r1830347507

I also pushed up d84d136749b458cdf25c956c4fa8c149416d04dc to remove the "\n" breaklines in the order notes and tidy up the logic a bit (hopefully that's okay!)

Since my latest question isn't a blocking one, I'm going to approve this!

Thanks Peter for all your work on this 💯

includes/emails/class-wcs-email-customer-notification.php
peterfabian: Initially, I thought it would make sense to show the notice/info only for manual triggers, but now that you've raised it, perhaps it makes sense also for scheduled notifications. 

There might be some cases when this will generate false positive messages. If the notifications get disabled globally, there might be some notifications that will run until their respective scheduled actions are unscheduled, but it's fairly improbable as we mass-unschedule based on the AS group.

tl;dr is: I agree, let's drop the `is_admin()` check.

includes/emails/class-wcs-email-customer-notification.php
peterfabian: Good catch, that line should have been inside the if.

peterfabian:
Thanks for the thorough check and changes you pushed, @mattallan ! 🙌 

I'm not totally sure about the change in logic, though, as after your changes, only one reason will be stated in the order note while multiple of them might be affecting the decision to skip sending the notification. 

I purposefully added it with separate `if`s to avoid the case where the merchant sees a skipped notification, fixes the reason why it was skipped just to receive another skipped order note afterwards because of another reason. 

What's your take?

includes/emails/class-wcs-email-customer-notification.php
peterfabian: Dropped it in [f7d14a2](https://github.com/Automattic/woocommerce-subscriptions-core/pull/710/commits/f7d14a2a8d83c5ef8963721cf0716f0dcef0473c).

mattallan:
Hey @peterfabian sorry for not getting to this yesterday! I had an A8C team visiting my local city and forgot we planned a lunch with them + other local automatticians 😅

Going back to the latest on this PR... Originally I was thinking that only showing one reason would be fine as I thought the likelihood of multiple reasons would only be small, but also having a list of reasons it was skipped on new lines felt difficult to read.

I just pushed up 8f1cb62 which introduces a slightly different approach to what you had before.

Here's my thinking:
- If a merchant has globally disabled the feature. I don't think they need to worry about all the other reasons the email wasn't sent, so we should just say it was skipped because the feature is off.
- As for the other cases, I've brought back the list of reasons and added `-` to each line (as a bullet point) to try and improve readability.

image

Please feel free to make any further changes!

peterfabian:
Awesome, that looks great! Thanks for adding it!

PR #714, opened 2024-11-04
Prevent adding products to the cart if a subscription renewal is already present

asumaran:
..waiting on discussion https://github.com/Automattic/woocommerce-payments/pull/9655#discussion_r1828282685 

Fixes https://github.com/Automattic/woocommerce-payments/pull/9655

## Description

Initially, the conversion issue from https://github.com/Automattic/woocommerce-payments/issues/9458 was being addressed in https://github.com/Automattic/woocommerce-payments/pull/9655. However, it seems that adding additional products to the cart should not be allowed (to be [confirmed](https://github.com/Automattic/woocommerce-payments/pull/9655#discussion_r1828318622)). Even though it’s currently possible to add extra products to the cart, these products are ignored during checkout, and only the subscription is paid for. This leads me to believe that no additional products should be allowed in the cart if a subscription renewal is present. This PR implements this latter approach.

I am maintaining the same behavior as when a subscription is added and a renewal already exists in the cart: The user is redirected to the product page with an error message. The renewal remains in the cart.

## How to test this PR

> [!NOTE]  
> In order to test this PR with WooPayments you'd need to clone this repo and symlink the repo to your local `subscription-core` directory inside the WooCommerce Subscriptions plugin:
>```
>woopayments/wp-content/plugins/woocommerce-subscriptions » ll vendor/woocommerce
>lrwxr-xr-x 56 2024-11-05 15:28 subscriptions-core -> >/Users/asumaran/Developer/woocommerce-subscriptions-core
>```

Update settings on WooCommerce > Settings > Subscriptions:

- Check “Allow multiple subscriptions and products to be purchased simultaneously.”
- Check “Accept Manual Renewals”

**Create a Subscription**

- Create a simple subscription.
- Go to the subscription page,
- Add the subscription to the cart
- Go to the checkout page and purchase the subscription.
- Navigate to WooCommerce > Subscriptions, and click into this new subscription.
- Use "Order actions" at the top right to "Create pending renewal order" and click the arrow button.
- Navigate to the new order through the link in the Order notes.
- The order will have a Customer payment page link above the Status, click that link.

**Test with subscriptions.**

- Try adding another subscription to the cart.
- ✅ Observe the subscription couldn't be added to the cart and it redirects to the subscription page with the following message "That subscription product can not be added to your cart as it already contains a subscription renewal."

**Test with simple products.**

- Try adding another subscription to the cart.
- ✅ Observe the product couldn't be added to the cart and it redirects to the product page with the following message "That product can not be added to your cart as it already contains a subscription renewal."

## Product impact


mattallan (approved):
Thanks @asumaran for the PR. I left a comment in slack p1731287018910809/1731076731.657249-slack-C055WHLA98D but overall I'm happy for this change to go in as I believe locking down renewal orders from having any product added to the cart is intended.

I just pushed up d155d9a8592b615e241b32a322d77c5dadcd3f36 to deprecate the old function as the name didn't match the logic inside it. I also removed the two notices as I don't think it's important to say "subscription product" for subscription and just "product" for everything else 😅

asumaran:
Thank you @mattallan for improving the PR. I ran a few tests, and everything worked as expected.

PR #716, opened 2024-11-07
Introduce filter for first renewal timestamp.

peterwilsoncc:
## Description

Introduces the filter `woocommerce_subscriptions_synced_first_renewal_payment_timestamp` within the method `WC_Subscriptions_Synchroniser::products_first_renewal_payment_time()`.

This filter allows plugins modifying the start date of syncronised subscriptions to modify the first renewal date too. At present the first renewal date is affected by the current time due to the `self::is_today()` condition. For plugins delaying the start date of a synced subscription, this can result in the first renewal being set before the start date, triggering a fatal error due to the logical fallacy of renewing a subscription before it begins.

Fixes https://github.com/woocommerce/woocommerce-subscriptions/issues/4677

**How can it break?**

Without interaction from other plugins, the code is pretty safe. If another plugin makes use of the filter and incorrectly calculates the first renewal date it will cause problems but I don't think it's a big concern or a situation in which Subs can protect developers from their own mistakes.

## How to test this PR

1. Enable the synced subs feature, not pro-rated, do not charge
2. Create a synced sub product:
   - Simple Subscription
   - Monthly Cycle
   - Renews on the 15th 
3. Add the following as an mu-plugin:
   ```php
   add_filter( 'woocommerce_subscriptions_synced_first_renewal_payment_timestamp', function( $timestamp ) {
   	return $timestamp - DAY_IN_SECONDS;
   } );
   ```
4. Purchase the product
5. Observe the first payment is on the 14th.
6. Delete the mu-plugin created above to avoid confusion later.

## Product impact

- [x] Will this PR affect WooCommerce Subscriptions? yes https://github.com/woocommerce/woocommerce-subscriptions/issues/4677
- [ ] Will this PR affect WooCommerce Payments? no

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Subscriptions Core Changelog ***
 
+= 7.7.0 - xxxx-xx-xx =
+* Dev - Introduce the filter `woocommerce_subscriptions_synced_first_renewal_payment_timestamp` to enable plugins to modify the first renewal date of synced subscriptions.
+
 = 7.6.0 - 2024-10-14 =
 * Fix - Correctly updates a subscription status to `cancelled` during a payment failure call when the current status is `pending-cancel`.
 * Fix - Clear the `cancelled_email_sent` meta when a subscription is reactivated to allow the customer to receive future cancellation emails.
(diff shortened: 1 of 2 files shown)


peterwilsoncc:
cc @mattallan Sorry mate, I can't request a review via the interface so need to do so like this. Let me know if you want me to change the version to allow more time for a review.

mattallan (approved):
Thanks @peterwilsoncc. Changes LGTM so I'm happy to approve and get this merged.

In looking at this again, I saw that our `products_first_renewal_payment_time()` function is hooked onto `woocommerce_subscriptions_product_first_renewal_payment_time` filter with priority 10, so I was wondering if you could've just hooked onto this same filter at a later priority and get the same functionality as using this new hook?

I guess the only difference here is you're returning the `$unmodified_first_renewal_timestamp` which could be handy in determining if this function modified the date.

PR #719, opened 2024-11-08
Replacing get_posts with wc_get_orders

wjrosa:
Fixes 2248-gh-woocommerce/woocommerce-subscriptions

## Description

This PR replaces the usage of the function `get_posts` with `wc_get_orders` to improve performance. To do that, I have updated all the parameters accordingly.

## How to test this PR

Code review. Check if the tests are still passing.

## Product impact

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,7 +1,6 @@
 *** WooCommerce Subscriptions Core Changelog ***
 
 = 7.7.0 - xxxx-xx-xx =
-* Dev: Replaces calls to get_posts with wc_get_orders to improve performance.
 * Dev - Minor refactoring of `init` method in `WC_Subscriptions_Upgrader` class.
 * Dev - Introduce the filter `woocommerce_subscriptions_synced_first_renewal_payment_timestamp` to enable plugins to modify the first renewal date of synced subscriptions.
 
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 *** WooCommerce Subscriptions Core Changelog ***
 
 = 7.7.0 - xxxx-xx-xx =
+* Dev: Replaces calls to get_posts with wc_get_orders to improve performance.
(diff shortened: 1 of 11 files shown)


mattallan (commented):
Hey @wjrosa I'm sorry to say but these functions will no longer be used and will be deprecated once we merge https://github.com/Automattic/woocommerce-subscriptions-core/pull/163 so I don't think we need to update them 😢 Sorry!

I think we can also ignore the `get_posts()` used in `wcs_get_subscription_id_from_key()` as it's an old helper function from back when a subscription object used to be an array of data stored in order meta (stored in meta `{$order_id}_{$product_id}` which we referred to as a subscriptions key) 🙈 

I guess it technically still works and it's not a deprecated helper 🤷 So maybe it's fine for us to update it? Although, just above this `get_posts()` is a direct `posts` table query so we'd need to update that as well.

I'll leave it up to you if you want to update it or close this PR entirely.

If we do decide to leave it, we should probably add an inline comment in case someone stumbles upon this `get_posts()` and has questions xD

wjrosa:
> Hey @wjrosa I'm sorry to say but these functions will no longer be used and will be deprecated once we merge #163 so I don't think we need to update them 😢 Sorry!
> 
> I think we can also ignore the `get_posts()` used in `wcs_get_subscription_id_from_key()` as it's an old helper function from back when a subscription object used to be an array of data stored in order meta (stored in meta `{$order_id}_{$product_id}` which we referred to as a subscriptions key) 🙈
> 
> I guess it technically still works and it's not a deprecated helper 🤷 So maybe it's fine for us to update it? Although, just above this `get_posts()` is a direct `posts` table query so we'd need to update that as well.
> 
> I'll leave it up to you if you want to update it or close this PR entirely.
> 
> If we do decide to leave it, we should probably add an inline comment in case someone stumbles upon this `get_posts()` and has questions xD

No worries! Thanks for letting me know. I have just added the comment above the `get_posts` call instead and removed the changelog entry 👍

mattallan (approved):
Thanks @wjrosa for the latest changes! Approving this :)

PR #722, opened 2024-11-13
In our customer notification emails, only display the subscription that is about to renew or expire, don't show all subscriptions purchased in the parent order.

### LINKED ISSUE #721: The "Subscription Info" table displayed in customer notification emails shows all subscriptions purchased in the parent/initial order
## Describe the bug
Reminder emails currently shows details for all subscriptions included in the original parent/initial order rather than just the one that's about to renew or expire. 

For example, if you purchase a weekly and an every 2nd day subscription in the same order, and then process the "upcoming renewal order" email for the weekly subscription, the "Subscriptions info" table shows both subscriptions.

Image

This is caused by each customer notification template using `\WC_Subscriptions_Order::add_sub_info_email( $order, $sent_to_admin, $plain_text );` where `$order` is the parent order.

### Solution

We should introduce a new function that supports loading the `subscriptions-info.php` template for a single subscription or list of subscriptions, and then use that inside our customer notification templates.

mattallan:
Fixes #721

## Description

While testing the new Customer Notification feature I noticed that in the table under "Subscription Information", it contained all of my subscriptions that I purchased in the parent/initial order.

For example, I purchased a weekly and an "every 2 days" subscription in the same order, and then processed the "upcoming renewal order" email for just the weekly subscription. When I viewed the email it the subscriptions table contained both of my subscriptions which is not correct:

image

We don't really have a function available which displays a single subscription info table, therefore this PR introduces a new function to fill this gap. The new `WC_Subscriptions_Email::subscription_details()` function added in this PR is similar to our existing `WC_Subscriptions_Email::order_details()` function and can be used to load the `subscriptions-info.php` template for single or multiple subscriptions.

## How to test this PR

> [!NOTE]
> To test customer notifications you will need to either:
> 1. Add `return true;` at the top of [`should_send_reminder_email()` ](https://github.com/Automattic/woocommerce-subscriptions-core/blob/f311fe765206ee47b9986bbca2d04dbaccc80889/includes/emails/class-wcs-email-customer-notification.php#L269), or
> 2. Add `define( 'WP_ENVIRONMENT_TYPE', 'production' );` to your `wp-config.php` file
>
> You will also need a way to test receiving an email. For testing locally, I just use the `WP Mail Logging` plugin.

1. Go to **WooCommerce > Settings > Subscriptions** and enable the new Customer Notifications feature
2. Add a weekly and monthly subscription product to your cart and go successfully through the checkout
3. While on `trunk`, go to the Edit Subscription page for the weekly Subscription and use the actions to run "Send upcoming renewal notification".
4. View the email and notice that both subscriptions are listed under "Subscription Information" the upcoming renewal email :x:
5. Switch to this branch and confirm only the relevant subscription is listed in the email

## Product impact

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Subscriptions Core Changelog ***
 
+= 7.7.1 - 2024-xx-xx =
+* Fix - Only show the individual subscription information in customer notification emails, not all subscriptions purchased in the initial order.
+
 = 7.7.0 - 2024-11-13 =
 * Add - New Customer Notification feature: sends reminder emails for upcoming subscription renewals, trials ending, and subscription expirations.
 * Fix - Prevent adding products to the cart if a subscription renewal is already present.
--- a/includes/class-wc-subscriptions-email.php
+++ b/includes/class-wc-subscriptions-email.php
@@ -295,6 +295,32 @@ class WC_Subscriptions_Email {
 	}
 
 	/**
+	 * Show the subscription details table.
+	 *
+	 * @param WC_Subscription $subscription The subscription.
+	 * @param WC_Order|null   $order        The order related to the subscription - defaults to parent order.
+	 * @param bool            $sent_to_admin Whether the email is sent to admin - defaults to false
+	 * @param bool            $plain_text Whether the email should use plain text templates - defaults to false
+	 * @param WC_Email        $email
+	 */
+	public static function subscription_details( $subscriptions, $order = null, $sent_to_admin = false, $plain_text = false, $skip_my_account_link = false ) {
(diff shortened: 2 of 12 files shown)


xristos3490 (approved):
Nice catch, and thanks for the swift PR, @mattallan!

This is testing well on my end! Here's the before and after:

|Before|After|
|---|---|
|Screenshot 2024-11-13 at 4 36 19 PM|Screenshot 2024-11-13 at 4 37 33 PM|

Mayisha (approved):
Thanks for the fix @mattallan. This is working perfectly as described.

**Before**

Screenshot 2024-11-13 at 8 54 36 PM

**After**

Screenshot 2024-11-13 at 8 54 16 PM

mattallan:
Thanks @xristos3490 and @Mayisha !

PR #727, opened 2024-11-18
Correctly style My Account > Subscriptions actions as buttons when using a block theme

mattallan:
Fixes 4748-gh-woocommerce/woocommerce-subscriptions

## Description

Stores using block themes will cause subscription action buttons to be displayed as links instead of the expected buttons. The issue with this is the links are tightly grouped making it confusing to understand that these are actually individual links.

This PR brings back the button styles when switching to a block theme for:
- The list of subscription actions on the single view subscription page (i.e "Cancel", "Change payment" etc)
- The "View" links on the related order table
- The "View" links on the subscriptions list

**View Subscription Actions**

| Before | After |
|--------|--------|
| image | image |

**Related Orders Actions**

| Before | After |
|--------|--------|
| image | image |

**Subscriptions List**

| Before | After |
|--------|--------|
| image| image |

## How to test this PR

1. While on `trunk` enable the `Twenty Twenty Four` theme and visit your My Account Page
2. When viewing My Account subscription pages, notice that the buttons are being displayed as links
3. Checkout this branch and confirm the links are now displayed properly as buttons.
4. Switch to a non-block theme like Storefront and confirm no changes.

## Product impact

### DIFF
--- a/assets/css/view-subscription.css
+++ b/assets/css/view-subscription.css
@@ -7,6 +7,11 @@
 	}
 }
 
+.subscription_details .button {
+	display: inline-block;
+	margin-bottom: 0.5em;
+}
+
 .subscription-auto-renew-toggle {
 	margin-left: 5px;
 	margin-bottom: 2px;
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Subscriptions Core Changelog ***
 
+= 7.8.0 - xxxx-xx-xx =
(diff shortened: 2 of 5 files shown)


Mayisha (approved):
Thanks for the UI update @mattallan. It looks better now 🎉 

In 'My Account > Subscriptions' page-
✅ Buttons are correct in the storefront theme in `trunk` as well as this branch.
✅ Buttons look like links in `twenty twenty four` theme in the `trunk` branch. In this branch the styles are correct and they are rendered as buttons.

Screenshot 2024-11-18 at 12 30 47 PM

PR #718, opened 2024-11-07
Replace all calls to get_post_meta

wjrosa:
Part of 2248-gh-woocommerce/woocommerce-subscriptions
Related https://github.com/Automattic/woocommerce-subscriptions-core/pull/718

## Description

This PR is part of 2248-gh-woocommerce/woocommerce-subscriptions, which suggests some changes towards a code more CRUD-like, and the replacement of some functions to make queries faster. This iteration replaces all `get_post_meta` calls with `get_meta`.

## How to test this PR

Code review. Check if the tests are still passing.

You can also build this branch on the subscriptions extension of your test environment and perform some smoke tests:
- Subscription product creation
- Subscription settings page navigation
- Subscription purchase
- My account subscription details
- Wp-admin subscription details

## Product impact

- [x] Added changelog entry (or does not apply)
- [x] Will this PR affect WooCommerce Subscriptions? yes
- [x] Will this PR affect WooCommerce Payments? yes
- [ ]  Added deprecated functions, hooks or classes to the [spreadsheet](https://docs.google.com/spreadsheets/d/1xw9xszcPMnWsp4C8OKZMsLzZob7tOmWT7qMqmEIq314/edit#gid=0)

PR #729, opened 2024-11-18
Fix fatal error on product page

wjrosa:
Fixes https://github.com/Automattic/woocommerce-subscriptions-core/pull/718#issuecomment-2479441073

## Description

This PR fixes a fatal error thrown on the product page reported by @annemirasol [here](https://github.com/Automattic/woocommerce-subscriptions-core/pull/718#issuecomment-2479441073). Since the list of products can be made of IDs, I will be keeping the original implementation here (before https://github.com/Automattic/woocommerce-subscriptions-core/pull/718)

## How to test this PR

Code review. You can also reproduce the issue following the steps highlighted by Anne [here](https://github.com/Automattic/woocommerce-subscriptions-core/pull/718#issuecomment-2479441073).

## Product impact

cc @annemirasol

### DIFF
--- a/includes/class-wc-subscriptions-product.php
+++ b/includes/class-wc-subscriptions-product.php
@@ -144,7 +144,7 @@ class WC_Subscriptions_Product {
 
 				$child_prices[] = $child_price;
 			} else {
-				$child_prices[] = $child_product->get_price();
+				$child_prices[] = $child_product->get_meta( '_price', true );
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


mattallan (approved):
Thanks @wjrosa, changes LGTM :+1:

Another option would be to add:

```
if ! ( $child_product ) {
    continue;
}
```

after the `wc_get_product()` to confirm we got a valid object back. I'm fine with reverting this change though to avoid the fatal.

wjrosa:
> Thanks @wjrosa, changes LGTM 👍
> 
> Another option would be to add:
> 
> ```
> if ! ( $child_product ) {
>     continue;
> }
> ```
> 
> after the `wc_get_product()` to confirm we got a valid object back. I'm fine with reverting this change though to avoid the fatal.

Nice, Matt! I reverted the method removal and added a similar check:
```
if ( ! $child_product instanceof WC_Product ) {
				continue;
			}
```

PR #733, opened 2024-11-22
Manually clear subscription object from cache after removing order items via My Account > View Subscription page

mattallan:
Fixes 4736-gh-woocommerce/woocommerce-subscriptions

## Description

On sites with caching enabled, there's an issue where after removing an order item from the My Account > View Subscription page, the subscription totals are not being calculated using the latest order item updates.

I wasn't able to reproduce the original issue locally, however after creating a Jurassic Ninja test site with Drop-in Cache Plugins enabled I was able to reproduce:
image

Given this, fixing the issue required clearing the subscription from cache before fetching a new instance of it.

To remove the order item we call `wcs_update_order_item_type()` which already clears the subscription order items cache (i.e. `wp_cache_delete( 'order-items-' . $order_or_subscription_id, 'orders' )` ), but this wasn't enough it seems.

## How to test this PR

1. Create a fresh JN test site with Drop-in Cache Plugins enabled (the default option)
2. Create two subscription products $15/monthly and $20/monthly
3. Purchase these two subscriptions together through the same checkout 
4. On the My Account > View Subscription page, click on the :x: next to the line item to remove it image
6. While on `trunk` you will noticed the line item totals were not updated image
7. While using these changes (will need to build zip and upload it to the JN test site)

## Product impact

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -2,6 +2,7 @@
 
 = 7.8.0 - xxxx-xx-xx =
 * Fix - Use block theme-styled buttons for subscription and related-orders actions on My Account pages.
+* Fix - Subscription totals not properly updating when customers remove items via the My Account > View Subscription page on some stores with caching enabled.
 * Update - Changed the link on the order thank-you page to take customers directly to their "My Account > Subscriptions" page.
 
 = 7.7.1 - 2024-11-13 =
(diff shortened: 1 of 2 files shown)


annemirasol (approved):
LGTM! 🚢 

I was able to repro both the issue and verify the fix!

PR #743, opened 2024-12-10
Change variation's single_add_to_cart_text() 

### LINKED ISSUE #741: WC_Product_Subscription_Variation::single_add_to_cart_text() should not default to add_to_cart_text()
## Describe the bug

A [subscription variation's single add to cart text](https://github.com/Automattic/woocommerce-subscriptions-core/blob/1692c41cf43e2c3c5ef0816a48196ffb241b9e72/includes/class-wc-product-subscription-variation.php#L113)

```
public function single_add_to_cart_text() {
	return apply_filters( 'woocommerce_product_single_add_to_cart_text', self::add_to_cart_text(), $this );
}
```

defaults to [`self::add_to_cart_text()` which runs through the `woocommerce_product_add_to_cart_text`](https://github.com/Automattic/woocommerce-subscriptions-core/blob/1692c41cf43e2c3c5ef0816a48196ffb241b9e72/includes/class-wc-product-subscription-variation.php#L103) filter that is typically associated with the shop loop. 

## To Reproduce

1. Create a variable subscription
2. make one variation - Name Your Price enabled
3. go to WooCommerce > Settings > Name Your Price and set a value for "Add to Cart Button Text for Shop" setting, but leave the "Add to Cart Button Text for Single Product" setting blank.

Image

### Expected behavior
With the "Add to Cart Button Text for Single Product " setting blank, the add to cart button text for the NYP variation should be the same as for the other variations. In the case of a variable sub, it should say "Sign up now"

### Actual behavior

NYP has some trickery to change the button text to the setting value when a variation is NYP-enabled. This value is gathered from the `$product->single_add_to_cart_text()` value and stored in the available variations that are JSON encoded on the `<form>`. 

Name Your Price targets both the `woocommerce_product_add_to_cart_text` and `woocommerce_product_single_add_to_cart_text` filters, but inside those filters there's no way to know if we are in the loop or on the single product page. What is happening with Variable Subscriptions is that the `single_add_to_cart_text()` uses the `add_to_cart_text()` as its default value. And since the latter runs through the typical loop `woocommerce_product_add_to_cart_text` filter it's picking up the "Add to Cart Button Text for Shop" setting that I am modifying from Name Your Price. In this example it's "Choose price" from the settings. On the single product page, this is a weird text.

Image

i think this is fixed if we switch `single_add_to_cart_text()` to:

```
public function single_add_to_cart_text() {
	return apply_filters( 'woocommerce_product_single_add_to_cart_text', WC_Subscriptions_Product::get_add_to_cart_text(), $this );
}
```

## Product impact

helgatheviking:
to use `WC_Subscriptions_Product::get_add_to_cart_text()` as default value. This avoids the `woocommerce_product_add_to_cart_text` filter. Closes #741.

### DIFF
--- a/includes/class-wc-product-subscription-variation.php
+++ b/includes/class-wc-product-subscription-variation.php
@@ -110,7 +110,7 @@ class WC_Product_Subscription_Variation extends WC_Product_Variation {
 	 * @return string
 	 */
 	public function single_add_to_cart_text() {
-		return apply_filters( 'woocommerce_product_single_add_to_cart_text', self::add_to_cart_text(), $this );
+		return apply_filters( 'woocommerce_product_single_add_to_cart_text', WC_Subscriptions_Product::get_add_to_cart_text(), $this );
 	}
 
 	/**


mattallan (approved):
Thanks @helgatheviking for fixing this for us!

I couldn't figure out why our variation product class was using `self::add_to_cart_text()` but our variable and single products were using `WC_Subscriptions_Product::get_add_to_cart_text()` 🤔 

This code hasn't been touched in 9years so I suspect it was an oversight from when it was originally developed.

| `trunk` | this branch |
|--------|--------|
| image | image |

helgatheviking:
Happy to... I caught it while I was fixing an issue in my own dusty code. 😄

PR #744, opened 2024-12-12
Suspend subscription only when failing renewal is most recent

### LINKED ISSUE #3556: Dispute bug? Other orders go to failed
**Describe the bug**
We have an issue, when a customer creates a dispute on one of the renewal payments of a subscription and we lose the dispute, in addition to the affected renewal going to failed (which is expected behaviour), another renewal order goes to failed as well, im not sure why, if its because it has same amount or not, but we see this every time that this exact thing happens.

The only thing in the note is:
Order status changed from Processing to Failed.
October 28, 2024 at 5:59 pm [Delete note]

**To Reproduce**
Have a subscription with multiple renewals,
Dispute one of the renewals (not the latest one for example).
Merchant loses the renewal.

**Expected behavior**
The specific order should go to refunded/failed , not any other orders.

**Screenshots**

**Environment (please complete the following information):**
 - WordPress Version - latest
 - WooCommerce Version - latest
 - Stripe Plugin Version - latest (8.8.1)
 - Woocommerce subscriptions - latest

annemirasol:
Fixes https://github.com/woocommerce/woocommerce-gateway-stripe/issues/3556

## Description
In https://github.com/woocommerce/woocommerce-gateway-stripe/issues/3556, when the merchant loses a dispute for one of a subscription's renewals, the disputed renewal order is set to `failed`, the subscription is set to `on-hold`, and the most recent renewal order, which may or may not be the same as the disputed order, is also set to `failed`.

This PR changes the behavior such that we only want to update the status of the affected renewal order, and only put the subscription on hold if the failing renewal is the most recent renewal.

This is similar to #625, which fixes successful renewals and subscription activation.

## How to test this PR
We want to simulate a case where a subscription renewal (that is not the most recent renewal) is disputed and the merchant lost the dispute.

1. Make sure your Stripe webhooks are working properly.
2. Since we are simulating a dispute using a special CC number, the dispute webhook might fire unrealistically too soon, before the charge info is associated with the order. Add a `sleep(5);` on the first line of [`WC_Stripe_Webhook_Handler::process_webhook_dispute()`](https://github.com/woocommerce/woocommerce-gateway-stripe/blob/b919ed35db707a0c9396ce29e07e570dc3f0b11c/includes/class-wc-stripe-webhook-handler.php#L337).
3. Purchase a subscription using `4242424242424242`.
4. In My Account > Subscriptions, verify the subscription is set to auto-renew as expected, and change the payment method to `4000000000000259`, a special dispute CC number.
5. In WooCommerce > Subscriptions > Edit Subscription, process a renewal for the subscription.
6. Wait for a few seconds and verify the dispute kicked in, and the renewal order has been put on hold.
7. In My Account > Subscriptions, change the payment method back to `4242424242424242`.
8. **Refresh the Edit Subscription page to load the updated payment method (important)**,  and then process a couple of renewals for the subscription.
   - These renewals should succeed, with no dispute.
9. In your Stripe dashboard, accept the dispute for the renewal in Step 5. (There should only be one dispute.)
10. In `develop`, the disputed order and the most recent order gets set to `failed`, and the subscription itself is set to `on-hold`.
11. In this branch, only the disputed order gets set to `failed`, and the subscription remains active.

| Before | After |
|--------|-------|
| Screenshot 2024-12-13 at 10 54 08 AM | Screenshot 2024-12-13 at 10 53 45 AM |

## Product impact

- [ ] Will this PR affect WooCommerce Subscriptions? yes
- [ ] Will this PR affect WooCommerce Payments? no

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,6 +1,7 @@
 *** WooCommerce Subscriptions Core Changelog ***
 
 = 7.8.0 - xxxx-xx-xx =
+* Fix - Prevent failing non-recent renewal order from suspending the subscription and marking the most recent renewal order as failed.
 * Fix - Use block theme-styled buttons for subscription and related-orders actions on My Account pages.
 * Fix - Subscription totals not properly updating when customers remove items via the My Account > View Subscription page on some stores with caching enabled.
 * Fix - Resolved unexpected errors during the renewal process when a subscription contains metadata with key "id".
--- a/includes/class-wc-subscriptions-renewal-order.php
+++ b/includes/class-wc-subscriptions-renewal-order.php
@@ -139,7 +139,8 @@ class WC_Subscriptions_Renewal_Order {
 				if ( $is_failed_renewal_order ) {
 					do_action( 'woocommerce_subscriptions_paid_for_failed_renewal_order', wc_get_order( $order_id ), $subscription );
 				}
-			} elseif ( 'failed' == $orders_new_status ) {
+			} elseif ( 'failed' === $orders_new_status && wcs_is_order_last_renewal_of_subscription( $order, $subscription ) ) {
+				// We will suspend the subscription if the latest renewal order fails.
 				$subscription->payment_failed();
 			}
 		}


mattallan (approved):
I'm honestly surprised we haven't seen any reports of this issue 🤔 I had a quick look through open/closed issues and couldn't find this one so I guess stores don't ever mark older renewal orders as failed once they've been processed?

Anyways, thanks @annemirasol for fixing this!
I've tested:
- [x] Confirmed while on `develop`. Marking an old renewal as failed, updated the subscription status to `on-hold` and also updated the latest renewal.
- [x] Confirmed when last renewal order failed, the subscription is put on hold.
- [x] Confirmed when an old renewal is marked as failed, the subscriptions status remains.

annemirasol:
The failing checks are due to `ubuntu-latest` no longer having `subversion` pre-installed. This will be fixed in https://github.com/Automattic/woocommerce-subscriptions-core/pull/748.

PR #746, opened 2024-12-13
Do not unset package contents for Shipping Subscriptions with free trials

### LINKED ISSUE #9771: ECE buttons not rendered for variable subscription products on the product details page
### Describe the bug
ECE buttons not rendered for variable subscription products on the product details page.

### To Reproduce
1. Create a variable subscription product, virtual type, with a non zero product price.
2. Shop the product
3. Note that on the Product Details Page the buttons are absent while present on the cart and checkout.

### Actual behavior
- For non $0 products, the express checkout buttons are rendered.
- If the variable selection is not made, as is done for a variable product, an alert is displayed for the shopper to first select the variable option.

### Screenshots
Image

Image

asumaran:
Fixes https://github.com/Automattic/woocommerce-payments/issues/9771

## Description

Enables shipping options for variable subscriptions with free trial.

## How to test this PR

To be tested along with https://github.com/Automattic/woocommerce-payments/pull/9955

### DIFF
--- a/includes/class-wc-subscriptions-cart.php
+++ b/includes/class-wc-subscriptions-cart.php
@@ -650,12 +650,6 @@ class WC_Subscriptions_Cart {
 
 		if ( 'none' === self::$calculation_type ) {
 			foreach ( $packages as $index => $package ) {
-				foreach ( $package['contents'] as $cart_item_key => $cart_item ) {
-					if ( WC_Subscriptions_Product::get_trial_length( $cart_item['data'] ) > 0 ) {
-						unset( $packages[ $index ]['contents'][ $cart_item_key ] );
-					}
-				}
-
 				if ( empty( $packages[ $index ]['contents'] ) ) {
 					unset( $packages[ $index ] );
 				}
--- a/includes/class-wc-subscriptions-cart.php
+++ b/includes/class-wc-subscriptions-cart.php
@@ -650,6 +650,12 @@ class WC_Subscriptions_Cart {
 
 		if ( 'none' === self::$calculation_type ) {
 			foreach ( $packages as $index => $package ) {
+				foreach ( $package['contents'] as $cart_item_key => $cart_item ) {
+					if ( WC_Subscriptions_Product::get_trial_length( $cart_item['data'] ) > 0 && apply_filters( 'wcs_unset_package_contents', '__return_true' ) ) {
+						unset( $packages[ $index ]['contents'][ $cart_item_key ] );
+					}
+				}
+
 				if ( empty( $packages[ $index ]['contents'] ) ) {
 					unset( $packages[ $index ] );
 				}


mattallan:
Hey @asumaran thanks for opening this PR and for all your investigation into this issue on WooPayments.

I had to go through the history books to find our why we unset shipping packages and it looks like it's always been this way (since we introduced multiple subscriptions per checkout in v2.0).

It also appears to be the intended behaviour as per [our FAQ documentation](https://woocommerce.com/document/subscriptions/faq/#section-3), which states:

> If your subscription includes a free trial, shipping is free for that first period if no other products are purchased in the same transaction. To charge an up-front shipping fee for a subscription with a free trial, you can include a signup fee for the cost of shipping.

---

There's a few things changing with removing this block that I wanted to call out:

1. Parent orders of a trial subscription will now have a shipping line item

| `develop` | This branch |
|--------|--------|
| image | image | 

> [!WARNING]
> This PR introduces a bug when you go through the block checkout. The order summary on the checkout page shows that flat rate shipping is "FREE" but then when you purchase the order, it's actually charged $10 for shipping.

Ignoring the bug above. I'm not overly opposed to this change overall but it does change the fundamentals for how we've handled shipping on the initial order of trial subscriptions.

I don't know why merchants would ship a free trial, but if it does need shipping, then there should probably be a shipping line item. This feels like it should be a whole setting on its own i.e. ask merchants if the trial period requires shipping.

Our current approach of telling merchants to include a signup fee to record shipping costs feels very outdated and I'm assuming this wouldn't work well with shipping extensions that use shipping line items to do things like attach tracking numbers etc.

2. Customers can now choose a shipping option on the checkout page for their initial trial product and a different shipping option for their recurring subscription.

| `develop` | This branch |
|--------|--------|
| image |image |

I think this will be confusing to customers and could potentially catch merchants by surprise. I don't think a merchant would want customers picking an Express shipping method for their trial product (which will be set to $0) and then some cheaper option for the recurring orders.

3. The shipping cost is set to 0 for free-trials regardless of which shipping option is selected. i.e. Express post and free shipping will both be set to $0.

| `develop` | This branch |
|--------|--------|
| image |image |

Similar to the above point. Regardless of the shipping method chosen, if the parent order is for a free trial, we set the cost to $0. Not sure if merchants may be confused seeing a new order come in with a shipping line item for Flat rate but no shipping was charged.

---

There's another change coming in this PR that I would consider a bug/break.

1. Synced subscriptions that are purchased before the synced date shouldn't have a shipping line item or prompting the customer to choose a shipping option for the initial purchase.

Say a store sells a subscription product that's synced to the 1st of the month. We apply a mock trial period on the cart item for the period between purchase date and the synced payment date. When a customer purchased this product on the 15th, this order doesn't need shipping and shouldn't have a shipping line item.

asumaran:
@mattallan, thank you for sharing your thorough feedback on my changes in this PR. I was uncertain about the simplicity and isolation of my changes. I’ll address each of your points but first, I want to focus on getting the WooPayments PR ready, and then I’ll shift my focus to this side of the issue.

asumaran:
Closing this PR since Tokenized Carts is now the default implementation in WooPayments. This allows us to get the data we need more accurately.

What we were previously doing here felt pretty _hacky_ to prevent the removal of shipping options from the package when a subscription included free trial and a sign-up fee (by using a filter controlled from WooPayments). To be honest, I wasn’t very happy with that solution.

However, with Tokenized Carts (which uses the Store API), the shipping options information is available in `extensions.subscriptions` of the Store API response making this PR no longer necessary. 

Closing this in favor of https://github.com/Automattic/woocommerce-payments/pull/10099

PR #751, opened 2024-12-18
Add support for signup fee coupons during subscription switches

### LINKED ISSUE #726: Sign-up fee discount not applied when switching subscription
## Describe the bug
Reported by @allie500 in p1731683240284119-slack-C055WHLA98D: 

A merchant has reported an issue with applying a sign up % discount coupon when a shopper with an active subscription of a variable subscription product switches to a new variation with a sign up fee. The coupon percentage is not applied to the sign up fee. 

The HE who escalated the issue to Woo-Devs was able to replicate the issue. I spun up a JN site and was also able to replicate the issue. I also didn't see anything relevant in the debug.log file nor the wc-logs.

When I applied the coupon no exception or errors were thrown. The coupon was showing up on the page but the percentage amount was not deducted from the sign up fee. 

Processing the order resulted in the full sign up fee being charged. I checked the coupons page in the wp-admin and it shows that the coupon has one usage.

## To Reproduce

1. Go to the coupon page in the wp-admin and make note of the current usage number for the coupon.
2. Buy the annual variation of the subscription.
3. Confirm the order was successful.
4. Then on the my account page go to the subscriptions subpage and click the upgrade or downgrade button.
5. Select the lifetime variation.
6. Apply the coupon: lifetime20
7. Observe that the amount due is not updated to reflect a 20% discount on the sign up price.
8. Observe that the coupon is successfully added to the order.
9. Finish checkout.
10. Go to the coupons page in the wp-admin and confirm that the number incremented.

### Expected behavior
Sign up fee coupons get applied when upgrading/downgrading subscriptions.

### Actual behavior
Sign up fee coupons get ignored when upgrading/downgrading subscriptions.

## Product impact

## Additional context

Initial investigation by @annemirasol traced the behavior to https://github.com/Automattic/woocommerce-subscriptions-core/blob/07bf070a5b2c9716bb20280055c4f3f07d83faed/includes/class-wc-subscriptions-coupon.php#L223, which seemingly skips "switches" on purpose when computing sign-up fee discounts.

annemirasol:
Fixes #726 

## Description
This PR adds support for signup fee coupon during subscription switches. It computes for the actual signup fee amount and applies the discount to that amount.

## How to test this PR
1. Go to WooCommerce > Settings > Subscriptions > Switching, and
   - Enable `Allow Switching: Between Subscription Variations`.
   - For Prorate Recurring Payment: `For Upgrades and Downgrades of All Subscription Products`
   - For Prorate Signup Fee: `Never (charge the full signup fee)`
3. Go to Marketing > Coupons and create a coupon: a signup fee % discount for 20%.
4. Create a Variable Subscription product:
   - Under Attributes, add "Membership Tier" with "Regular|Premium" as values, and check `Used for variations` before saving.
   - Under Variations, create:
      - Regular: 80.00/month, with 5.00 signup fee
      - Premium: 100.00/month, with 5.00 signup fee
5. As a shopper, purchase the "Regular" variation.
6. Go to My Account > Subscriptions > View, and click "Upgrade or Downgrade". Upgrade to "Premium" and go to checkout.
7. Add the signup fee % discount coupon.
8. Verify that the "Order Summary Signup Fee" shows 25.00, but the discount amount is only 1.00 (20% of 5.00, the signup fee prior to upgrade costs).
9. Test for different combinations of `Prorate Recurring Payment` and `Prorate Signup Fee`: no matter the combination, the discount amount should never be more than 1.00.
10. Test using different coupon types:
    - signup fee coupon, e.g. 10.00: discount should always be 5.00 or less
    - recurring fee/recurring fee percent coupon: discount should be applied to upgrade cost excluding signup fee. 
11. Test a subscription downgrade.

## Product impact

### DIFF
--- a/includes/class-wc-subscriptions-coupon.php
+++ b/includes/class-wc-subscriptions-coupon.php
@@ -219,10 +219,14 @@ class WC_Subscriptions_Coupon {
 				}
 			}
 
-			// Compute the sign-up fee. If it's a switch, we need to get the signup fee less
-			// recurring payment upgrade/downgrade costs.
+			// Compute the true sign-up fee. If it's a subscription upgrade, we need to get the fee
+			// before extra charges, e.g. prorated recurring payment, were applied.
 			if ( $is_switch ) {
-				$sign_up_fee = (int) $cart_item['data']->get_meta( '_subscription_sign_up_fee_prorated' );
+				$is_downgrade = isset( $cart_item['subscription_switch']['upgraded_or_downgraded'] ) &&
+										'downgraded' === $cart_item['subscription_switch']['upgraded_or_downgraded'];
+				$sign_up_fee  = $is_downgrade ?
+										WC_Subscriptions_Product::get_sign_up_fee( $cart_item['data'] ) :
+										(int) $cart_item['data']->get_meta( '_subscription_sign_up_fee_prorated' );
 			} else {
 				$sign_up_fee = WC_Subscriptions_Product::get_sign_up_fee( $cart_item['data'] );
 			}
--- a/includes/class-wc-subscriptions-coupon.php
+++ b/includes/class-wc-subscriptions-coupon.php
@@ -222,10 +222,10 @@ class WC_Subscriptions_Coupon {
 			// Compute the original sign-up fee. If it's a switch, we need to get the signup fee less
 			// upgrade costs.
 			if ( $is_switch ) {
-				$sign_up_fee_prorated = (int) $cart_item['data']->get_meta( '_subscription_sign_up_fee_prorated' );
-				$price_prorated       = (int) $cart_item['data']->get_meta( '_subscription_price_prorated' );
+				$sign_up_fee_prorated = (float) $cart_item['data']->get_meta( '_subscription_sign_up_fee_prorated' );
+				$price_prorated       = (float) $cart_item['data']->get_meta( '_subscription_price_prorated' );
 
-				if ( 0 === $sign_up_fee_prorated && 0 === $price_prorated ) {
+				if ( 0.0 === $sign_up_fee_prorated && 0.0 === $price_prorated ) {
 					// No prorated recurring fees, i.e. no extra upgrade costs, so we can use the original sign-up fee.
 					$sign_up_fee = WC_Subscriptions_Product::get_sign_up_fee( $cart_item['data'] );
 				} else {
--- a/includes/class-wc-subscriptions-coupon.php
+++ b/includes/class-wc-subscriptions-coupon.php
@@ -219,10 +219,18 @@ class WC_Subscriptions_Coupon {
 				}
 			}
 
-			// Compute the sign-up fee. If it's a switch, we need to get the signup fee less
-			// recurring payment upgrade/downgrade costs.
+			// Compute the original sign-up fee. If it's a switch, we need to get the signup fee less
+			// upgrade costs.
 			if ( $is_switch ) {
-				$sign_up_fee = (int) $cart_item['data']->get_meta( '_subscription_sign_up_fee_prorated' );
+				$sign_up_fee_prorated = (int) $cart_item['data']->get_meta( '_subscription_sign_up_fee_prorated' );
+				$price_prorated       = (int) $cart_item['data']->get_meta( '_subscription_price_prorated' );
+
+				if ( 0 === $sign_up_fee_prorated && 0 === $price_prorated ) {
+					// No prorated recurring fees, i.e. no extra upgrade costs, so we can use the original sign-up fee.
+					$sign_up_fee = WC_Subscriptions_Product::get_sign_up_fee( $cart_item['data'] );
+				} else {
+					$sign_up_fee = $sign_up_fee_prorated;
+				}
 			} else {
 				$sign_up_fee = WC_Subscriptions_Product::get_sign_up_fee( $cart_item['data'] );
 			}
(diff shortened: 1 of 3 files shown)


annemirasol:
@mattallan Thank you again for the pointers. Added you as reviewer, for extra 👀

includes/class-wc-subscriptions-coupon.php
```diff
@@ -219,8 +219,24 @@ public static function get_discount_amount_for_cart_item( $cart_item, $discount,
...
 			}
 
-			// Apply sign-up discounts. Exclude switch cart items because their initial amount is entirely sign-up fees but should be treated as initial amounts
-			if ( ! $is_switch && WC_Subscriptions_Product::get_sign_up_fee( $cart_item['data'] ) > 0 ) {
+			// Compute the original sign-up fee. If it's a switch, we need to get the signup fee less
+			// upgrade costs.
+			if ( $is_switch ) {
+				$sign_up_fee_prorated = (int) $cart_item['data']->get_meta( '_subscription_sign_up_fee_prorated' );
+				$price_prorated       = (int) $cart_item['data']->get_meta( '_subscription_price_prorated' );
+
```
Mayisha, line 229: How about we cast it to `float` in case the prorated fee is a fraction?

includes/class-wc-subscriptions-coupon.php
```diff
@@ -219,8 +219,24 @@ public static function get_discount_amount_for_cart_item( $cart_item, $discount,
...
+				if ( 0 === $sign_up_fee_prorated && 0 === $price_prorated ) {
```
Mayisha: Quick question for my understanding - why the `price_prorated` is also being checked here? 🤔

Mayisha (commented):
I have come across a confusing scenario.

- I created 2 coupons
    - a signup fee % discount for 20% (signup percent)
    - a signup fee discount for 10 (signup 10)
- I created a variable subscription 'Membership Tier' as mentioned in the PR description but with different signup fees.
    - Regular with a signup fee 5.
    - Premium with a signup fee 10.

**Case 1:**
- Purchased the 'Regular' variation.
- Upgraded to 'Premium' and went to the checkout page.
- The signup fee is 30 (not 10) 🤔 
- Used the `signup percent` coupon. The amount is -2. (This is the correct percentage for signup fee 10 but why does it show 30 here? )
- Used the `signup 10` coupon. The amount is -10. (also correct caps for signup fee 10)
- In this case the sign up fee is off. Also with multiple coupons used, the discount is more than the sign up fee.

Screenshot 2025-01-09 at 2 02 35 AM


**Case 2:**
- Purchased the 'Premium' variation.
- Downgraded to 'Regular' and went to the checkout page.
- Used the `signup 10` coupon. The amount is -5. (also correct)
- Used the `signup percent` coupon. The amount is 0. (If I remove the previous coupon the discount becomes -1 which is correct)
- In this case, is it the expected behavior for multiple coupons.

Screenshot 2025-01-09 at 1 58 53 AM

includes/class-wc-subscriptions-coupon.php
annemirasol: Good question! In https://github.com/Automattic/woocommerce-subscriptions-core/pull/751/commits/3ee2abcea61ee0e03d133f591dc21d711323075a, I simplified the logic using `meta_exists`. Hopefully it's more readable.

annemirasol:
@Mayisha 

> Case 1:

> Purchased the 'Regular' variation.
> Upgraded to 'Premium' and went to the checkout page.
> The signup fee is 30 (not 10) 🤔

Did you set recurring fees to be prorated, and signup to be charged full amount? If that is the case, the signup fee _displayed_ is 30 because it is 10 (the true signup fee) + 20 (the difference between Regular and Premium).

The signup fee being 30.00 is part of the issue why signup fee coupons are currently not supported. In [this function](https://github.com/woocommerce/woocommerce-subscriptions/blob/5d69ad00fc801743fefc37cbeebaaf44105e8d4d/includes/switching/class-wcs-switch-totals-calculator.php#L501), we overwrite the true signup fee with (true signup fee + upgrade costs from the recurring fee difference).

> Used the signup percent coupon. The amount is -2. (This is the correct percentage for signup fee 10 but why does it show 30 here? )
> Used the signup 10 coupon. The amount is -10. (also correct caps for signup fee 10)
> In this case the sign up fee is off. Also with multiple coupons used, the discount is more than the sign up fee.

Good catch! I need to add logic for when multiple coupons are used.

Mayisha:
> Did you set recurring fees to be prorated, and signup to be charged full amount? If that is the case, the signup fee displayed is 30 because it is 10 (the true signup fee) + 20 (the difference between Regular and Premium).

You are right. I have these settings-

Screenshot 2025-01-09 at 11 21 52 AM

annemirasol:
> > Used the signup percent coupon. The amount is -2. (This is the correct percentage for signup fee 10 but why does it show 30 here? )
Used the signup 10 coupon. The amount is -10. (also correct caps for signup fee 10)
In this case the sign up fee is off. Also with multiple coupons used, the discount is more than the sign up fee.
>
> Good catch! I need to add logic for when multiple coupons are used.


I checked the sign up fee coupon behavior for initial subscription purchases, and the behavior seems to be the same: multiple sign up fee coupons applied to the same cart can lead to a total discount that is greater than the actual sign up fee.

Screenshot 2025-01-10 at 11 02 45 AM

@mattallan Is this a known issue, or a non-issue/expected behavior?

barryhughes:

PR #756, opened 2025-01-02
[WooCommerce 9.6] Preview Email support for subscription emails

mattallan:
Fixes 4760-gh-woocommerce/woocommerce-subscriptions

> [!NOTE]
> You will need to checkout `checking_can_renew_early_invalid_subscription` on the woocommerce-subscriptions repo (see 4770-gh-woocommerce/woocommerce-subscriptions) before testing this branch.

## Description

This PR adds Subscriptions support for the new Preview Email feature coming in WooCommerce 9.6 (releasing this month). Since originally opening the issue, some of the issues I found have already been addressed in WooCommerce core and so there were really only two key issues that needed to be fixed:
- When previewing the Switching emails, the `WCS_Email_New_Switch_Order` and `WCS_Email_Completed_Switch_Order` emails expected `$email->subscriptions` to be an array of subscriptions [see here](https://github.com/Automattic/woocommerce-subscriptions-core/blob/13783514ffb9833f93b38355a8d7b5cbc8cff89a/includes/emails/class-wcs-email-completed-switch-order.php#L141).
- When previewing the majority of our subscription emails, `$email->object` was expected to be an instance of `WC_Subscription`, not a `WC_Order`

image

While working on these changes I found a couple of small issues that I've also addressed in this PR since they impacted testing. These include:
- 2050be8a5485246e95151fb88e1d9fe78cc2f001 - Store managers can manually trigger the customer notification email from Admin Edit Subscription page, so we should be calling `wcs_can_user_renew_early()` with the subscription's customer ID, not the current logged in user.
- cd24deb168c384f6d4c1453310a0b7e9c0b07e42 - Undefined `$subscription` variable called in our `subscription_details()` function.

## How to test this PR

1. Download the latest [WooCommerce core nightly build or download the 9.6 beta/RC](https://github.com/woocommerce/woocommerce/releases)
2. While on `trunk` of `subscriptions-core`, go to **WooCommerce > Settings > Emails**
3. Under the table of emails, you should see the new email improvements, including a preview of the email.
4. From the dropdown, attempt to preview the "Subscription Cancelled" email and you should see some errors.

| Subscription cancelled | Switch Completed | Customer Notification |
|--------|--------|--------|
| image | image | image |

5. Checkout this branch and confirm all subscription emails are previewing correctly without any errors

## Product impact

### DIFF
--- a/includes/class-wc-subscriptions-email.php
+++ b/includes/class-wc-subscriptions-email.php
@@ -12,6 +12,24 @@
 class WC_Subscriptions_Email {
 
 	/**
+	 * List of all core subscription email classes.
+	 *
+	 * @var array
+	 */
+	public static $email_classes = [
+		'WCS_Email_New_Renewal_Order'              => true,
+		'WCS_Email_New_Switch_Order'               => true,
+		'WCS_Email_Processing_Renewal_Order'       => true,
+		'WCS_Email_Completed_Renewal_Order'        => true,
+		'WCS_Email_Customer_On_Hold_Renewal_Order' => true,
+		'WCS_Email_Completed_Switch_Order'         => true,
+		'WCS_Email_Customer_Renewal_Invoice'       => true,
+		'WCS_Email_Cancelled_Subscription'         => true,
+		'WCS_Email_Expired_Subscription'           => true,
+		'WCS_Email_On_Hold_Subscription'           => true,
+	];
+
+	/**
 	 * Bootstraps the class and hooks required actions & filters.
 	 *
 	 * @since 1.0.0 - Migrated from WooCommerce Subscriptions v1.0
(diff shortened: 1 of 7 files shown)


includes/class-wc-subscriptions-email.php
```diff
@@ -11,6 +11,24 @@
  */
 class WC_Subscriptions_Email {
 
+	/**
+	 * List of all core subscription email classes.
+	 *
+	 * @var array
+	 */
+	public static $email_classes = [
```
mattallan, line 19: I moved the email classes into a list like this so that I could easily create a function to check if the email being previewed is a subscription email (see `is_subscription_email()`)

annemirasol (changes requested):
Great job being on top of upcoming features ✨  This looks great!

I see some PHP notices in the preview and logs for `Cancelled Subscription`, `Expired Subscription` and `Suspended Subscription`.


Screenshot 2025-01-03 at 4 09 49 PM

```
Notice: Function customer_id was called incorrectly. Order properties should not be accessed directly. Backtrace: require_once('wp-admin/admin.php'), do_action('admin_init'), WP_Hook->do_action, WP_Hook->apply_filters, WC_Admin->preview_emails, Automattic\WooCommerce\Internal\Admin\EmailPreview\EmailPreview->render, Automattic\WooCommerce\Internal\Admin\EmailPreview\EmailPreview->render_preview_email, WCS_Email_On_Hold_Subscription->get_content_html, wc_get_template_html, wc_get_template, include('/plugins/woocommerce-subscriptions-core/templates/emails/on-hold-subscription.php'), WC_Subscription->get_time, WC_Subscription->get_date, apply_filters('woocommerce_subscription_get_last_order_date_created_date'), WP_Hook->apply_filters, WC_Subscriptions_Email_Preview->mock_last_order_date_created, WC_Subscription->__get, WC_Abstract_Legacy_Order->__get, wc_doing_it_wrong Please see [in /var/www/html/wp-includes/functions.php on line 6114 January 3, 2025](https://developer.wordpress.org/advanced-administration/debug/debug-wordpress/)
```

mattallan:
Gah! Thanks @annemirasol for catching this! I forgot I flicked `WP_Debug` off on my test store so there were a few more issues I didn't catch 🙈 

Latest fixes:
- [f70e96b](https://github.com/Automattic/woocommerce-subscriptions-core/pull/756/commits/f70e96b0c3037b4e00c04261dd753a291c816d76) - use proper getter `$subscription->get_customer_id()` instead of accessing the prop directly.
- [1ecd7](https://github.com/Automattic/woocommerce-subscriptions-core/pull/756/commits/761ecd754287b2bb0dfd5c716c9775d31482bc69) - while previewing emails, we use a dummy orders/subscription object that isn't saved to the database. This means calling `wc_get_order( $dummy_object_id )` won't work which causes our `wcs_get_edit_post_link()` to return null inside preview emails which is then passed to `esc_url()` which throws the error.

includes/class-wc-subscriptions-email.php
```diff
@@ -306,11 +317,17 @@ public static function order_details( $order, $sent_to_admin = false, $plain_tex
...
+		}
+
+		$order = ! $order && ! empty( $subscriptions ) ? reset( $subscriptions )->get_parent() : $order;
+
 		wc_get_template(
 			$template,
 			array(
-				'order'                => ! $order ? $subscription->get_parent() : $order,
-				'subscriptions'        => is_array( $subscriptions ) ? $subscriptions : [ $subscriptions ],
+				'order'                => $order ?? null,
```
annemirasol: Isn't `$order ?? null` just `$order`?

annemirasol (approved):
Email previews look good, code LGTM! 🚢

includes/class-wc-subscriptions-email.php
mattallan: Yes, yes it is. Don't remember what I was thinking when I first did this 🤔 Thanks for catching this! Fixed in [2e83a79](https://github.com/Automattic/woocommerce-subscriptions-core/pull/756/commits/2e83a79ed321c149fab5ea5fdb59d2c776002b39)

PR #759, opened 2025-01-08
Add safeguards to guard against unexpected types (in the subscriptions totals template)

### LINKED ISSUE #750: Fatal error can occur in subscription-totals-table.php
## Describe the bug

In `subscription-totals-table.php` we create a product object [here](https://github.com/Automattic/woocommerce-subscriptions-core/blob/14e7b3d5df82a0f4d4942b8119e85087e4863a65/templates/myaccount/subscription-totals-table.php#L27)

The is filtered by `woocommerce_subscriptions_order_item_product`, and we check the status of the object before using it [here](https://github.com/Automattic/woocommerce-subscriptions-core/blob/14e7b3d5df82a0f4d4942b8119e85087e4863a65/templates/myaccount/subscription-totals-table.php#L28-L77)

However, it is subsequently used without a check [here](https://github.com/Automattic/woocommerce-subscriptions-core/blob/14e7b3d5df82a0f4d4942b8119e85087e4863a65/templates/myaccount/subscription-totals-table.php#L79)

On rare occasions, where the value has been filtered, such as in 9093977-zd-a8c, this can result in a fatal error.

```
Uncaught Error: Call to a member function get_purchase_note() on bool in /www/[redacted]/public/wp-content/plugins/woocommerce-subscriptions/vendor/woocommerce/subscriptions-core/templates/myaccount/subscription-totals-table.php:79  
```

## To Reproduce

1.  Return `false` on `woocommerce_subscriptions_order_item_product`
2. Check the view subscription page in My Account 

### Actual behavior

Improper use of the filter can have unexpected results.

## Product impact

- [ yes] Does this issue affect WooCommerce Subscriptions? yes/no/tbc, add issue ref
- [ no] Does this issue affect WooCommerce Payments? yes/no/tbc, add issue ref

barryhughes:
There's a great summary of the problem in the [linked issue](https://github.com/Automattic/woocommerce-subscriptions-core/issues/750) but, in brief, the problem is that (within the [subscription totals template](https://github.com/Automattic/woocommerce-subscriptions-core/blob/14e7b3d5df82a0f4d4942b8119e85087e4863a65/templates/myaccount/subscription-totals-table.php)) we support filtering of product objects:

```php
$_product = apply_filters( 'woocommerce_subscriptions_order_item_product', $item->get_product(), $item );
```

Of course, nothing really stops a filter function from changing the filtered thing into some other type. Indeed, based on Nic's note in the original issue, it sounds like there has been at least one real-world report of this in which `$_product` is set to `(bool) false`. 

In situations like this, trying to call product object methods will obviously fail *("Fatal error: call to a member function get_purchase_note() on bool"),* so we need to guard against that risk. So, in this PR, we:

- Check if `$_product` has been filtered to an unexpected value
- If so, we log a warning
- We also place some additional safeguards around our subsequent use of `$_product`

Fixes #750

## How to test this PR

To test this you will need:

1. A WordPress site running WooCommerce Subscriptions.
2. The ability to log in as a customer who has one or more subscriptions (if necessary, go ahead and create these things).

Let's start by verifying we haven't introduced any problems to the happy path:

1. As the customer, login to the **My Account** area, click on the **Subscriptions** tab, and click to view one of your subscriptions.
2. It should render correctly and without errors. Tip: pay particular attention to the **Subscription Totals** area, which is one of the things we have modified.

Let's also verify a variant of the happy path, in which the catalog visibility of one of the subscription products is set to 'hidden':

1. Edit one of the subscription products.
2. Change the catalog visibility to private and save.
3. As the customer, reload the subscription view you were looking at previously. Everything should still render correctly and, within the totals table, the item name should no longer be a link (because of the change to the catalog visibility setting):

<div align="center">Subscription totals table, showing unlinked items</div>

For our next piece of testing, please now introduce some custom code to change `$_product` into an unexpected type. You can do this by adding a snippet like the following to `wp-content/mu-plugins/test-pr-759.php`, or however else you prefer:

```php
<?php

add_filter( 'woocommerce_subscriptions_order_item_product', '__return_false' );
```

1. Refresh the customer's view of the subscription.
2. It should again render nicely and without obvious errors.
3. This time, for any subscription products where the catalog visibility is private, the individual items **should still be linked** (it's *possible* this is the reason some third parties may have been filtering `$_product` to false).
4. Finally, if you visit the **WooCommerce ‣ Status ‣ Logs** screen, you should be able to locate a log entry providing information about this problem:

<div align="center">Logged error</div>

## Product impact

- [x] Added changelog entry (or does not apply) **added**
- [x] Will this PR affect WooCommerce Subscriptions? yes/no/tbc, add issue ref **it surfaces through that plugin, but doesn't have any other impacts**
- [x] Will this PR affect WooCommerce Payments? yes/no/tbc, add issue ref **no**
- [x]  Added deprecated functions, hooks or classes to the [spreadsheet](https://docs.google.com/spreadsheets/d/1xw9xszcPMnWsp4C8OKZMsLzZob7tOmWT7qMqmEIq314/edit#gid=0) **no deprecations**

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,11 +1,13 @@
 *** WooCommerce Subscriptions Core Changelog ***
 
+= 8.0.0 - xxxx-xx-xx =
+* Fix - Safeguards added to the Subscriptions Totals template used in the My Account area, to guard against fatal errors that could arise in unusual conditions.
+
 = 7.9.0 - 2025-01-10 =
 * Add - Compatibility with WooCommerce's new preview email feature introduced in 9.6.
 * Fix - Prevent payment gateways that complete Change Payment requests asynchronously from blocking future attempts to update payment methods for all subscriptions.
 * Fix - Resolved an issue that could lead to undefined variable when attempting to send a customer notification email with no order.
 * Fix - Prevents the customer's postal code being used as their billing city in the current session, following a change to payment details.
-* Fix - Safeguards added to the Subscriptions Totals template used in the My Account area, to guard against fatal errors that could arise in unusual conditions.
 * Dev - Use the subscription's customer ID during the `wcs_can_user_renew_early()` function call when sending customer notification emails.
 * Dev - Fix PHP deprecated warnings caused by calling esc_url with null.
 
--- a/changelog.txt
+++ b/changelog.txt
@@ -5,6 +5,7 @@
 * Fix - Prevent payment gateways that complete Change Payment requests asynchronously from blocking future attempts to update payment methods for all subscriptions.
 * Fix - Resolved an issue that could lead to undefined variable when attempting to send a customer notification email with no order.
 * Fix - Prevents the customer's postal code being used as their billing city in the current session, following a change to payment details.
+* Fix - Safeguards added to the Subscriptions Totals template used in the My Account area, to guard against fatal errors that could arise in unusual conditions.
 * Dev - Use the subscription's customer ID during the `wcs_can_user_renew_early()` function call when sending customer notification emails.
 * Dev - Fix PHP deprecated warnings caused by calling esc_url with null.
 
--- a/templates/myaccount/subscription-totals-table.php
+++ b/templates/myaccount/subscription-totals-table.php
@@ -4,7 +4,7 @@
  *
  * @package WooCommerce_Subscription/Templates
  * @since 1.0.0 - Migrated from WooCommerce Subscriptions v2.6.0
- * @version 7.1.0
+ * @version 7.2.0
  */
 
 if ( ! defined( 'ABSPATH' ) ) {
--- a/templates/myaccount/subscription-totals-table.php
+++ b/templates/myaccount/subscription-totals-table.php
@@ -35,7 +35,6 @@ if ( ! defined( 'ABSPATH' ) ) {
 						'entity_type' => gettype( $_product ),
 					)
 				);
-				continue;
 			}
 
 			if ( apply_filters( 'woocommerce_order_item_visible', true, $item ) ) {
@@ -51,7 +50,7 @@ if ( ! defined( 'ABSPATH' ) ) {
 					<?php endif; ?>
 					<td class="product-name">
 						<?php
-						if ( $_product && ! $_product->is_visible() ) {
+						if ( is_a( $_product, WC_Product::class ) && ! $_product->is_visible() ) {
 							echo wp_kses_post( apply_filters( 'woocommerce_order_item_name', $item['name'], $item, false ) );
 						} else {
 							echo wp_kses_post( apply_filters( 'woocommerce_order_item_name', sprintf( '<a href="%s">%s</a>', get_permalink( $item['product_id'] ), $item['name'] ), $item, false ) );
@@ -89,7 +88,8 @@ if ( ! defined( 'ABSPATH' ) ) {
 				<?php
 			}
 
-			$purchase_note = $_product->get_purchase_note();
+			$purchase_note = is_a( $_product, WC_Product::class ) ? $_product->get_purchase_note() : false;
+
 			if ( $subscription->has_status( array( 'completed', 'processing' ) ) && $purchase_note ) {
 				?>
 				<tr class="product-purchase-note">
--- a/templates/myaccount/subscription-totals-table.php
+++ b/templates/myaccount/subscription-totals-table.php
@@ -4,7 +4,7 @@
  *
  * @package WooCommerce_Subscription/Templates
  * @since 1.0.0 - Migrated from WooCommerce Subscriptions v2.6.0
- * @version 7.9.0 - Safety checks added to guard against fatal errors in certain unusual conditions
+ * @version 7.1.0
  */
 
 if ( ! defined( 'ABSPATH' ) ) {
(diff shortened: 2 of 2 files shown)


mattallan (changes requested):
Thanks @barryhughes for the PR!

I'm wondering if this change is a little too strict?

With these changes, improper use of this filter from a third-party will result in customers seeing missing line items, or an empty subscription items table with a totals section that doesn't make sense.

E.g. say I have a subscription with multiple items, and a store has bad custom code returning false for one of their products: 

```
add_filter( 'woocommerce_subscriptions_order_item_product', function( $product ) {
	if ( $product->get_id() === 2514 ) {
		return false;
	}
	return $product;
} );
```

The customer would see:

image

The previous PR that changed this line was only released in the last release (https://github.com/Automattic/woocommerce-subscriptions-core/pull/718), so while this PR will fix the fatal error, I suspect we will quickly hear from merchants that they have missing line items.

Since we don't use the `$_product` object for much on this template (check visibility and purchase note) and pull most of the data from the subscription line items, I'm wondering if we should just add safeguards to any lines using `$_product`, but still show the line items in the table.

Thoughts?

I really like your addition of logging a warning for unexpected data, so what I'm thinking is keep the logging in, but remove the `continue;`, then add some safeguards to the `$purchase_note = $_product->get_purchase_note();` code.

barryhughes:
Yeah ... I think you're right that this isn't the correct solution. In fact, I guess it's not impossible (especially given how recently the call to `$_product->get_purchase_note()` was introduced) that a 3PD might deliberately set `$_product` to `false`, in order to influence things [in this area](https://github.com/Automattic/woocommerce-subscriptions-core/blob/7.8.0/templates/myaccount/subscription-totals-table.php#L41-L45):

```php
if ( $_product && ! $_product->is_visible() ) {
	echo wp_kses_post( apply_filters( 'woocommerce_order_item_name', $item['name'], $item, false ) );
} else {
	echo wp_kses_post( apply_filters( 'woocommerce_order_item_name', sprintf( '<a href="%s">%s</a>', get_permalink( $item['product_id'] ), $item['name'] ), $item, false ) );
}
```

Whether or not it's the right way to do things, it probably wouldn't have caused significant issues before now. I'll rework this.

barryhughes:
Bit of a sidebar, in the sense I don't know that this PR is the right place to address it (also, I'm not sure if we've receievd realworld reports of problems about this), but it occurs that we also lack any checks on `$item`.

```php
foreach ( $subscription->get_items() as $item_id => $item ) {
	/* Much of the template's work happens within this loop.
	 * We expect $item to be a WC_Order_Item_Product but it's
	 * another filtered value (woocommerce_order_get_items)
	 * and could be set to almost anything.
	 */
}
```

From a defensive coding perspective, that could also be worth closing off.

barryhughes:
@mattallan thanks for the review: updated as discussed! I also 'upgraded' the pre-existing check on `$_product` to ensure it is a product object and isn't merely truthy. Further changes are contained in https://github.com/Automattic/woocommerce-subscriptions-core/pull/759/commits/9ffa14a0b24d2987d85352844f66d891d92070aa. Testing instructions updated.

templates/myaccount/subscription-totals-table.php
```diff
@@ -4,7 +4,7 @@
  *
  * @package WooCommerce_Subscription/Templates
  * @since 1.0.0 - Migrated from WooCommerce Subscriptions v2.6.0
- * @version 1.0.0 - Migrated from WooCommerce Subscriptions v2.6.0
+ * @version 7.9.0 - Safety checks added to guard against fatal errors in certain unusual conditions
```
mattallan:
```suggestion
 * @version 7.1.0
```

Just checking, this should be the version of WooCommerce Subscriptions these changes are shipped in (i.e. next version is 7.1). Is that correct?

mattallan (approved):
Thanks @barryhughes for the recent changes and discussion.

> but it occurs that we also lack any checks on `$item`.

Good call out! Probably worth addressing this in a separate issue/PR 💯

templates/myaccount/subscription-totals-table.php
barryhughes: Yes, sorry I hadn't pieced that together at the time I made this change. Thanks!

barryhughes: In fact, though I applied your change, this should now be `7.2.0`, right (because 7.1.0 just shipped)? I'll make that adjustment, but shout out if I'm mistaken

changelog.txt
```diff
@@ -5,6 +5,7 @@
 * Fix - Prevent payment gateways that complete Change Payment requests asynchronously from blocking future attempts to update payment methods for all subscriptions.
 * Fix - Resolved an issue that could lead to undefined variable when attempting to send a customer notification email with no order.
 * Fix - Prevents the customer's postal code being used as their billing city in the current session, following a change to payment details.
+* Fix - Safeguards added to the Subscriptions Totals template used in the My Account area, to guard against fatal errors that could arise in unusual conditions.
```
mattallan: Small change needed here. This changelog needs to be moved to the next version as 7.9 was shipped last week 😄

mattallan (approved):
Thanks @barryhughes, just a small change in the changelog needed but happy to approve and get this merged once it's fixed up :)

changelog.txt
barryhughes: Sure thing! Added a new section for 8.0.0 and moved it there → https://github.com/Automattic/woocommerce-subscriptions-core/pull/759/commits/68da33640c80ec59cecd2838a9e7e04fced22e05

mattallan (approved):
Thanks @barryhughes LGTM :)

PR #767, opened 2025-01-16
Make the admin notification mechanics contextually aware (of the user and of the admin screen)

barryhughes:
If an admin notice is added via the `wcs_add_admin_notice()` function, then it will display on the next request for an admin screen—regardless of who the current user is or what screen they are viewing. Here's a hypothetical example:

- Merchant changes the subscription end date in the editor.
- In response, we add an admin notice to tell them the change was successful (or possibly to inform them that there was a problem of some kind).
- If some other user, like a contributor, just happens to request a different admin page at approximately the same time (and immediately before the admin's instance of the subscription editor finishes reloading), then that is the user who will, quite unexpectedly, see the admin notice.
- Additionally, these operate as "flash" messages by default—and so, once they've been consumed, they are gone: in other words, in this hypothetical scenario, the merchant would never see the message. Only the contributor would see it, and they'd probably be a little confused by it.

In this change, we tighten things up a little such that:

- If a notice is enqueued, we assume its intended audience is the current user (and that's the only person that will see it). We also make it possible to explicitly specify the intended recipient.
- Additionally, it can be 'locked' to a specific context, by specifying the relevant screen ID (this would mean we can constrain certain notices so they only render in the subscription editor, to pick an example).

**How can this code break?**

We previously stored these notices in a transient named `_wcs_admin_notices`, but will now use one or more transients named `_wcs_admin_notices_<user_id>`. So, an enqueued notice could be lost upon update (I think this is acceptable, and would be a very temporary problem that probably would impact very few users indeed ... but we *could* soften the landing if we think it's necessary to do so).
   
Updates http://github.com/woocommerce/woocommerce-subscriptions/issues/4776

## How to test this PR

Let's first test that existing admin notices continue work as expected. To do this, select an existing subscription and contrive a problem in which the subscription end date is *earlier* than the next payment date:

<div align="center">Subscription payment dates meta box</div>

Normally it's not possible to do this, so setting it up will require direct edits to the relevant row of the `wc_orders_meta` table (assuming you are using HPOS). Find the `_schedule_end` entry, and alter the date manually as needed.

With that done, return to the editor and click update. Upon reload, you should see that an admin notice is present on the page telling you that the end date is incorrect:

Admin notices at the top of the subscription editor screen

On to the real testing! Let's confirm that admin notices are 'constrained' to the current user. Probably the easiest way to do this is to have two separate browser contexts: in one, you should be logged in as an admin user or shop manager, in the other you should be logged in as a contributor (or other less privileged user). You also need to add the following snippet to a suitable location, such as `wp-content/mu-plugins/test-pr-767.php`:

```php
<?php

add_action( 'admin_footer', function () {
	if (
		! isset( $_GET[ 'create-wcs-message' ] )
		|| ! function_exists( 'wcs_add_admin_notice' )
	) {
		return;
	}

	$user = wp_get_current_user();

	$username = $user instanceof WP_User ? $user->display_name : '&lt;unknown&gt;';
	wcs_add_admin_notice( "Hey there, $username! Time for a coffee?", 'success');
} );
```

- As the admin user, visit `example.com/wp-admin?create-wcs-message=1`. This triggers the code in the snippet.
- When the page finishes loading you should see ... nothing new (or, at least, you should not see an admin notice, unless perhaps WordPress adds one that is unrelated to the scenario we are testing).
- Now switch to the other browser context, where you are logged in as a contributor. Load or refresh one of the admin pages:
    - Without this fix, you should see a message reading something like _Hey there, **admin user**! Time for a coffee?_
    - If you repeat but with this fix, you should not see any admin notice message.
- Switch back to the original browser context, where you are logged in as an admin user. Load or refresh another admin page, but make sure `?create-wcs-message=1` is no longer in the URL.
    - Without this fix, you will not see an admin notice (it was consumed by the contributor).
    - *With* this fix, you *should* see the admin notice.

> [!TIP]
> In case it isn't clear, note that you need to run through the whole set of testing steps once without this fix, and then again *with* the fix. You can't flip between branches as you go along, or you will get unexpected results.

> [!NOTE]
> The change goes beyond what these testing instructions describe. For example, as noted earlier, it now becomes possible to lock a message to a specific screen. I have not detailed testing steps for this but will *probably* do so in a further PR that takes advantage of that functionality. Nonetheless, feel free to explore this area now if you feel so inclined!

## Product impact

- [x] Will this PR affect WooCommerce Subscriptions? ***Yes → [#4776](http://github.com/woocommerce/woocommerce-subscriptions/issues/4776)***
- [ ] Will this PR affect WooCommerce Payments? ***No***
- [ ]  Added deprecated functions, hooks or classes to the [spreadsheet](https://docs.google.com/spreadsheets/d/1xw9xszcPMnWsp4C8OKZMsLzZob7tOmWT7qMqmEIq314/edit#gid=0) ***None***

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -3,7 +3,7 @@
 = 8.0.0 - xxxx-xx-xx =
 * Fix - Safeguards added to the Subscriptions Totals template used in the My Account area, to guard against fatal errors that could arise in unusual conditions.
 * Fix - Correctly load product names with HTML on the cart and checkout shipping rates.
-* Fix - Admin notices are now locked down to the current user by default.
+* Fix - Improve our admin notice handling to ensure notices are displayed to the intended admin user.
 
 = 7.9.0 - 2025-01-10 =
 * Add - Compatibility with WooCommerce's new preview email feature introduced in 9.6.
--- a/changelog.txt
+++ b/changelog.txt
@@ -3,6 +3,7 @@
 = 8.0.0 - xxxx-xx-xx =
 * Fix - Safeguards added to the Subscriptions Totals template used in the My Account area, to guard against fatal errors that could arise in unusual conditions.
 * Fix - Correctly load product names with HTML on the cart and checkout shipping rates.
+* Fix - Admin notices are now locked down to the current user by default.
 
 = 7.9.0 - 2025-01-10 =
 * Add - Compatibility with WooCommerce's new preview email feature introduced in 9.6.
--- a/includes/admin/wcs-admin-functions.php
+++ b/includes/admin/wcs-admin-functions.php
@@ -27,7 +27,25 @@ if ( ! defined( 'ABSPATH' ) ) {
  * @return void
  */
 function wcs_add_admin_notice( $message, $notice_type = 'success', $user_id = null, $screen_id = null ) {
-	$user_id = null === $user_id ? get_current_user_id() : $user_id;
+	$user_id = (int) ( null === $user_id ? get_current_user_id() : $user_id );
+
+	if ( $user_id < 1 ) {
+		wc_get_logger()->warning(
+			sprintf(
+				/* Translators: %1$s: notice type ('success' or 'error'), %2$s: notice text. */
+				'Admin notices can only be added if a user is currently logged in. Attemped (%1$s) notice: "%2$s"',
+				$notice_type,
+				$message
+			),
+			array(
+				'backtrace' => true,
+				'user_id'   => $user_id,
+			)
+		);
+
+		return;
+	}
+
 	$notices = get_transient( '_wcs_admin_notices_' . $user_id );
 
 	if ( ! is_array( $notices ) ) {
--- a/includes/admin/wcs-admin-functions.php
+++ b/includes/admin/wcs-admin-functions.php
@@ -73,8 +73,9 @@ function wcs_display_admin_notices( $clear = true ) {
 		$notice_output = array();
 
 		foreach ( $notices as $index => $notice ) {
-			// Ensure the notice data now has the expected shape.
+			// Ensure the notice data now has the expected shape. If it does not, remove it.
 			if ( ! is_array( $notice ) || ! isset( $notice['message'] ) || ! array_key_exists( 'screen_id', $notice ) ) {
+				unset( $notices[ $index ] );
 				continue;
 			}
 
@@ -92,7 +93,7 @@ function wcs_display_admin_notices( $clear = true ) {
 
 		// $notice_output may be empty if some notices were withheld, due to not matching the screen context.
 		if ( ! empty( $notice_output ) ) {
-			echo '<div id="moderated" class="' . esc_attr( $class ) . '">' . wp_kses_post( implode( "</p>\n<p>", $notice_output ) ) . '</p></div>';
+			echo '<div id="moderated" class="' . esc_attr( $class ) . '"><p>' . wp_kses_post( implode( "</p>\n<p>", $notice_output ) ) . '</p></div>';
 		}
 	};
 
--- a/includes/admin/wcs-admin-functions.php
+++ b/includes/admin/wcs-admin-functions.php
@@ -13,59 +13,118 @@ if ( ! defined( 'ABSPATH' ) ) {
 }
 
 /**
- * Store a message to display via @see wcs_display_admin_notices().
+ * Registers an admin notice to be displayed once to the current (or other specified) user.
  *
- * @param string The message to display
- * @since 1.0.0 - Migrated from WooCommerce Subscriptions v2.0
+ * @see   wcs_display_admin_notices()
+ * @since 1.0.0 Migrated from WooCommerce Subscriptions v2.0.
+ * @since 7.2.0 Added support for specifying the target user and context.
+ *
+ * @param string      $message     The message to display.
+ * @param string      $notice_type Either 'success' or 'error'.
+ * @param int|null    $user_id     The specific user who should see this message. If not specified, defaults to the current user.
+ * @param string|null $screen_id   The screen ID for which the message should be displayed. If not specified, it will show on the next admin page load.
+ *
+ * @return void
  */
-function wcs_add_admin_notice( $message, $notice_type = 'success' ) {
+function wcs_add_admin_notice( $message, $notice_type = 'success', $user_id = null, $screen_id = null ) {
+	$user_id = null === $user_id ? get_current_user_id() : $user_id;
+	$notices = get_transient( '_wcs_admin_notices_' . $user_id );
 
-	$notices = get_transient( '_wcs_admin_notices' );
-
-	if ( false === $notices ) {
+	if ( ! is_array( $notices ) ) {
 		$notices = array();
 	}
 
-	$notices[ $notice_type ][] = $message;
+	$notices[ $notice_type ][] = array(
+		'message'   => $message,
+		'screen_id' => $screen_id,
+	);
 
-	set_transient( '_wcs_admin_notices', $notices, 60 * 60 );
+	set_transient( '_wcs_admin_notices_' . $user_id, $notices, HOUR_IN_SECONDS );
 }
 
 /**
- * Display any notices added with @see wcs_add_admin_notice()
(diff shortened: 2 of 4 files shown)


includes/admin/wcs-admin-functions.php
```diff
@@ -13,59 +13,119 @@
...
+ * @param string      $message     The message to display.
+ * @param string      $notice_type Either 'success' or 'error'.
+ * @param int|null    $user_id     The specific user who should see this message. If not specified, defaults to the current user.
+ * @param string|null $screen_id   The screen ID for which the message should be displayed. If not specified, it will show on the next admin page load.
+ *
+ * @return void
  */
-function wcs_add_admin_notice( $message, $notice_type = 'success' ) {
+function wcs_add_admin_notice( $message, $notice_type = 'success', $user_id = null, $screen_id = null ) {
+	$user_id = null === $user_id ? get_current_user_id() : $user_id;
```
mattallan: Should we add safeguards in both `wcs_add_admin_notice()` and `wcs_display_admin_notices()` for when `get_current_user_id()` returns 0?

For instance, if this function is called when someone isn't logged in (guest), we should probably return early to prevent `_wcs_admin_notices_0` being created and visible to any non-logged in user.

This feels extremely edge case, but wanted to raise it :)

includes/admin/wcs-admin-functions.php
```diff
@@ -13,59 +13,119 @@
...
 	}
 
-	if ( false !== $clear ) {
+	// Under certain circumstances, the caller may not wish for the rendered messages to be cleared from the queue.
+	if ( false === $clear ) {
+		return;
+	}
+
+	// If all notices were rendered, clear the queue. If only some were rendered, clear what we can.
+	if ( empty( $notices ) ) {
```
mattallan: What circumstances will `$notices` be empty?

If all notices are displayed won't `$notices` end up looking something like:

```
$notices = [
   'success' => [],
   'error'   => []
];
```

I think we need to `unset( $notices['success'] )` somewhere if all notices are displayed. Is that correct?

includes/admin/wcs-admin-functions.php
```diff
@@ -13,59 +13,119 @@
...
-	$notices = get_transient( '_wcs_admin_notices' );
+		foreach ( $notices as $index => $notice ) {
+			// Ensure the notice data now has the expected shape. If it does not, remove it.
+			if ( ! is_array( $notice ) || ! isset( $notice['message'] ) || ! array_key_exists( 'screen_id', $notice ) ) {
+				unset( $notices[ $index ] );
+				continue;
+			}
 
-	if ( false !== $notices && ! empty( $notices ) ) {
+			$screen    = get_current_screen();
```
mattallan: ~~Should we move this screen id logic to within a `! empty( $notice['screen_id'] )` check to avoid unnecessary processing.~~

~~Another thing I noticed is that we're calling `get_current_screen()` on every foreach loop and also every time `$handle_notices` function is called, however this should only needed to be called once.~~

Edit: I just looked up `get_current_screen()` and it basically does nothing... so my performance concerns above here are ill-founded and can be ignored 😅

changelog.txt
```diff
@@ -3,6 +3,7 @@
 = 8.0.0 - xxxx-xx-xx =
 * Fix - Safeguards added to the Subscriptions Totals template used in the My Account area, to guard against fatal errors that could arise in unusual conditions.
 * Fix - Correctly load product names with HTML on the cart and checkout shipping rates.
+* Fix - Admin notices are now locked down to the current user by default.
```
mattallan: Offering an alternative changelog entry:

```suggestion
* Fix - Improve our admin notice handling to ensure notices are displayed to the intended admin user.
```

🤷

mattallan (commented):
Thanks for including unit tests and nice work on this one @barryhughes !!
Also amazing PR description :)

I've left a few initial comments, but so far these change look good!

changelog.txt
barryhughes: Yes, that's nicer. Committed!

includes/admin/wcs-admin-functions.php
barryhughes: Yep, it's just a thin wrapper around a global var (I suppose we *could* access that directly, but I think the global should be treated as an internal implementation detail of WP rather than part of the public interface).

I realize you struck through the original thought, but, I tend to agree with what you wrote: avoiding repeat calls here is easy and, imo, doesn't reduce the clarity of the code (which is probably the only reason not to do it).

Updated in https://github.com/Automattic/woocommerce-subscriptions-core/pull/767/commits/4669e19f5adcff063f5b47b0408dc51b9a9b4346.

includes/admin/wcs-admin-functions.php
barryhughes: Yes, you're right. This was left over from an earlier iteration. Corrected in https://github.com/Automattic/woocommerce-subscriptions-core/pull/767/commits/b32d4d7131137bb3eed929e9a4dd4ad764c2fa3e. I'll probably also modify the relevant tests, but I think it makes sense to attend to [this comment](https://github.com/Automattic/woocommerce-subscriptions-core/pull/767#discussion_r1922951659) before doing so.

includes/admin/wcs-admin-functions.php
barryhughes: Agreed! An edge case, but not an impossible one. If this inadvertently happened every *n*-th request, then a transient like `_wcs_admin_notices_0` would be created and would continue to balloon in size; it would not automatically be cleared because the messages would never be rendered.

I addressed this in https://github.com/Automattic/woocommerce-subscriptions-core/pull/767/commits/0498cc25631012d6b877b5c4408ebf5119f3ad38 and added a further test to cover it, which also logs a warning if this situation arises (can be seen in the WC logs):

<div align="center">Shows WooCommerce logs, and a log entry describing an attempt to call wcs_add_admin_notice() when no user is logged in.</div>

You can trigger this/simulate the problem with a snippet like the following:

```php
add_action( 'admin_init', function () {
	wcs_add_admin_notice( 'Test bad notice', 'error', 0 );
} );
```

If I'm taking it too far and we'd rather just bail out without logging a warning, I'm happy to unwind that aspect (perhaps it's bloat, even if it won't generally be triggered). A *doing-it-wrong* is another possibility, here.

barryhughes:
@mattallan requesting your further review. I think I addressed all of your concerns, but of course let me know if I missed anything/if any of the adjustments are off-key. Thanks again for all the great feedback!

includes/admin/wcs-admin-functions.php
```diff
@@ -13,59 +13,141 @@
...
  */
-function wcs_add_admin_notice( $message, $notice_type = 'success' ) {
+function wcs_add_admin_notice( $message, $notice_type = 'success', $user_id = null, $screen_id = null ) {
+	$user_id = (int) ( null === $user_id ? get_current_user_id() : $user_id );
+
+	if ( $user_id < 1 ) {
+		wc_get_logger()->warning(
+			sprintf(
+				/* Translators: %1$s: notice type ('success' or 'error'), %2$s: notice text. */
+				'Admin notices can only be added if a user is currently logged in. Attemped (%1$s) notice: "%2$s"',
```
mattallan:
```suggestion
				'Admin notices can only be added if a user is currently logged in. Attempted (%1$s) notice: "%2$s"',
```

mattallan (approved):
This is looking great @barryhughes !

I tested:
- [x] existing uses of `wcs_display_admin_notices()` throughout subscriptions are still working as expected
- [x] transient is properly cleared after noticed is displayed
- [x] tested user specific notices and screen specific:

```
wcs_add_admin_notice( 'This is a test notice.', 'error' );
wcs_add_admin_notice( 'Show this to user 2.', 'error', 2 );
wcs_add_admin_notice( 'Only show this notice on subscription list.', 'success', 1, 'edit-shop_subscription' );
wcs_add_admin_notice( 'Invalid guest notice.', 'success', 0 );
```

All appears to be working as expected 💯 I've left a small comment to fix a typo but apart from that this is good !
Thanks again

includes/admin/wcs-admin-functions.php
barryhughes: Woops, great catch!

PR #769, opened 2025-01-23
Fix WooCommerce Subscriptions unit tests

alefesouza:
Fixes https://github.com/woocommerce/woocommerce-subscriptions/pull/4771

## Description

This PR undo a change introduced by https://github.com/Automattic/woocommerce-subscriptions-core/pull/718 which made a unit test stop working due to the HPOS support.

This change was made on a deprecated class and would require some more updates to support HPOS, for example use `date_created_gmt` instead of `post_created_gmt` [here](https://github.com/Automattic/woocommerce-subscriptions-core/blob/8588c3500ce5468d0eaa68c9481229946188dbe8/includes/upgrades/class-wcs-repair-2-0.php#L405).

## How to test this PR

1. Load this PR changes into WooCommerce Subscriptions test environment.
2. Run unit tests.

## Product impact

### DIFF
--- a/includes/upgrades/class-wcs-repair-2-0.php
+++ b/includes/upgrades/class-wcs-repair-2-0.php
@@ -395,8 +395,7 @@ class WCS_Repair_2_0 {
 	public static function repair_start_date( $subscription, $item_id, $item_meta ) {
 		global $wpdb;
 
-		$order      = wc_get_order( $subscription['order_id'] );
-		$start_date = $order->get_meta( '_paid_date', true );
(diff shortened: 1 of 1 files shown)


alefesouza:
@mattallan as the class is deprecated I think it should not be updated anymore, please tell me if you think we should make this deprecated class support HPOS so I can work on updating it.

mattallan (approved):
> as the class is deprecated I think it should not be updated anymore

100% agree with you there.

Since this function is deprecated and has other non-HPOS friendly code (like you mentioned the use of `post_date_gmt`), reverting this change makes sense.

Thanks @alefesouza ! Approving

PR #753, opened 2024-12-20
Add hooks before and after the subscription info table email

webdados:
Updates #752 

## Description

As mentioned on the issue:

> Depending on the subscription business, the information on the "Subscription information" email template might not be enough for the customer to understand why is he getting this email.
> 
> (...)
> 
> I have two suggestions:
> 
> - Load the order details email template after the subscription information table
> - Add hooks before and after the subscription information table so that developers can decide what else to add

This PR implements the second suggestion.

## How to test this PR

Hook into `wcs_email_before_subscription_info` or `wcs_email_after_subscription_info` and add any content you want to the subscription information template on emails.

## Product impact

- [x] Will this PR affect WooCommerce Subscriptions?

Issue #123, opened 2022-03-12
Undefined variable: $user_can_suspend in WCS_Customer_Suspension_Manager


vHeemstra:
## Note
Since there is no public place for the WooCommerce Subscriptions plugin code to report bugs, I'll submit it here. Since this package is part of the plugin.

## Describe the bug
When a user suspends their subscription the number of times he/she is allowed (if set value is not `unlimited`), PHP throws a warning:
`Notice: Undefined variable: user_can_suspend in /var/www/vhosts/domain-name.com/wp-content/plugins/woocommerce-subscriptions/includes/class-wcs-customer-suspension-manager.php on line 79`

The problem is inside `WCS_Customer_Suspension_Manager::can_customer_put_subscription_on_hold()`:
It's because `$user_can_suspend` is not set outside the `if` statement.

## Possible cause
Before the `if` statement, the method uses the variable name `$can_user_suspend`, which is similar, but not the same.

## Solution
Simply typo fix, I think. ;)

## To Reproduce
1. Looking at the source code of this function alone is enough. It's faulty code :)

### Expected behavior
No errors

### Actual behavior
Errors

## Product impact

- [X] Does this issue affect WooCommerce Subscriptions? yes
- [ ] Does this issue affect WooCommerce Payments? unknown, probably

## Additional context
WooCommerce Subscriptions
Plugin version 4.0.2

lkraav:
Yep this is real and should be caught by static analysis :thinking: 

Sample traceback https://sentry.io/share/issue/3ef5d8f8d4974545adf2c9c3794dc11b/

mattallan:
Thanks for the report!
I've pushed up a PR to fix this in the main plugin repo and it should go out in 4.1 (due to go out today or sometime this week)

PR #770, opened 2025-01-24
Add subscription item table to our customer notification emails and introduce before/after hooks for further customisations

### LINKED ISSUE #752: Allow adding additional content to the "Subscription information" email template
## Description
Depending on the subscription business, the information on the "Subscription information" email template might not be enough for the customer to understand why is he getting this email.

For example, I sell premium WordPress plugins, using WooCommerce Subscriptions and a 3rd party licensing plugin.

This is the email the customer gets:
Image

As you can see, there's absolutely no information about the subscription items. Yes, he can click the subscription ID, log in, and see its details, but we have a serious risk here that the customer just ignores this email, as he sees no identifiable information.

I have two suggestions:

- Load the order details email template after the subscription information table
- Add hooks before and after the subscription information table so that developers can decide what else to add

## Product impact

mattallan:
Fixes #752

## Description

We received the following feedback related to ur new [Subscriptions notification feature](https://woocommerce.com/document/subscriptions/subscriptions-notifications/) we released in 6.9.0: 
> the information on the "Subscription information" email template might not be enough for the customer to understand why is he getting this email.

> there's absolutely no information about the subscription items. Yes, they can click the subscription ID, log in, and see its details, but we have a serious risk here that the customer just ignores this email, as they see no identifiable information

The author opened a PR (https://github.com/Automattic/woocommerce-subscriptions-core/pull/753) to address these issues, however, while I was reviewing the PR, there were a few things I wanted to change so I figured it was easier to open a new PR instead of working from their fork.

Key changes in this PR:
- 0e10ca601537b9573efb2e99e9034bf59324f523 - Adds the subscription item table to all of our customer notification emails.
- 9f2f7fbfe2158a9f0a665c805f6478176bb437e3 - Adds new hooks at the start and end of the `subscription-info.php` template for further customisations
- 9aa0ad6d219d9c08eb49f0342c1b7792295a923a - Removes the "Here are the details" paragraphs as it felt unnecessary.

| Before | After |
|--------|--------|
| image | image | 

> [!NOTE]
> When I started looking more closely at this issue, I noticed there are a lot of inconsistencies across many of our subscription emails. For example:
> - our Cancelled email doesn't have the `"Subscriptions #123` header between the two tables, but our Expired and Suspended emails do (see images below)
> - these emails show a different subscription info table where the recurring price is the second column, vs our customer notification emails have it in the 4th column
>
> 🙈 I had to stop myself from turning PR this into a massive refactor... I do think we should tackle this at some stage, but maybe it can wait for the Woo core email improvements that are also in the pipeline.

| Cancelled email | Expired email |
|--------|--------|
| image | image | 

## How to test this PR

With WooCommerce 9.6 installed, you are to preview emails within **WooCommerce > Settings > Emails** which makes testing this feature a little easier. That said, it uses dummy data and doesn't use the plain text templates properly, therefore testing this PR requires:
1. Installing and activate the [WP Mail Logging](https://wordpress.org/plugins/wp-mail-logging/) plugin for ease with testing emails locally.
2. Adding `define( 'WP_ENVIRONMENT_TYPE', 'production' );` to your `wp-config.php` file. This is needed because notification emails are not sent on development/test sites.

### 1. Testing email template changes

1. Go to **WooCommerce > Settings > Subscriptions** and enable the "Customer Notification" feature (set any reminder time)
2. Go to **WooCommerce > Settings > Emails**
3. Enable the "Customer Notification: Automatic renewal notice" email and make sure it's set to HTML format.
4. Enable the "Customer Notification: Manual renewal notice" email and set this one to **plain text** format
image
5. Visit the WP Admin Edit subscription page for a subscription purchased with Stripe (or any other automatic payment method).
6. From the subscription actions, run the "Send upcoming renewal notification" action.
image
8. View the email and confirm you see an email with both the subscription info table and subscription items table:
image
9. Go back to Admin Edit Subscription and switch the subscription from Stripe to manual renewal:
image
10. Now that subscription is manual, run the "Send upcoming renewal notification" action again.
11. Because we set this to a plain text email, confirm the new subscription item table is correct in plain text 🙈👍 
image

--

### 2. Testing the new hooks

This PR introduces new `woocommerce_subscriptions_email_before_subscription_info` and `woocommerce_subscriptions_email_after_subscription_info` hooks, as well as adds the subscriptions item table which is attached to the `woocommerce_subscriptions_email_order_details` hook.

For testing customisation options, you can use the following snippet:

```php
add_action( 'woocommerce_before_template_part', function( $template_name, $template_path, $located, $args ) {
	if ( in_array( $template_name, [ 'emails/subscription-info.php', 'emails/plain/subscription-info.php' ], true ) ) {
		echo '<p style="color:red;"><strong>[[ TESTING ]] Before subscription info template</strong></p>';
	}
}, 10, 4 );

add_action( 'woocommerce_after_template_part', function( $template_name, $template_path, $located, $args ) {
	if ( in_array( $template_name, [ 'emails/subscription-info.php', 'emails/plain/subscription-info.php' ], true ) ) {
		echo '<p style="color:red;"><strong>[[ TESTING ]] After subscription info template</strong></p>';
	}
}, 10, 4 );

add_action( 'woocommerce_subscriptions_email_order_details', function() {
	echo '<p style="color:red;"><strong>[[ TESTING ]] After subscription item table</strong></p>';
}, 20, 0 );
```

Rather than going through and running the WP Admin actions, I used the email preview to confirm this is working as intended.

1. Add the above snippet to your store.
2. Go to **WooCommerce > Settings > Emails** and scroll to the bottom of the page.
3. Using the dropdown, preview a Customer Notification email and confirm your customisations are there:
image

## Product impact

> [!IMPORTANT]
> Once we ship these changes, we'll need to update [our documentation](https://woocommerce.com/document/subscriptions/subscriptions-notifications/) to include the latest images.

### DIFF
--- a/templates/emails/customer-notification-auto-trial-ending.php
+++ b/templates/emails/customer-notification-auto-trial-ending.php
@@ -64,6 +64,18 @@ do_action( 'woocommerce_email_header', $email_heading, $email ); ?>
 // Show subscription details.
 \WC_Subscriptions_Email::subscription_details( $subscription, $order, $sent_to_admin, $plain_text, true );
 
+/**
+ * 'woocommerce_subscriptions_email_order_details' hook.
+ *
+ * @since 7.2.0
+ *
+ * @param WC_Subscription|WC_Order $subscription  The subscription object.
+ * @param bool                     $sent_to_admin Whether the email is being sent to an admin.
+ * @param bool                     $plain_text    Whether the email is being sent as plain text.
+ * @param WC_Email                 $email         The email object, useful for accessing the email's properties and methods.
+ *
+ * @hooked WC_Subscriptions_Email::order_details() - 10.
+ */
 do_action( 'woocommerce_subscriptions_email_order_details', $subscription, $sent_to_admin, $plain_text, $email );
 
 /**
--- a/templates/emails/customer-notification-auto-trial-ending.php
+++ b/templates/emails/customer-notification-auto-trial-ending.php
@@ -3,7 +3,7 @@
  * Customer Notification: Free trial of an automatically renewed subscription is about to expire email.
  *
  * @package WooCommerce_Subscriptions/Templates/Emails
- * @version x.x.x
+ * @version 7.2.0
  */
 if ( ! defined( 'ABSPATH' ) ) {
 	exit; // Exit if accessed directly
@@ -12,7 +12,7 @@ if ( ! defined( 'ABSPATH' ) ) {
 /**
  * @hooked WC_Emails::email_header() Output the email header.
  *
- * @since x.x.x
+ * @since 6.9.0
  */
 do_action( 'woocommerce_email_header', $email_heading, $email ); ?>
 
@@ -76,6 +76,6 @@ if ( $additional_content ) {
 /**
  * @hooked WC_Emails::email_footer() Output the email footer.
  *
- * @since x.x.x
+ * @since 6.9.0
  */
 do_action( 'woocommerce_email_footer', $email );
--- a/templates/emails/customer-notification-auto-trial-ending.php
+++ b/templates/emails/customer-notification-auto-trial-ending.php
@@ -64,18 +64,7 @@ do_action( 'woocommerce_email_header', $email_heading, $email ); ?>
 // Show subscription details.
 \WC_Subscriptions_Email::subscription_details( $subscription, $order, $sent_to_admin, $plain_text, true );
 
-/**
- * 'woocommerce_subscriptions_email_order_details' hook.
- *
- * @since 7.2.0
- *
- * @param WC_Subscription|WC_Order $subscription  The subscription object.
- * @param bool                     $sent_to_admin Whether the email is being sent to an admin.
- * @param bool                     $plain_text    Whether the email is being sent as plain text.
- * @param WC_Email                 $email         The email object, useful for accessing the email's properties and methods.
- *
- * @hooked WC_Subscriptions_Email::order_details() - 10.
- */
+/** This action is documented in templates/emails/customer-notification-auto-renewal.php */
 do_action( 'woocommerce_subscriptions_email_order_details', $subscription, $sent_to_admin, $plain_text, $email );
 
 /**
--- a/templates/emails/customer-notification-auto-trial-ending.php
+++ b/templates/emails/customer-notification-auto-trial-ending.php
@@ -71,6 +71,8 @@ do_action( 'woocommerce_email_header', $email_heading, $email ); ?>
 // Show subscription details.
 \WC_Subscriptions_Email::subscription_details( $subscription, $order, $sent_to_admin, $plain_text, true );
 
+do_action( 'woocommerce_subscriptions_email_order_details', $subscription, $sent_to_admin, $plain_text, $email );
+
 /**
  * Show user-defined additional content - this is set in each email's settings.
  */
--- a/templates/emails/customer-notification-auto-trial-ending.php
+++ b/templates/emails/customer-notification-auto-trial-ending.php
@@ -59,13 +59,6 @@ do_action( 'woocommerce_email_header', $email_heading, $email ); ?>
 		);
 		?>
 	</p>
-
-	<p>
-		<?php
-			esc_html_e( 'Here are the details:', 'woocommerce-subscriptions' );
-		?>
-	</p>
-
 <?php
 
 // Show subscription details.
--- a/templates/emails/plain/subscription-info.php
+++ b/templates/emails/plain/subscription-info.php
@@ -4,7 +4,7 @@
  *
  * @author  Brent Shepherd / Chuck Mac
  * @package WooCommerce_Subscriptions/Templates/Emails
- * @version 1.0.0 - Migrated from WooCommerce Subscriptions v3.0.4
+ * @version 7.2.0
  */
 
 if ( ! defined( 'ABSPATH' ) ) {
@@ -16,6 +16,8 @@ if ( empty( $subscriptions ) ) {
 
 $has_automatic_renewal = false;
 $is_parent_order       = wcs_order_contains_subscription( $order, 'parent' );
+
+do_action( 'woocommerce_subscriptions_email_before_subscription_info', $subscriptions, $order, $has_automatic_renewal, $is_parent_order, $is_admin_email, $plain_text );
 
 echo "\n\n" . __( 'Subscription information', 'woocommerce-subscriptions' ) . "\n\n";
 foreach ( $subscriptions as $subscription ) {
@@ -61,3 +63,5 @@ if ( $has_automatic_renewal && ! $is_admin_email && $subscription->get_time( 'ne
 		)
 	);
 }
+
+do_action( 'woocommerce_subscriptions_email_after_subscription_info', $subscriptions, $order, $has_automatic_renewal, $is_parent_order, $is_admin_email, $plain_text );
--- a/templates/emails/plain/subscription-info.php
+++ b/templates/emails/plain/subscription-info.php
@@ -16,7 +16,7 @@ if ( empty( $subscriptions ) ) {
 $has_automatic_renewal = false;
 $is_parent_order       = wcs_order_contains_subscription( $order, 'parent' );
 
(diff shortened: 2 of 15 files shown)


templates/emails/customer-notification-auto-trial-ending.php
```diff
@@ -59,18 +59,13 @@
...
-			esc_html_e( 'Here are the details:', 'woocommerce-subscriptions' );
-		?>
-	</p>
-
 <?php
 
 // Show subscription details.
 \WC_Subscriptions_Email::subscription_details( $subscription, $order, $sent_to_admin, $plain_text, true );
 
+do_action( 'woocommerce_subscriptions_email_order_details', $subscription, $sent_to_admin, $plain_text, $email );
```
barryhughes, line 68: Something of a passing thought rather than a request for a change here in this PR:

- So I don't think we have a requirement for documenting hooks in this project, and in any case this is admittedly pretty clear and readable (to me) without.
- Yet, if I was authoring an action callback for `woocommerce_subscriptions_email_order_details` would I expect the first param to be of type `WC_Subscription`, as this implies, or of type `WC_Order`, as [this implies](https://github.com/Automattic/woocommerce-subscriptions-core/blob/7.9.0/templates/emails/admin-new-renewal-order.php#L23)?
- One is of course a subclass of the other, but it still feels like there is the potential for subtle bugs (or at least friction when developing on top of hooks like this, where the ostensible type may vary—but the 3PD may not initially recognize that), and so a param comment *could* make the situation (slightly) clearer.

Again, mostly talking out loud rather than requesting a change here.

templates/emails/plain/subscription-info.php
```diff
@@ -17,6 +16,8 @@
 $has_automatic_renewal = false;
 $is_parent_order       = wcs_order_contains_subscription( $order, 'parent' );
 
+do_action( 'woocommerce_subscriptions_email_before_subscription_info', $subscriptions, $order, $has_automatic_renewal, $is_parent_order, $is_admin_email, $plain_text );
```
barryhughes: Another hook musing 🙃 

This same hook has 6 params here, but has 7 in the matching HTML template. This is definitely workable (callback authors can state their hook accepts 7 args and the 7th can be optional—it will then work in both cases), *but,* is this good dev-ex?

Could it introduce friction/frustration if they notice the 7-arg hook first, then find it isn't being invoked in a plain-text context (or vice versa) ... might it be crisper and clearer to include a 7th arg here (set to false or null), for instance?

templates/emails/plain/subscription-info.php
```diff
@@ -61,3 +62,5 @@
 		)
 	);
 }
+
+do_action( 'woocommerce_subscriptions_email_after_subscription_info', $subscriptions, $order, $has_automatic_renewal, $is_parent_order, $is_admin_email, $plain_text );
```
barryhughes: 🖐🏼 Question for my own education: we currently have a mix of hook prefixes in use (some like this use `woocommerce_subscriptions_*` and others use `wcs_*`) ... is this longer form what we're using going forward, or is it context-dependent?

barryhughes: Thinking here that hooks represent a commitment, and also seeing that you mentioned the attractiveness of a wider refactoring, do we have a strong enough case here for introducing these new before/after hooks (considering a developer can use the existing `woocommerce_before_template_part` and `woocommerce_after_template_part` hooks)?

I'm not opposed, to be clear, and understand that specific hooks like this are superior in some respects, but still wanted to pose the question.

barryhughes (approved):
Great! Tests well: thanks, @mattallan.

I left some questions and observations, some are just for my own benefit and none are blockers, so I'm hitting the *approve* button.

templates/emails/plain/subscription-info.php
mattallan: > Question for my own education: we currently have a mix of hook prefixes in use (some like this use woocommerce_subscriptions_* and others use wcs_*) ... is this longer form what we're using going forward, or is it context-dependent?

Good question!
I don't know when this started, but our preference changed to use the long-form "woocommerce_subscriptions_" to avoid any potential confusion and conflicts between similar features coming to WooCommerce like "WooCommerce Shipping" may also use the `wcs_` prefix.

It's a bit of a weird situation because all of our helper functions use `wcs_x_y_z()` and so changing to use `woocommerce_subscriptions_x_y_z()` prefix is going to result in very long function names, and I'm not really sure what the best alternative is.

We're pretty bad at doing this consistently but a general rule I try to stick to is:
- Use the `wcs_` prefix for functions inside a `x-functions.php` file
- Use `WC_Subscription_` for class names
- Use `woocommerce_subscriptions_` prefix for hooks

---

> considering a developer can use the existing `woocommerce_before_template_part` and `woocommerce_after_template_part` hooks

That's a really good point.

My preference would be to not create an extra level of hooks if there's already an alternative and so I'm going to just remove this.

templates/emails/plain/subscription-info.php
mattallan: Thanks for noticing this! I don't believe this was intended and think I may have added the 7th parameter later and forgot to add it to the plain text email templates.

As per my comment on https://github.com/Automattic/woocommerce-subscriptions-core/pull/770#discussion_r1931496461, I'm going to remove these new before/after hooks since developers can already use `woocommerce_before|after_template_part` to achieve the same thing.

mattallan: Thanks @barryhughes for noticing this !

As per https://github.com/Automattic/woocommerce-subscriptions-core/pull/770#discussion_r1931496461 I'm going to remove these additional do_actions in favour of having 3rd-parties to use the existing before/after hooks inside `wc_get_template()`.

We can look at introducing these again later if we're a proper need for them :)

Fixed in 661a67bc

templates/emails/customer-notification-auto-trial-ending.php
mattallan: Thanks @barryhughes for highlighting this and I agree!

I've pushed aa738eab to add docblocks to these hooks. Let me know if it addresses the concerns :)

mattallan:
Thanks @barryhughes for the review!

Summary of latest changes:
- [`661a67b`](https://github.com/Automattic/woocommerce-subscriptions-core/pull/770/commits/661a67bcf6bf67e40aa6aeb5c2a0cd26a87055cb) I've removed the new before/after hooks in favour of being able to use the existing `woocommerce_before|after_template_part` hooks (I've updated the code snippet in the PR description)
- [`aa738ea`](https://github.com/Automattic/woocommerce-subscriptions-core/pull/770/commits/aa738eab5bd63e847c4430f982a66219c9e2f650) - Adds documentation to the new do_actions added to all customer notification emails.

Would you mind taking another look at this? Thanks

templates/emails/plain/subscription-info.php
barryhughes: > ...but a general rule I try to stick to is:

Thank you, that's helpful! 👍🏼

templates/emails/customer-notification-auto-trial-ending.php
barryhughes: That's great, thank you!

One last thought would be that we could document just one of the relevant hooks, and then 'point' to that from the others (as described in [4.1 Duplicate Hooks](https://developer.wordpress.org/coding-standards/inline-documentation-standards/php/#4-hooks-actions-and-filters)), something like this:

```php
/** This action is documented in templates/emails/file-with-canonical-docs.php */
```

The advantage being that, if we wanted to amend the docblock later, there is only one hook to update and therefore less chance of any single docblock being 'out of sync' with the others ... *but* this is perhaps getting nitpicky, and I'm not sure that even WP Core follows this all of the time.

barryhughes (approved):
Nice, thank you!

🪝 Left one more thought about hooks, however, definitely not any sort of blocker and can be ignored.

🔬 Re-tested, but using the updated instructions/snippets. Works great!

templates/emails/customer-notification-auto-trial-ending.php
mattallan: Not nitpicky at all and I'm glad you suggested this! You had me sold on amending one docblock instead of 10.

Addressed in bfb21f5f59056568b8b53796fc7260342b7e785b

mattallan:
Thanks @barryhughes for the review and testing. I've addressed the latest comment in bfb21f5f59056568b8b53796fc7260342b7e785b and will merge this once the checks turn green :)

PR #773, opened 2025-01-28
Fix blocks checkout local pickup

### LINKED ISSUE #399: Shipping Method Selection Conflict with Local Pickup Block
## Describe the bug
[WooCommerce Blocks](https://github.com/woocommerce/woocommerce-blocks) has a dedicated block for selecting Local Pickup during checkout. [There is a dedicated P2 about this functionality here for reference](https://woosupport.wordpress.com/2023/01/27/local-pickup-on-checkout/).

Due to the way subscriptions splits carts into multiple carts to handle recurring billing, there is a conflict when it comes to shipping method selection. This screenshot shows the behaviour you see:

Screenshot 2023-01-30 at 11 28 34

## To Reproduce

1. Go to Settings > Shipping > Local Pickup. Enable it and create some pickup locations.
2. Ensure another shipping method, such as a flat rate, is also enabled.
3. Add both subscription and regular products to your cart.
4. Go to checkout.
5. Select Local Pickup.
6. See how the rates are not updated for recurring carts (see above screenshot).

### Expected behavior

When you select Local Pickup, the shipping method for the recurring carts should also switch to Local Pickup.

### Actual behavior

When you select Local Pickup, the shipping method for the recurring carts remains as default and cannot be changed.

## Product impact

## Additional context

A potential/partial solution to keep recurring and regular carts in sync was created here https://github.com/Automattic/woocommerce-subscriptions-core/pull/264 however this is no longer applicable because we're not passing a `null` package ID. The action would receive package ID 0 (when local pickup is used, only one package is supported).

Therefore the logic in subscriptions needs to be either:

1. A local pickup rate ID was returned, so set all recurring carts to local pickup.

or

2. Create (via slot fills) an interface to select shipping methods (or pickup) for each recurring package separately. 

cc @senadir

alefesouza:
Fixes #399

## Description

Alongside https://github.com/woocommerce/woocommerce/pull/54946 this PR fixes #399 issue by disabling the "supports local pickup" flag on the blocks checkout page shipping rates when there's a subscription on the cart.

This still allows the local pickup methods to be selected but changes the blocks checkout page to show all shipping methods on a single page.

| Before | After |
|-|-|
| Screenshot 2025-01-28 at 7 30 26 PM | Screenshot 2025-01-28 at 7 59 53 PM |
|  Screenshot 2025-01-28 at 7 28 53 PM | |

## How to test this PR

* Load woocommerce/woocommerce#54946 changes on your WooCommerce installation.
* On WooCommerce direction run:

```
pnpm install
cd plugins/woocommerce
pnpm watch:build
```

* Follow https://github.com/woocommerce/woocommerce/pull/54946 testing instructions.

## Product impact

### DIFF
--- a/woocommerce-subscriptions-core.php
+++ b/woocommerce-subscriptions-core.php
@@ -8,3 +8,18 @@
  * Requires WP: 5.6
  * Version: 7.9.0
  */
+
+// Tell WooCommerce Blocks that all shipping methods
+// do not support local pickup because it originally does not
+// support mixing pickup and shipping methods.
+function disable_local_pickup_shipping_methods() {
+  add_filter('woocommerce_shipping_method_supports', function ( $supports, $feature ) {
+    if ( $feature === 'local-pickup' && WC_Subscriptions_Cart::cart_contains_subscription() ) {
+      return false;
+    }
+
+    return $supports;
+  }, 3, 10);
+}
+
+add_action( 'woocommerce_blocks_enqueue_checkout_block_scripts_before', 'disable_local_pickup_shipping_methods' );


alefesouza:
Closing to be discussed on the RFC pgizk6-7o-p2

PR #775, opened 2025-01-31
Customer Notification fixes and slight improvements

### LINKED ISSUE #764: Customer subscriptions notifications for "upcoming automatic renewal" are scheduled despite this notification type being disabled
## Describe the bug
I noticed that:
- when I have "Customer Notification: Automatic renewal notice" **disabled** in WooCommerce Settings -> Emails
- when I then enable the Customer Notifications feature in WooCommerce Settings -> Subscriptions

the software immediately starts to schedule these notifications anyway.

## To Reproduce
See above.

### Expected behavior
When the "Customer Notification: Automatic renewal notice" email type is disabled, I expect that they don't get scheduled.

### Actual behavior
See above.

## Product impact

## Additional context
I've been chatting with WooCommerce Support (case 9265049) but they appear to have misunderstood the reported issue.

mattallan:
Fixes #764 

## Description

Customer Notifications is a relatively new feature and while there hasn't been any critical bugs reported so far, one of our users has shared some concerns and warranted feedback (see #764) which I'm addressing in this PR.

### Bug Fixes

- c013987b5f86269125b9391136bb8830bc27f134 - **Our batch processor wasn't scheduling notifications for subscriptions that have `pending-cancel` status.**
  - There are two ways we schedule these notification events in Action Scheduler:
     1. **Batch Processor** - When a merchant first enables the Customer Notification feature, or updates the Reminder Timing settings, we use a batch processor unschedule and reschedule new actions using these settings.
     2. **Individual Subscription Update** - When a subscriptions date or status is updated, we also ensure all the notifications are properly scheduled in Action Scheduler.

> [!NOTE]
> This bug only exists within our batch processor code. Our code which scheduled these actions when a single subscription is saved is working correctly for pending cancelled subscriptions.

---

### Improvements/Changes

- 2a3565f933e7b4bda35d047ec2119b9bf3fee50a - Prevent calling `WCS_Email_Customer_Notification_{X}->trigger()` when the store has disabled the email.
  - If a merchant has enabled the Customer Notification feature in **WooCommerce > Settings > Subscriptions**, but has disabled the Trial End Notification email, we shouldn't be invoking the Email->trigger() function.
- 5b67c9fd72301e0569f896522fe72ff2dd8e3860 - Moves adding subscription notes to only when WCS_Debug is turned on.
  - Previously, if the Customer Notification settings were enabled but the notification email wasn't sent, we'd always add a note on the subscription to explain why. In the majority of cases, these notes would only create confusion to merchants who might be thinking they need to take action on this.
  - The reasons notification emails don't get sent are:
    1. Email is disabled
    2. Recipient not found
    3. Site is in development mode
    4. The reminder timer set (i.e. 1 week before subscription event) is greater than the actual Subscription's next payment or trial end etc.
  - Reasons 1, 3 & 4 are mostly unhelpful to merchants/store managers and as for 2, we don't have precedence for adding order notes if the recipient is not found and so I don't have a good enough reason why we should do it for these type of emails.

> [!NOTE]
> I've kept the subscription notes idea there for debugging purposes, but have made it so they require WCS_DEBUG to be enabled now.

- 2617772e9fec1546c4a697433724241990fd86f4 - Allow filtering the list of notifications that get scheduled in Action Scheduler
  - After enabling the Customer Notifications feature, by default, we will schedule the following scheduled actions:
    1. Next Payment notification - if the subscription has a next payment in the future
    2. Trial End notification - if the subscription has a trial end date in the future
    3. End/Expiration notification - if the subscription has an end date in the future (i.e. subscription has a defined billing length or the subscription has been canceled)
  - Some merchants may only be using this feature for the Next Payment notification and don't want scheduled actions created for the expiration notification. This filter allows them to remove it.

### Testing batch processor bug is fixed

1. Start on `trunk` branch
2. Go to **WooCommerce > Settings > Subscriptions** and disable the Customer Notifications feature.
3. Via Admin Edit Subscription page of My Account > View Subscription, mark a subscription as pending cancelled and make sure it has an end date that's at least 5 days in the future. While you're here, copy the subscription ID.
7. Go back to **WooCommerce > Settings > Subscriptions** and turn on the Customer Notification feature.
8. Visit **Tools > Scheduled Actions** and search for the pending cancelled subscription ID.
9. Notice that there's no pending `woocommerce_scheduled_subscription_customer_notification_expiration` action scheduled.
10. Switch to this branch and repeat the above steps, however this time, there should be a pending `woocommerce_scheduled_subscription_customer_notification_expiration` scheduled action.

> [!WARNING]
> If action scheduler isn't running by itself, you may need to manually kick-off the batch processor by running the following scheduled actions: `wcs_schedule_pending_batch_processes` & `wcs_run_batch_process`

### Testing small tweaks/improvements

1. Have the following custom code running on your store:
```php
add_filter( 'woocommerce_subscription_valid_customer_notification_types', function( $notification_types ) {
	return array_diff( $notification_types, [ 'end' ] );
}, 10, 1 );


// Place these in your wp-config.php
define( 'WCS_DEBUG', true );
define( 'WP_ENVIRONMENT_TYPE', 'development' );
```
2. Go to any Admin Edit Subscription page for an active subscription on your store and modify the dates by setting any valid next payment date, trial end date and end date. Make a note of the subscription ID being used for testing.
   - Update the status/dates ^ should trigger the customer notification actions for this subscription to be rescheduled in AS.
image
4. Go to **WooCommerce > Settings > Emails** and make sure the "Customer Notification: Free trial expiration: automatic payment notice" email is disabled.
5. Go to **Tools > Scheduled Actions** and search for the subscription ID.
6. Confirm there's only the trial expiration notification (there shouldn't be an end date notification since we're using the filter to remove it. Also note there's no renewal notification as this subscription has a valid trial date in the future)
7. Manually run the `woocommerce_scheduled_subscription_customer_notification_trial_expiration` scheduled action and confirm this email wasn't sent.
8. Go to the Edit Subscription and confirm there's no notes saying it was skipped.
9. While on the Edit Subscription page, delete the trial end date on the subscription and save.
10. Go back Action Scheduler and now run the `woocommerce_scheduled_subscription_customer_notification_renewal` scheduled action for this subscription.
11. Because `WCS_DEBUG` is enabled and `WP_ENVIRONMENT_TYPE` is set to a development site, you should see a note on the subscription explaining that the email was not sent.
image
13. Toggle `WCS_DEBUG` off and update the next payment date (+1 day or minute) and then re-run the `woocommerce_scheduled_subscription_customer_notification_renewal` scheduled action again and confirm no order notes were displayed and confirm no email was sent.
14. Now set `WP_ENVIRONMENT_TYPE` to `'production'`, update the next payment date again to create the scheduled action, then run the `woocommerce_scheduled_subscription_customer_notification_renewal` scheduled action one last time.
15. Confirm the next payment reminder email was sent.

Sorry for the large amount of testing instructions 😅 There's a lot involved when fully testing these changes. The changes are fairly straight forward though and considering I've tested the latest commits, feel free to focus more on code review :)

## Product impact

### DIFF
--- a/includes/class-wcs-action-scheduler-customer-notifications.php
+++ b/includes/class-wcs-action-scheduler-customer-notifications.php
@@ -497,7 +497,24 @@ class WCS_Action_Scheduler_Customer_Notifications extends WCS_Scheduler {
 			}
 		}
 
-		return $notifications;
+		/**
+		 * 'woocommerce_subscription_valid_customer_notification_types' filter.
+		 *
+		 * Allows filtering the list of notification types that will be scheduled for a particular subscription.
+		 *
+		 * Default array format returned:
+		 * Array(
+		 *   'next_payment', // exists if the subscription contains a next payment date in the future.
+		 *   'trial_end',    // exists if the subscription contains a trial end date in the future.
+		 *   'end'           // exists if the subscription contains an end date in the future.
+		 * )
+		 *
+		 * @since 7.2.0
+		 *
+		 * @param array           $notifications Array of valid notification types.
+		 * @param WC_Subscription $subscription  Subscription object.
+		 */
+		return apply_filters( 'woocommerce_subscription_valid_customer_notification_types', $notifications, $subscription );
 	}
 
 	/**
--- a/includes/class-wcs-action-scheduler-customer-notifications.php
+++ b/includes/class-wcs-action-scheduler-customer-notifications.php
@@ -515,7 +515,7 @@ class WCS_Action_Scheduler_Customer_Notifications extends WCS_Scheduler {
 		 * @param array           $notifications Array of valid notification types.
 		 * @param WC_Subscription $subscription  Subscription object.
 		 */
-		return apply_filters( 'woocommerce_subscription_valid_customer_notification_types', $notifications, $subscription );
+		return (array) apply_filters( 'woocommerce_subscription_valid_customer_notification_types', $notifications, $subscription );
 	}
 
 	/**
--- a/includes/class-wcs-action-scheduler-customer-notifications.php
+++ b/includes/class-wcs-action-scheduler-customer-notifications.php
@@ -498,15 +498,16 @@ class WCS_Action_Scheduler_Customer_Notifications extends WCS_Scheduler {
 		}
 
 		/**
-		 * 'woocommerce_subscription_valid_customer_notification_types' filter.
+		 * Filter: `woocommerce_subscription_valid_customer_notification_types`.
 		 *
 		 * Allows filtering the list of notification types that will be scheduled for a particular subscription.
 		 *
 		 * Default array format returned:
-		 * Array(
-		 *   'next_payment', // exists if the subscription contains a next payment date in the future.
-		 *   'trial_end',    // exists if the subscription contains a trial end date in the future.
-		 *   'end'           // exists if the subscription contains an end date in the future.
+		 *
+		 * array(
+		 *     'next_payment', // Exists if the subscription contains a next payment date in the future.
+		 *     'trial_end',    // Exists if the subscription contains a trial end date in the future.
+		 *     'end'           // Exists if the subscription contains an end date in the future.
 		 * )
 		 *
 		 * @since 7.2.0
--- a/includes/emails/class-wcs-email-customer-notification.php
+++ b/includes/emails/class-wcs-email-customer-notification.php
@@ -289,7 +289,7 @@ class WCS_Email_Customer_Notification extends WC_Email {
 	}
 
 	/**
-	 * If WCS_DEBUG is enabled, attaches a note to the subscription to detail why a reminder email was not sent.
+	 * If WCS_DEBUG or WP_DEBUG is enabled, attach a note to the subscription to detail why a reminder email was not sent.
 	 *
 	 * @param WC_Subscription $subscription
 	 * @param array|string    $reasons
@@ -297,7 +297,7 @@ class WCS_Email_Customer_Notification extends WC_Email {
 	 * @return false
 	 */
 	private function log_reminder_email_not_sent( $subscription, $reasons ) {
-		if ( defined( 'WCS_DEBUG' ) && WCS_DEBUG ) {
+		if ( ( defined( 'WCS_DEBUG' ) && WCS_DEBUG ) || ( defined( 'WP_DEBUG' ) && WP_DEBUG ) ) {
 			$reasons = (array) $reasons;
 
 			// translators: %1$s: email title, %2$s: list of reasons why email was skipped.
--- a/includes/emails/class-wcs-email-customer-notification.php
+++ b/includes/emails/class-wcs-email-customer-notification.php
@@ -274,7 +274,7 @@ class WCS_Email_Customer_Notification extends WC_Email {
 		$skipped_reasons = [];
 
 		if ( ! WC_Subscriptions_Email_Notifications::should_send_notification() ) {
-			$skipped_reasons[] = __( 'Not a production site', 'woocommerce-subscriptions' );
+			$skipped_reasons[] = __( 'Not a production site, or notifications have been globally disabled', 'woocommerce-subscriptions' );
 		}
 
 		if ( ! $this->get_recipient() ) {
--- a/includes/emails/class-wcs-email-customer-notification.php
+++ b/includes/emails/class-wcs-email-customer-notification.php
@@ -254,7 +254,7 @@ class WCS_Email_Customer_Notification extends WC_Email {
 	}
 
 	/**
-	 * Determine whether the customer reminder email should be sent and add an order note if it shouldn't.
+	 * Determines whether the customer reminder email should be sent.
 	 *
 	 * Reminder emails are not sent if:
 	 * - The Customer Notification feature is disabled.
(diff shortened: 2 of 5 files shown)


includes/class-wcs-action-scheduler-customer-notifications.php
```diff
@@ -497,7 +497,25 @@ public static function get_valid_notifications( $subscription ) {
...
+		 *     'trial_end',    // Exists if the subscription contains a trial end date in the future.
+		 *     'end'           // Exists if the subscription contains an end date in the future.
+		 * )
+		 *
+		 * @since 7.2.0
+		 *
+		 * @param array           $notifications Array of valid notification types.
+		 * @param WC_Subscription $subscription  Subscription object.
+		 */
+		return apply_filters( 'woocommerce_subscription_valid_customer_notification_types', $notifications, $subscription );
```
barryhughes: From a defensive coding perspective, what are your thoughts on applying a cast here?

```suggestion
		return (array) apply_filters( 'woocommerce_subscription_valid_customer_notification_types', $notifications, $subscription );
```

I'm thinking that, conceivably, the first callback might do something like `return false` (either inadvertently, or as a misconceived way of wiping the list of notifications). Then, if a hypothetical second callback did something like the following, we'd have a fatal error:

```php
function filter_callback_2( $notifications ) {
	$additional_notifications = [ 'foo', 'bar' ];
	return array_merge( $notifications, $additional_notifications ); # Oh-oh!
}
```

The cast doesn't solve all possible problems, but it might reduce the overall risk of fatals just a little.

includes/emails/class-wcs-email-customer-notification.php
```diff
@@ -267,30 +267,43 @@ public function get_default_additional_content() {
...
-			if ( ! $this->get_recipient() ) {
-				$should_skip[] = __( 'Recipient not found', 'woocommerce-subscriptions' );
-			}
+		$skipped_reasons = [];
 
-			if ( WCS_Action_Scheduler_Customer_Notifications::is_subscription_period_too_short( $subscription ) ) {
-				$should_skip[] = __( 'Subscription billing cycle too short', 'woocommerce-subscriptions' );
-			}
+		if ( ! WC_Subscriptions_Email_Notifications::should_send_notification() ) {
+			$skipped_reasons[] = __( 'Not a production site', 'woocommerce-subscriptions' );
```
barryhughes: Not essential we change this (because even if I'm right, it's not really new as of this PR), but would the following be more accurate?

```suggestion
			$skipped_reasons[] = __( 'Not a production site, or notifications have been globally disabled', 'woocommerce-subscriptions' );
```

I'm just being mindful here that `::should_send_notification()` in turn calls `::notifications_globally_enabled()`, so the current environment isn't the only reason for this (if I'm reading correctly, at least).

includes/emails/class-wcs-email-customer-notification.php
```diff
@@ -267,30 +267,43 @@ public function get_default_additional_content() {
...
+			$reasons = (array) $reasons;
+
 			// translators: %1$s: email title, %2$s: list of reasons why email was skipped.
-			$subscription->add_order_note( sprintf( __( 'Skipped sending "%1$s": %2$s', 'woocommerce-subscriptions' ), $this->title, '<br>- ' . implode( '<br>- ', $should_skip ) ) );
-			return false;
+			$subscription->add_order_note( sprintf( __( 'Skipped sending "%1$s": %2$s', 'woocommerce-subscriptions' ), $this->title, '<br>- ' . implode( '<br>- ', $reasons ) ) );
 		}
 
-		return true;
+		return false;
```
barryhughes: I understand the rationale for this change, and I'm not really suggesting we re-work it, but still thought I'd leave a few thoughts (but, it's not really critical stuff so do feel free to skip). From the related discussion:

> The reasons notification emails don't get sent are:
>
> 1. Email is disabled
> 2. Recipient not found
> 3. Site is in development mode
> 4. The reminder timer set (i.e. 1 week before subscription event) is greater than the actual Subscription's next payment or trial end etc.
>
> Reasons 1, 3 & 4 are mostly unhelpful to merchants/store managers.

Merchants are our most important constituent but, to provide a second lens, what if we also think of support engineers as being another tier of person for whom we want to make life easy. Faced with an inquiry about problems in this area, they now need to know that `WCS_DEBUG` should be enabled and need to ask the person they are supporting (the merchant) to try again and capture whatever notices might then be recorded (so, additional back-and-forth and a longer time span before they hopefully find a resolution to the problem at hand).

I guess I'm wondering if there is a different path. What if, as an alternative, we log via `wc_get_logger()` or, we do that and, when `WCS_DEBUG` is enabled, we additionally write those messages as order notes for increased visibility?

Using the logger means the support engineer can always direct the customer (merchant) to look for a specific category of errors/warnings/notices, without a further step in betweeen. Global settings/controls around Woo's logger also mean merchants can exercise control at that level if the flow of information was ever too much.

Again, this is more a *thing* that piqued my interest than a concrete request to change what we have today.

barryhughes (commented):
Looking good! Posting a few initial thoughts (changing location, and don't want to lose them), though I still need to complete testing.

barryhughes:
> Testing small tweaks/improvements
> 
> [...]
> 7. Go to the Edit Subscription and confirm there's no notes saying it was skipped.

~I may be mistaken/may be getting a little lost, but actually there *is* a note saying it was skipped (and I'd have thought that would be expected, because at this point we're in a debug/development context as per the setup instructions—but I wanted to check in, just in case I've misunderstood the instructions or tested incorrectly).~ Ignore ...!

barryhughes (approved):
Great, thank you.

I left a few notes/ideas, but none are blockers and it tests as described.

includes/emails/class-wcs-email-customer-notification.php
mattallan: Thanks @barryhughes for bringing this up!

I personally don't like the idea of using order notes for this and would prefer to move it entirely to just use `wc_get_logger()`.

That said, if we were to switch over to writing logs instead of order notes, should we be concerned with writing too many logs? Say if a very large store sells both weekly and yearly subscriptions and they have intentionally set their notification timing to be 10 days before renewals.

All of their weekly subscriptions renewal notifications will be skipped because the period is too short and logs will be written for every weekly renewal 🤔 

This was kind of the reason I had it behind `WCS_DEBUG` being enabled to avoid a lot of unnecessary logs.

A different approach I'm thinking is:
- Always write a log if the email was not sent due to an error occurring (i.e. `should_send_notification()` or `get_recipient()` has returned `false`)
- Only log if an email wasn't sent due a setup issue if `WCS_DEBUG` is enabled (i.e. if the email is disabled or `is_subscription_period_too_short()` returns true)

I agree that `WCS_DEBUG` is not a well known constant, so we could do something like `WCS_DEBUG || WP_DEBUG` (which is what I think you asked in Slack earlier)?

Thinking more long-term, I think it would be probably better to have a "Enable debug mode" option in **WooCommerce > Settings > Subscriptions** but that might be out of scope for this PR 😅

includes/emails/class-wcs-email-customer-notification.php
mattallan: I agree. Thanks!

includes/class-wcs-action-scheduler-customer-notifications.php
mattallan: Thanks for making this suggestion!

includes/emails/class-wcs-email-customer-notification.php
barryhughes: tl;dr → some replies below but, though I posed the question, I also don't think we *need* to change anything else here and now. Better to ship a solid improvement and iterate later if we need to, than delay delivery of a great improvement.

> I personally don't like the idea of using order notes for this and would prefer to move it entirely to just use wc_get_logger().

That's fair. 

> So we could do something like WCS_DEBUG || WP_DEBUG (which is what I think you asked in Slack earlier)?

I like that a lot.

> Should we be concerned with writing too many logs?
> [...]
> Thinking more long-term, I think it would be probably better to have a "Enable debug mode" option in WooCommerce > Settings > Subscriptions but that might be out of scope for this PR.

Overwhelming the logger is definitely a concern that's been discussed elsewhere (outside of the Subscriptions context), so there's probably something weight to that concern.

That said, the file logger *does* have log rotation, and it does also support a concept of a minimum logging threshold (set via `WC_LOG_THRESHOLD`) ... but, at least at present, those things also aren't easily adjustable by 'regular' users.

Another strategy I was turning over, besides a debug mode option like you suggested (which I like), would be a mechanism that pools log entries. For instance, when an email can't be sent because the site is in development mode:

- Record the subscription ID in a transient, or else increment a counter held in a transient *(transient, or other suitable vehicle).*
- Via either a unique scheduled action (or a recurring one, or even a *unique* recurring action), get the transient contents and use it to write a single long entry for all instances of the problem that have occurred so far, then delete the transient.

Conceptually, this would be *somewhat* akin to the browser console in cases where the exact same error is repeated multiple times in succession. This could even be bundled up into a reusable mechanism of some kind. That said, I guess it's an open question as to whether that would scale better (it's also way outside the scope for this PR, and, if it has any legs at all, would *probably* be better added to WC Core itself).

mattallan: > though I posed the question, I also don't think we need to change anything else here and now. Better to ship a solid improvement and iterate later if we need to, than delay delivery of a solid improvement.

I agree and thanks for mentioning this!

I've just pushed up a small commit (2e1a58d2) to make it so we also log these notes when WP_DEBUG is enabled but as for the rest, I will open a separate issue for this to be looked into/worked on sometime later.

Thanks again for the discussion!

I've retested the latest changes and will merge this once GH checks pass.

PR #762, opened 2025-01-11
Load HTML in product name on shipping rates

alefesouza:
Fixes woocommerce/woocommerce-subscriptions#4128

## Description

This PR fixes raw HTML being shown to the user on the shipping rates section of the classic cart and checkout page when the product name contains HTML.

This fix for the cart and checkout blocks is included on woocommerce/woocommerce#54386. 

| Before | After |
|-|-|
| Screenshot 2025-01-10 at 9 40 16 PM | Screenshot 2025-01-10 at 9 39 24 PM |
| Screenshot 2025-01-10 at 9 48 11 PM | Screenshot 2025-01-10 at 9 48 28 PM |

## How to test this PR

1. Create a subscription with HTML in the name.
2. Add this subscription to the cart.
3. Add a normal product to the cart.
4. Access the classic cart or checkout page.
5. The HTML in the product name for the shipping rate is not loaded.

## Product impact

- [x] Added changelog entry (or does not apply)
- [ ] Will this PR affect WooCommerce Subscriptions? yes/no/tbc, add issue ref
- [ ] Will this PR affect WooCommerce Payments? yes/no/tbc, add issue ref
- [ ]  Added deprecated functions, hooks or classes to the [spreadsheet](https://docs.google.com/spreadsheets/d/1xw9xszcPMnWsp4C8OKZMsLzZob7tOmWT7qMqmEIq314/edit#gid=0)

PR #784, opened 2025-02-13
Fixes a fatal error in 8.0.0 caused by passing the incorrect amount of args to `woocommerce_cart_item_name` filter

mattallan:
Reverts: https://github.com/Automattic/woocommerce-subscriptions-core/pull/762

See slack thread for details: p1739441477388209-slack-C07CBV7SKPH

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Subscriptions Core Changelog ***
 
+= 8.0.1 - 2025-xx-xx =
+* Fix: Revert a fix released in 8.0.0 which attempted to display HTML in product names within the cart/checkout shipping package section.
+
 = 8.0.0 - 2025-02-13 =
 * Fix - Recommend WooPayments when there is no available payment gateway.
 * Fix - Safeguards added to the Subscriptions Totals template used in the My Account area, to guard against fatal errors that could arise in unusual conditions.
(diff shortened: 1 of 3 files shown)


mattallan:
The changes introduced in #762 throw a fatal error when other plugins or third-party is using the `woocommerce_cart_item_name` hook. This same hook in WooCommerce Core has 3 args attached to it, however, we were only ever calling it with one arg.

Here's a quick snippet to reproduce the fatal:


```php
add_filter( 'woocommerce_cart_item_name', function( $content, $cart_item, $cart_item_key ) {
	return $content;
}, 9, 3 );
```

james-allan:
Changes look good to me. I confirmed the error by hooking onto the filter with something like this:

```php
function dummy_function_filter( $name, $cart_item, $cart_item_key ) {
	return $name . 'dummy';
}

add_filter( 'woocommerce_cart_item_name', 'dummy_function_filter', 11, 3 );
```

<p align="center">
Screenshot 2025-02-13 at 8 39 25 pm
</p> 

On this branch there's no error.

james-allan (approved):

changelog.txt
```diff
@@ -1,5 +1,8 @@
 *** WooCommerce Subscriptions Core Changelog ***
 
+= 8.0.1 - 2025-xx-xx =
+* Fix: Revert a fix released in 8.0.0 which attempted to display HTML in product names within the cart/checkout shipping package section.
```
```suggestion
* Fix - Revert a change released in 7.2.0 which triggered the "woocommerce_cart_item_name" filter with the wrong number of parameters.
```

PR #787, opened 2025-02-19
Allow the Cancelled Subscription email to be dispatched twice

### LINKED ISSUE #771: Email type "Expired Subscription" is not being sent
## Describe the bug
In one of my stores using WooCommerce Subscriptions, I have enabled two email types:
- Cancelled Subscription
- Expired Subscription

Both are supposed to be sent to the store manager.

What happens is:
- When a customer cancels their subscription, the corresponding email gets sent just fine
- At the date and time the subscription expires, the "Expired Subscription" email is **not** being sent.

We want to use these expiration notification emails to ensure that we cancel the service being provided to the former subscriber.

## To Reproduce
See above.

### Expected behavior
I would expect that the store manager receives the notification email at the date and time the subscription expires.

### Actual behavior
The email notification mentioned does not get sent.

## Product impact
WooCommerce Subscriptions

## Additional context
n/a, nothing gets logged because nothing happens

barryhughes:
We offer merchants email-based notifications when a subscription is cancelled. When enabled, it will be dispatched when a subscription is set to either `pending-cancel` or `cancelled`, but not both.

In other words, if a subscription is initially set to `pending-cancel` (typically, this is the result of the customer self-serving), then the merchant will receive an email. When it then finally ends (when the status is updated to `cancelled`), the merchant will **not** receive the same cancellation email a second time.

However, as identified in the linked issue, this is not always desirable and, for some merchants, it may be useful to receive the email twice.

Fixes #771

## Description

There are two separate but related changes bundled up in this PR. The first and probably the most important is a change to support optionally sending the Cancelled Subscription email twice:

New setting for the Cancelled Subscription email

Secondly, the Cancelled Subscription email itself has received a small tweak, just to help differentiate between the two different cancellation states. When a subscription is *pending cancellation,* we say as much and indicate when it will finally end:

<div align="center">Revised text for pending cancellation emails</div>

## How to test this PR

When testing, you will require something like [Email Log](https://wordpress.org/plugins/email-log/) in order to review the emails that are dispatched.

1. Visit **WooCommerce ‣ Settings ‣ Emails** and look for the **Cancelled Subscription** email. Click to manage the settings.
    - Verify that, first time you try this, the new **Always Send** checkbox is disabled.
    - Please also ensure the email is enabled.
2. As a customer, purchase a subscription and then visit the **My Account ‣ Subscription** area. Locate the subscription and cancel it (this sets the subscription to `pending-cancel`).
    - The cancellation email should have been dispatched. 
    - You should see the new wording, as in the above screenshot: *A subscription belonging to [name] is pending cancellation, and will end on [date].*
3. As the admin, locate the same subscription in the admin environment and click to `Cancel Now`. The cancellation email should **not** be sent a second time.
4. Still acting as the admin user/merchant, find another active subscription and cancel it. The cancellation email **should** be dispatched for this other subscription.
5. Acting as the customer for one of the fully cancelled subscriptions, go ahead and resubscribe ... then cancel again. The cancellation email **should** be dispatched (this is to ensure we haven't introduced a regression for #658).
6. Now return to the **Cancelled Subscription** settings. Enable **Always Send** and save.
7. Repeat step 2 (results should be the same).
8. Repeat step 3, but this time the email should be sent a second time.

## Product impact

- [x] Will this PR affect WooCommerce Subscriptions? ***yes—#771*** 
- [x]  Added deprecated functions, hooks or classes to the [spreadsheet](https://docs.google.com/spreadsheets/d/1xw9xszcPMnWsp4C8OKZMsLzZob7tOmWT7qMqmEIq314/edit#gid=0) ***no deprecations***

## Thoughts and open questions

- You may also wish to perform testing in which you take advantage of the `woocommerce_subscription_use_pending_cancel` filter hook (ie, to always skip the `pending-cancel` status and go straight to `cancelled`).
- How do we feel about the changes to the email text—too subtle? Or perhaps not needed?


tests/unit/includes/test-class-wcs-email-cancelled-subscription.php
```diff
@@ -0,0 +1,134 @@
...
+	 * Initialize WC_Emails (required by our subject-under-test), and setup our email watcher.
+	 *
+	 * @return void
+	 */
+	public function set_up() {
+		$this->email_subject = '';
+		$this->email_sent    = false;
+
+		new WC_Emails();
+		$this->sut          = new WCS_Email_Cancelled_Subscription();
```
leonardola: What does sut mean? This might not be clear for people reading the code later. Maybe change it to `$email_cancelled_subscription_mock` instead?

leonardola (commented):
With **Always Send** disabled:

> Verify that, first time you try this, the new Always Send checkbox is disabled.

✅ 

> The cancellation email should have been dispatched.

✅ 

> You should see the new wording, as in the above screenshot: A subscription belonging to [name] is pending cancellation, and will end on [date].

✅ 

> As the admin, locate the same subscription in the admin environment and click to Cancel Now. The cancellation email should not be sent a second time.

✅ 

> Still acting as the admin user/merchant, find another active subscription and cancel it. The cancellation email should be dispatched for this other subscription.

✅ 

> Acting as the customer for one of the fully cancelled subscriptions, go ahead and resubscribe ... then cancel again. The cancellation email should be dispatched (this is to ensure we haven't introduced a regression for https://github.com/Automattic/woocommerce-subscriptions-core/issues/658).

✅ 

With **Always Send** enabled:

> Now return to the Cancelled Subscription settings. Enable Always Send and save. 
The cancellation email should have been dispatched.

 ✅ 

> You should see the new wording, as in the above screenshot: A subscription belonging to [name] is pending cancellation, and will end on [date].

✅

tests/unit/includes/test-class-wcs-email-cancelled-subscription.php
barryhughes: SUT := *Subject (or System) Under Test*

I thought it was a fairly common term, but definitely happy to change it if it makes things clearer to all. Thanks for flagging!

barryhughes: Updated in https://github.com/Automattic/woocommerce-subscriptions-core/pull/787/commits/70f6fdb47d79949dfdc5abbe395ab37db1e9e5d1. @leonardola let me know if that works for you 👍🏼

leonardola (approved):
LGTM

PR #803, opened 2025-03-04
Fix subscription product tutorial position

alefesouza:
Fixes https://github.com/woocommerce/woocommerce-subscriptions/issues/4808

## Description

This PR fixes the subscription product tutorial popup position, including when resizing the window or moving the sections.

## How to test this PR

1. Comment out [this](https://github.com/Automattic/woocommerce-subscriptions-core/blob/d5df34193616042553b13e5d121dc56f5bebf585/includes/admin/class-wc-subscriptions-admin.php#L924) and [this](https://github.com/Automattic/woocommerce-subscriptions-core/blob/d5df34193616042553b13e5d121dc56f5bebf585/includes/admin/class-wc-subscriptions-admin.php#L940) line.
2. Access `/wp-admin/post-new.php?post_type=product&subscription_pointers=true`.
3. The tutorial popup should be right positioned.
4. Move the page sections.
5. The tutorial popup should be right positioned.

## Product impact

### DIFF
--- a/assets/js/admin/admin-pointers.js
+++ b/assets/js/admin/admin-pointers.js
@@ -1,6 +1,15 @@
 jQuery( function ( $ ) {
+	let observer = null;
+
 	if ( arePointersEnabled() ) {
-		setTimeout( showSubscriptionPointers, 800 ); // give TinyMCE a chance to finish loading
+		observer = new MutationObserver( showSubscriptionPointers );
+	 
+	 	observer.observe( document.getElementById( 'poststuff' ), {
+			attributes: true,
+			childList: true,
+			characterData: false,
+			subtree:true,
+		} );
 	}
 
 	$( 'select#product-type' ).on( 'change', function () {
@@ -62,5 +71,9 @@ jQuery( function ( $ ) {
 			pointer: 'wcs_pointer',
 			action: 'dismiss-wp-pointer',
 		} );
+
+		if ( observer ) {
+			observer.disconnect();
+		}
 	}
 } );


alefesouza:
The E2Es should be fixed after merging #802

alefesouza:
@james-allan 

> After clicking the dismiss button another one pops up. This behavior is strange. A small improvement could be to move the “X” to the top right corner of the blue section and change “Dismiss” to “Next”.

Actually this second popup is shown when the `Simple subscription` option is selected, do you recommend any change?

james-allan:
> > After clicking the dismiss button another one pops up. This behavior is strange. A small improvement could be to move the “X” to the top right corner of the blue section and change “Dismiss” to “Next”.
>
> Actually this second popup is shown when the Simple subscription option is selected, do you recommend any change?

Yeah I just noticed that too. That "Dismiss" vs "Next" suggestion came from the original report. Because there are only 2 tips we display - the product type one and the price one - I don't think we need a "Next" button. The price tip isn't relevant until the user selects a subscription product type anyway.

james-allan (commented):
The position of the tip look good.


| Before | After |
|--------|--------|
| Screenshot 2025-03-05 at 2 50 08 pm | Screenshot 2025-03-05 at 2 52 36 pm | 

Are you seeing the same issue where dismiss is being squeezed to the right? I notice Poli's screenshots didn't have this issue so it might just be a local thing for me.

james-allan:
> Are you seeing the same issue where dismiss is being squeezed to the right?

Never mind, there was another plugin displaying tips on this screen too and they were doing some overriding of the core wp styles. 😞 It looks fine now. 

<p align="center">
Screenshot 2025-03-05 at 2 56 34 pm
</p>

james-allan (approved):
Just noting that there's no changelog in this PR. Given this is minor visual thing, I'm ok with leaving one out. I'll leave that up to you though.

PR #807, opened 2025-03-11
Subscription product editor consistency tweaks

alefesouza:
Fixes woocommerce/woocommerce-subscriptions#4807

## Description

Fix visual inconsistencies in the subscription and variable subscription editor.

| Before | After |
|-|-|
| Screenshot 2025-03-10 at 10 05 47 PM | Screenshot 2025-03-10 at 10 05 19 PM |

## How to test this PR

1. Change the store currency to EUR or another one that do not dot as decimal separator.
2. Create a new simple subscription.
3. The subscription prices dropbox should have the correct font size (14px).
4. The sign-up fee placeholder should use the correct decimal separator.
5. Change to variable subscription.
6. Create variations.
7. Edit one variation.
8. The fields should match woocommerce/woocommerce-subscriptions#4807 requirements.

## Product impact

### DIFF
--- a/includes/admin/class-wc-subscriptions-admin.php
+++ b/includes/admin/class-wc-subscriptions-admin.php
@@ -317,7 +317,7 @@ class WC_Subscriptions_Admin {
 			</label>
 			<span class="wrap">
 				<?php // Translators: %s: formatted example price value. ?>
-				<input type="text" id="_subscription_price" name="_subscription_price" class="wc_input_price wc_input_subscription_price" placeholder="<?php printf( esc_attr_x( 'e.g. %s', 'example price', 'woocommerce-subscriptions' ), esc_attr( wc_format_localized_price( '5.90' ) ) ); ?>" step="any" min="0" value="<?php echo esc_attr( wc_format_localized_price( $chosen_price ) ); ?>" />
+				<input type="text" id="_subscription_price" name="_subscription_price" class="wc_input_price wc_input_subscription_price" placeholder="<?php echo esc_attr( sprintf( _x( 'e.g. %s', 'example price', 'woocommerce-subscriptions' ), wc_format_localized_price( '5.90' ) ) ); ?>" step="any" min="0" value="<?php echo esc_attr( wc_format_localized_price( $chosen_price ) ); ?>" />
 				<label for="_subscription_period_interval" class="wcs_hidden_label"><?php esc_html_e( 'Subscription interval', 'woocommerce-subscriptions' ); ?></label>
 				<select id="_subscription_period_interval" name="_subscription_period_interval" class="wc_input_subscription_period_interval wc-enhanced-select">
 				<?php foreach ( wcs_get_subscription_period_interval_strings() as $value => $label ) { ?>
@@ -356,7 +356,7 @@ class WC_Subscriptions_Admin {
 					// translators: %s is a currency symbol / code
 					'label'             => sprintf( __( 'Sign-up fee (%s)', 'woocommerce-subscriptions' ), get_woocommerce_currency_symbol() ),
 					'placeholder'       => // translators: %s the formatted example price value.
-						sprintf( esc_attr_x( 'e.g. %s', 'example price', 'woocommerce-subscriptions' ), wc_format_localized_price( '5.90' ) ),
+						esc_attr( sprintf( _x( 'e.g. %s', 'example price', 'woocommerce-subscriptions' ), wc_format_localized_price( '5.90' ) ) ),
 					'description'       => __( 'Optionally include an amount to be charged at the outset of the subscription. The sign-up fee will be charged immediately, even if the product has a free trial or the payment dates are synced.', 'woocommerce-subscriptions' ),
 					'desc_tip'          => true,
 					'type'              => 'text',
--- a/includes/admin/class-wc-subscriptions-admin.php
+++ b/includes/admin/class-wc-subscriptions-admin.php
@@ -316,10 +316,7 @@ class WC_Subscriptions_Admin {
 				?>
 			</label>
 			<span class="wrap">
-				<input type="text" id="_subscription_price" name="_subscription_price" class="wc_input_price wc_input_subscription_price" placeholder="<?php
-					// translators: %s the formatted example price value.
-					echo sprintf( esc_attr_x( 'e.g. %s', 'example price', 'woocommerce-subscriptions' ), wc_format_localized_price( '5.90' ) );
-				?>" step="any" min="0" value="<?php echo esc_attr( wc_format_localized_price( $chosen_price ) ); ?>" />
+				<input type="text" id="_subscription_price" name="_subscription_price" class="wc_input_price wc_input_subscription_price" placeholder="<?php echo esc_attr_x( 'e.g.', 'example price', 'woocommerce-subscriptions' ); ?> <?php echo esc_attr( wc_format_localized_price( '5.90' ) ); ?>" step="any" min="0" value="<?php echo esc_attr( wc_format_localized_price( $chosen_price ) ); ?>" />
 				<label for="_subscription_period_interval" class="wcs_hidden_label"><?php esc_html_e( 'Subscription interval', 'woocommerce-subscriptions' ); ?></label>
 				<select id="_subscription_period_interval" name="_subscription_period_interval" class="wc_input_subscription_period_interval wc-enhanced-select">
 				<?php foreach ( wcs_get_subscription_period_interval_strings() as $value => $label ) { ?>
--- a/includes/admin/class-wc-subscriptions-admin.php
+++ b/includes/admin/class-wc-subscriptions-admin.php
@@ -316,10 +316,12 @@ class WC_Subscriptions_Admin {
 				?>
 			</label>
 			<span class="wrap">
-				<input type="text" id="_subscription_price" name="_subscription_price" class="wc_input_price wc_input_subscription_price" placeholder="<?php
-				  // translators: %s the formatted example price value.
-				  echo sprintf( esc_attr_x( 'e.g. %s', 'example price', 'woocommerce-subscriptions' ), wc_format_localized_price( '5.90' ) );
-				?>" step="any" min="0" value="<?php echo esc_attr( wc_format_localized_price( $chosen_price ) ); ?>" />
+				<input type="text" id="_subscription_price" name="_subscription_price" class="wc_input_price wc_input_subscription_price" placeholder="
+				<?php
+				// translators: %s the formatted example price value.
+				echo esc_attr( sprintf( 'e.g. %s', wc_format_localized_price( '5.90' ) ) );
+				?>
+				" step="any" min="0" value="<?php echo esc_attr( wc_format_localized_price( $chosen_price ) ); ?>" />
 				<label for="_subscription_period_interval" class="wcs_hidden_label"><?php esc_html_e( 'Subscription interval', 'woocommerce-subscriptions' ); ?></label>
 				<select id="_subscription_period_interval" name="_subscription_period_interval" class="wc_input_subscription_period_interval wc-enhanced-select">
 				<?php foreach ( wcs_get_subscription_period_interval_strings() as $value => $label ) { ?>
--- a/includes/admin/class-wc-subscriptions-admin.php
+++ b/includes/admin/class-wc-subscriptions-admin.php
@@ -316,7 +316,8 @@ class WC_Subscriptions_Admin {
 				?>
 			</label>
 			<span class="wrap">
-				<input type="text" id="_subscription_price" name="_subscription_price" class="wc_input_price wc_input_subscription_price" placeholder="<?php echo esc_attr_x( 'e.g.', 'example price', 'woocommerce-subscriptions' ); ?> <?php echo esc_attr( wc_format_localized_price( '5.90' ) ); ?>" step="any" min="0" value="<?php echo esc_attr( wc_format_localized_price( $chosen_price ) ); ?>" />
+				<?php // Translators: %s: formatted example price value. ?>
+				<input type="text" id="_subscription_price" name="_subscription_price" class="wc_input_price wc_input_subscription_price" placeholder="<?php printf( esc_attr_x( 'e.g. %s', 'example price', 'woocommerce-subscriptions' ), esc_attr( wc_format_localized_price( '5.90' ) ) ); ?>" step="any" min="0" value="<?php echo esc_attr( wc_format_localized_price( $chosen_price ) ); ?>" />
 				<label for="_subscription_period_interval" class="wcs_hidden_label"><?php esc_html_e( 'Subscription interval', 'woocommerce-subscriptions' ); ?></label>
 				<select id="_subscription_period_interval" name="_subscription_period_interval" class="wc_input_subscription_period_interval wc-enhanced-select">
 				<?php foreach ( wcs_get_subscription_period_interval_strings() as $value => $label ) { ?>
--- a/includes/admin/class-wc-subscriptions-admin.php
+++ b/includes/admin/class-wc-subscriptions-admin.php
@@ -318,7 +318,7 @@ class WC_Subscriptions_Admin {
 			<span class="wrap">
 				<input type="text" id="_subscription_price" name="_subscription_price" class="wc_input_price wc_input_subscription_price" placeholder="<?php
 				  // translators: %s the formatted example price value.
-				  echo sprintf( esc_attr_x( 'e.g. %s', 'example price', 'woocommerce-subscriptions' ), str_replace( get_woocommerce_currency_symbol(), '', strip_tags( wc_price( 5.90 ) ) ) );
+				  echo sprintf( esc_attr_x( 'e.g. %s', 'example price', 'woocommerce-subscriptions' ), wc_format_localized_price( '5.90' ) );
 				?>" step="any" min="0" value="<?php echo esc_attr( wc_format_localized_price( $chosen_price ) ); ?>" />
 				<label for="_subscription_period_interval" class="wcs_hidden_label"><?php esc_html_e( 'Subscription interval', 'woocommerce-subscriptions' ); ?></label>
 				<select id="_subscription_period_interval" name="_subscription_period_interval" class="wc_input_subscription_period_interval wc-enhanced-select">
@@ -358,7 +358,7 @@ class WC_Subscriptions_Admin {
 					// translators: %s is a currency symbol / code
 					'label'             => sprintf( __( 'Sign-up fee (%s)', 'woocommerce-subscriptions' ), get_woocommerce_currency_symbol() ),
 					'placeholder'       => // translators: %s the formatted example price value.
-						sprintf( esc_attr_x( 'e.g. %s', 'example price', 'woocommerce-subscriptions' ), str_replace( get_woocommerce_currency_symbol(), '', strip_tags( wc_price( 9.90 ) ) ) ),
+						sprintf( esc_attr_x( 'e.g. %s', 'example price', 'woocommerce-subscriptions' ), wc_format_localized_price( '9.90' ) ),
 					'description'       => __( 'Optionally include an amount to be charged at the outset of the subscription. The sign-up fee will be charged immediately, even if the product has a free trial or the payment dates are synced.', 'woocommerce-subscriptions' ),
 					'desc_tip'          => true,
 					'type'              => 'text',
--- a/includes/admin/class-wc-subscriptions-admin.php
+++ b/includes/admin/class-wc-subscriptions-admin.php
@@ -316,12 +316,10 @@ class WC_Subscriptions_Admin {
 				?>
 			</label>
 			<span class="wrap">
-				<input type="text" id="_subscription_price" name="_subscription_price" class="wc_input_price wc_input_subscription_price" placeholder="
-				<?php
-				// translators: %s the formatted example price value.
-				echo esc_attr( sprintf( 'e.g. %s', wc_format_localized_price( '5.90' ) ) );
-				?>
-				" step="any" min="0" value="<?php echo esc_attr( wc_format_localized_price( $chosen_price ) ); ?>" />
+				<input type="text" id="_subscription_price" name="_subscription_price" class="wc_input_price wc_input_subscription_price" placeholder="<?php
+					// translators: %s the formatted example price value.
+					echo sprintf( esc_attr_x( 'e.g. %s', 'example price', 'woocommerce-subscriptions' ), wc_format_localized_price( '5.90' ) );
+				?>" step="any" min="0" value="<?php echo esc_attr( wc_format_localized_price( $chosen_price ) ); ?>" />
 				<label for="_subscription_period_interval" class="wcs_hidden_label"><?php esc_html_e( 'Subscription interval', 'woocommerce-subscriptions' ); ?></label>
 				<select id="_subscription_period_interval" name="_subscription_period_interval" class="wc_input_subscription_period_interval wc-enhanced-select">
 				<?php foreach ( wcs_get_subscription_period_interval_strings() as $value => $label ) { ?>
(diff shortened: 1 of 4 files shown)


includes/admin/class-wc-subscriptions-admin.php
```diff
@@ -316,7 +316,10 @@ public static function subscription_pricing_fields() {
 				?>
 			</label>
 			<span class="wrap">
-				<input type="text" id="_subscription_price" name="_subscription_price" class="wc_input_price wc_input_subscription_price" placeholder="<?php echo esc_attr_x( 'e.g. 5.90', 'example price', 'woocommerce-subscriptions' ); ?>" step="any" min="0" value="<?php echo esc_attr( wc_format_localized_price( $chosen_price ) ); ?>" />
+				<input type="text" id="_subscription_price" name="_subscription_price" class="wc_input_price wc_input_subscription_price" placeholder="<?php
+				  // translators: %s the formatted example price value.
+				  echo sprintf( esc_attr_x( 'e.g. %s', 'example price', 'woocommerce-subscriptions' ), str_replace( get_woocommerce_currency_symbol(), '', strip_tags( wc_price( 5.90 ) ) ) );
```
james-allan: Rather than doing a string replace on the currency symbol can we just use the same function as the one below?

```suggestion
				  echo sprintf( esc_attr_x( 'e.g. %s', 'example price', 'woocommerce-subscriptions' ), wc_format_localized_price( '5.90' ) );
```

includes/admin/class-wc-subscriptions-admin.php
```diff
@@ -354,7 +357,8 @@ public static function subscription_pricing_fields() {
 					'class'             => 'wc_input_subscription_intial_price wc_input_subscription_initial_price wc_input_price  short',
 					// translators: %s is a currency symbol / code
 					'label'             => sprintf( __( 'Sign-up fee (%s)', 'woocommerce-subscriptions' ), get_woocommerce_currency_symbol() ),
-					'placeholder'       => _x( 'e.g. 9.90', 'example price', 'woocommerce-subscriptions' ),
+					'placeholder'       => // translators: %s the formatted example price value.
+						sprintf( esc_attr_x( 'e.g. %s', 'example price', 'woocommerce-subscriptions' ), str_replace( get_woocommerce_currency_symbol(), '', strip_tags( wc_price( 9.90 ) ) ) ),
```
james-allan: Same feedback here. 

```suggestion
						sprintf( esc_attr_x( 'e.g. %s', 'example price', 'woocommerce-subscriptions' ), wc_format_localized_price( '9.90' ) ),
```

templates/admin/html-variation-price.php
```diff
@@ -15,50 +15,36 @@
...
-		<select name="variable_subscription_trial_period[<?php echo esc_attr( $loop ); ?>]" class="wc_input_subscription_trial_period wc-enhanced-select">
-		<?php foreach ( wcs_get_available_time_periods() as $key => $value ) : ?>
-			<option value="<?php echo esc_attr( $key ); ?>" <?php selected( $key, WC_Subscriptions_Product::get_trial_period( $variation_product ) ); ?>><?php echo esc_html( $value ); ?></option>
-		<?php endforeach; ?>
-		</select>
-	</p>
-</div>
 <div class="variable_subscription_pricing variable_subscription_pricing_2_3 show_if_variable-subscription" style="display: none">
-	<p class="form-row form-row-first form-field show_if_variable-subscription _subscription_price_field">
+	<p class="form-row dimensions_field form-row-first form-field show_if_variable-subscription _subscription_price_field">
```
james-allan: It looks like you're adding the `dimensions_field` class here to apply the existing styling for dimension fields, giving us the same flex display and spacing. eg

Screenshot 2025-03-12 at 3 52 21 pm

I wonder if using `dimensions_field` for our pricing fields is the best approach. Third parties may rely on this class specifically for dimension fields, reusing it here could lead to unexpected styling or behavior.

I think we need to replicate the same styles and apply them to an existing class, like `_subscription_price_field`, to keep things clear and avoid potential conflicts.

james-allan (changes requested):
Hey @alefesouza. It's looking pretty neat so far. I left a couple of comments. Let me know what you think. :) 

Just noting that the PHPCS is throwing the following errors on this PR:

Screenshot 2025-03-12 at 4 44 35 pm

Let me know if you need any help getting those resolved.

includes/admin/class-wc-subscriptions-admin.php
alefesouza: Thanks, I searched for a function that to do this but couldn't find, fixed on https://github.com/Automattic/woocommerce-subscriptions-core/pull/807/commits/bd8a70a9a302329f3e0d4fb930bd1c784052ee70.

includes/admin/class-wc-subscriptions-admin.php
alefesouza: Fixed on https://github.com/Automattic/woocommerce-subscriptions-core/pull/807/commits/bd8a70a9a302329f3e0d4fb930bd1c784052ee70.

templates/admin/html-variation-price.php
alefesouza: Fixed on https://github.com/Automattic/woocommerce-subscriptions-core/pull/807/commits/94aa04689425a8928c96ce4adcaa588c6da3c7cf.

james-allan (changes requested):
Thanks for fixing those issues @alefesouza. 

I created the following list based on the issue description:

- [x] the placeholders for the price fields now correctly displays in the stores currency format. 

| $ | € |
|--------|--------|
| Screenshot 2025-03-27 at 2 25 10 pm | Screenshot 2025-03-27 at 2 26 17 pm | 


⚠️  I noticed the placeholder on a simple subscription was incorrect: 

Screenshot 2025-03-27 at 2 27 01 pm

I think this extra spacing is coming from the HTML. I'll leave an inline comment about it. 

- [x] All dropdowns in the editor should use the same text size, 14px, while we are using 12px.
- [x] We should follow the same field order like simple products: Subscription price, Stop renewing after, Sign-up fee, Free trial.

<p align="center">
Screenshot 2025-03-27 at 2 34 03 pm
</br>
<sup>Subscription price > Stop renewing after (length) > Sign-up fee > Free trial</sup>
</p> 

- [x] There should be a 12px margin between “Free trial” and “Subscription price” fields. 


--- 

I noticed while testing that the (?) help icons have an inconsistent size. This happens on `trunk` too so feel free to punt this to a different issue if you want to follow up on it. 

Screenshot 2025-03-27 at 2 37 26 pm

1.4em appears to be the correct size based on the other elements. It's just the **Stop renewing after** and **Free trial** ones with the 1.2em size.

includes/admin/class-wc-subscriptions-admin.php
```diff
@@ -316,7 +316,12 @@ public static function subscription_pricing_fields() {
 				?>
 			</label>
 			<span class="wrap">
-				<input type="text" id="_subscription_price" name="_subscription_price" class="wc_input_price wc_input_subscription_price" placeholder="<?php echo esc_attr_x( 'e.g. 5.90', 'example price', 'woocommerce-subscriptions' ); ?>" step="any" min="0" value="<?php echo esc_attr( wc_format_localized_price( $chosen_price ) ); ?>" />
+				<input type="text" id="_subscription_price" name="_subscription_price" class="wc_input_price wc_input_subscription_price" placeholder="
+				<?php
+				// translators: %s the formatted example price value.
+				echo esc_attr( sprintf( 'e.g. %s', wc_format_localized_price( '5.90' ) ) );
+				?>
```
james-allan: There's a couple of issues with the code here: 

1. All the whitespace after `placeholder="` and `<?php` is being rendered in the HTML. 
2. This should be translated like before. 

```php
// translators: %s the formatted example price value.
echo sprintf( esc_attr_x( 'e.g. %s', 'example price', 'woocommerce-subscriptions' ), wc_format_localized_price( '5.90' ) );
```

alefesouza: Fixed on https://github.com/Automattic/woocommerce-subscriptions-core/pull/807/commits/0aaa1807824b07a13e6b80bf802894e674ab8792.

templates/admin/html-variation-price.php
```diff
@@ -15,50 +15,33 @@
...
-</div>
 <div class="variable_subscription_pricing variable_subscription_pricing_2_3 show_if_variable-subscription" style="display: none">
-	<p class="form-row form-row-first form-field show_if_variable-subscription _subscription_price_field">
+	<p class="form-row multiple_fields form-row-first form-field show_if_variable-subscription _subscription_price_field">
 		<label for="variable_subscription_price[<?php echo esc_attr( $loop ); ?>]">
 			<?php
 			// translators: placeholder is a currency symbol / code
 			printf( esc_html__( 'Subscription price (%s)', 'woocommerce-subscriptions' ), esc_html( get_woocommerce_currency_symbol() ) );
 			?>
+			<?php echo wcs_help_tip( _x( 'Choose the subscription price, billing interval and period.', 'woocommerce-subscriptions' ) ); ?>
```
james-allan:
```suggestion
			<?php echo wcs_help_tip( __( 'Choose the subscription price, billing interval and period.', 'woocommerce-subscriptions' ) ); ?>
```

The second arg for `_x()` is supposed to contain the context of the string. eg `'used in help tip'`. 

Our other use of this string ([here](https://github.com/automattic/woocommerce-subscriptions-core/blob/8.1.0/includes/admin/class-wc-subscriptions-admin.php#L298)) just uses a standard `__()` call so I think that's appropriate here too.

templates/admin/html-variation-price.php
```diff
@@ -73,3 +56,30 @@
 		</select>
 	</p>
 </div>
+<div class="variable_subscription_trial variable_subscription_pricing_2_3 show_if_variable-subscription variable_subscription_trial_sign_up" style="display: none">
+	<p class="form-row form-row-first form-field show_if_variable-subscription sign-up-fee-cell">
+		<label for="variable_subscription_sign_up_fee[<?php echo esc_attr( $loop ); ?>]">
+			<?php printf( esc_html__( 'Sign-up fee (%s)', 'woocommerce-subscriptions' ), esc_html( get_woocommerce_currency_symbol() ) ); ?>
+			<?php echo wcs_help_tip( _x( 'Optionally include an amount to be charged at the outset of the subscription. The sign-up fee will be charged immediately, even if the product has a free trial or the payment dates are synced.', 'woocommerce-subscriptions', 'woocommerce-subscriptions' ) ); ?>
```
james-allan:
```suggestion
			<?php echo wcs_help_tip( __( 'Optionally include an amount to be charged at the outset of the subscription. The sign-up fee will be charged immediately, even if the product has a free trial or the payment dates are synced.', 'woocommerce-subscriptions' ) ); ?>
```

Similar to the comment above, this `_x()` did have the second param, however it wasn't offering any further context other than the text domain. In the other place where there's a similar string ([here](https://github.com/automattic/woocommerce-subscriptions-core/blob/8.1.0/includes/admin/class-wc-subscriptions-admin.php#L358)) we just use a `__()`.

includes/admin/class-wc-subscriptions-admin.php
```diff
@@ -316,7 +316,7 @@ public static function subscription_pricing_fields() {
 				?>
 			</label>
 			<span class="wrap">
-				<input type="text" id="_subscription_price" name="_subscription_price" class="wc_input_price wc_input_subscription_price" placeholder="<?php echo esc_attr_x( 'e.g. 5.90', 'example price', 'woocommerce-subscriptions' ); ?>" step="any" min="0" value="<?php echo esc_attr( wc_format_localized_price( $chosen_price ) ); ?>" />
+				<input type="text" id="_subscription_price" name="_subscription_price" class="wc_input_price wc_input_subscription_price" placeholder="<?php echo esc_attr_x( 'e.g.', 'example price', 'woocommerce-subscriptions' ); ?> <?php echo esc_attr( wc_format_localized_price( '5.90' ) ); ?>" step="any" min="0" value="<?php echo esc_attr( wc_format_localized_price( $chosen_price ) ); ?>" />
```
james-allan:
```suggestion
				<?php // Translators: %s: formatted example price value. ?>
				<input type="text" id="_subscription_price" name="_subscription_price" class="wc_input_price wc_input_subscription_price" placeholder="<?php printf( esc_attr_x( 'e.g. %s', 'example price', 'woocommerce-subscriptions' ), wc_format_localized_price( '5.95' ) ); ?>" step="any" min="0" value="<?php echo esc_attr( wc_format_localized_price( $chosen_price ) ); ?>" />
```

**Note:** If you accept this commit, just check the spacing and formatting because it might not be correct. 

I hadn't noticed this before but this line essentially has 'e.g.' translated on it's own and then the price (5.90) is appended to it. 

Whereas below it's handled by 'e.g. %s' being translated all together and then %s is replaced via `sprintf()`. 

We should try to remain consistant between the two. The second approach of translating the whole string/sentence together is more correct and inline with [best practices](https://codex.wordpress.org/I18n_for_WordPress_Developers#Best_Practices). ie it gives langues the opportunity to reorder the terms. For example: "5.90 e.g.". So this commit suggestion uses the same format `printf( esc_attr_x( 'e.g. %s', ... ) )`

templates/admin/html-variation-price.php
```diff
@@ -15,50 +15,33 @@
...
 
-		<label for="variable_subscription_period_interval[<?php echo esc_attr( $loop ); ?>]" class="wcs_hidden_label"><?php esc_html_e( 'Billing interval:', 'woocommerce-subscriptions' ); ?></label>
-		<select name="variable_subscription_period_interval[<?php echo esc_attr( $loop ); ?>]" class="wc_input_subscription_period_interval wc-enhanced-select">
-		<?php foreach ( wcs_get_subscription_period_interval_strings() as $key => $value ) : ?>
-			<option value="<?php echo esc_attr( $key ); ?>" <?php selected( $key, WC_Subscriptions_Product::get_interval( $variation_product ) ); ?>><?php echo esc_html( $value ); ?></option>
-		<?php endforeach; ?>
-		</select>
+		<span class="wrap">
+			<input type="text" class="wc_input_price wc_input_subscription_price" name="variable_subscription_price[<?php echo esc_attr( $loop ); ?>]" value="<?php echo esc_attr( wc_format_localized_price( WC_Subscriptions_Product::get_regular_price( $variation_product ) ) ); ?>" placeholder="<?php echo esc_attr_x( 'e.g.', 'example price', 'woocommerce-subscriptions' ); ?> <?php echo esc_attr( wc_format_localized_price( '9.90' ) ); ?>">
```
james-allan:
```suggestion
			<?php // Translators: %s: formatted example price value. ?>
			<input type="text" class="wc_input_price wc_input_subscription_price" name="variable_subscription_price[<?php echo esc_attr( $loop ); ?>]" value="<?php echo esc_attr( wc_format_localized_price( WC_Subscriptions_Product::get_regular_price( $variation_product ) ) ); ?>" placeholder="<?php printf( esc_attr_x( 'e.g. %s', 'example price', 'woocommerce-subscriptions' ), wc_format_localized_price( '9.90' ) ); ?>">
```

Same commit suggestion here - using the same format `printf( esc_attr_x( 'e.g. %s', ... ) )`

templates/admin/html-variation-price.php
```diff
@@ -73,3 +56,30 @@
...
+	<p class="form-row form-row-first form-field show_if_variable-subscription sign-up-fee-cell">
+		<label for="variable_subscription_sign_up_fee[<?php echo esc_attr( $loop ); ?>]">
+			<?php printf( esc_html__( 'Sign-up fee (%s)', 'woocommerce-subscriptions' ), esc_html( get_woocommerce_currency_symbol() ) ); ?>
+			<?php echo wcs_help_tip( _x( 'Optionally include an amount to be charged at the outset of the subscription. The sign-up fee will be charged immediately, even if the product has a free trial or the payment dates are synced.', 'woocommerce-subscriptions', 'woocommerce-subscriptions' ) ); ?>
+		</label>
+		<input type="text" class="wc_input_price wc_input_subscription_intial_price wc_input_subscription_initial_price" name="variable_subscription_sign_up_fee[<?php echo esc_attr( $loop ); ?>]" value="<?php echo esc_attr( wc_format_localized_price( WC_Subscriptions_Product::get_sign_up_fee( $variation_product ) ) ); ?>" placeholder="<?php echo esc_attr_x( 'e.g.', 'example price', 'woocommerce-subscriptions' ); ?> <?php echo esc_attr( wc_format_localized_price( '9.90' ) ); ?>">
```
james-allan, line 66:
```suggestion
		<?php // Translators: %s: formatted example price value. ?>
		<input type="text" class="wc_input_price wc_input_subscription_intial_price wc_input_subscription_initial_price" name="variable_subscription_sign_up_fee[<?php echo esc_attr( $loop ); ?>]" value="<?php echo esc_attr( wc_format_localized_price( WC_Subscriptions_Product::get_sign_up_fee( $variation_product ) ) ); ?>" placeholder="<?php printf( esc_attr_x( 'e.g. %s', 'example price', 'woocommerce-subscriptions' ), wc_format_localized_price( '9.90' ) ); ?>">
```

Similar suggestion here using `printf( esc_attr_x( 'e.g. %s', ... ) )`

james-allan (changes requested):
Hey Alefe,

Sorry for another set of review comments. I've left a couple of suggestions in line, let me know what you think. :)

barryhughes:
I think we still have some inconsistent help icon sizes. Both these screenshots are from the variations tab:

<div align="center"></div>

<div align="center"></div>

The second set in particular might be 'inherited', so I'm not too sure how much control we can exercise here, but I wanted to note it.

barryhughes:
Changes for simple subscriptions look great!

<div align="center"> </div>

I notice we don't have a tooltip for the Sale Price field, but wonder if that would be useful (especially in the context of *variable* subscriptions—but I'll add a further and more specific note about that). Also, if I open the sale price scheduling field, the positioning of the help icon feels a little 'off' (all the othe help icons are to the right-hand side of the fields):

<div align="center"></div>

barryhughes:
For *variable* subscriptions, the sale price field doesn't include a description of the interval:

<div align="center"></div>

That is, if you compare it to the sale price for simple subscriptions, we see that some text is added to indicate if the sale price will be applied every day, week, month, etc. That is missing here. Should it be added? Or, following my earlier note, should we add a helpful tooltip to explain how this works?

includes/admin/class-wc-subscriptions-admin.php
```diff
@@ -316,7 +316,8 @@ public static function subscription_pricing_fields() {
 				?>
 			</label>
 			<span class="wrap">
-				<input type="text" id="_subscription_price" name="_subscription_price" class="wc_input_price wc_input_subscription_price" placeholder="<?php echo esc_attr_x( 'e.g. 5.90', 'example price', 'woocommerce-subscriptions' ); ?>" step="any" min="0" value="<?php echo esc_attr( wc_format_localized_price( $chosen_price ) ); ?>" />
+				<?php // Translators: %s: formatted example price value. ?>
+				<input type="text" id="_subscription_price" name="_subscription_price" class="wc_input_price wc_input_subscription_price" placeholder="<?php printf( esc_attr_x( 'e.g. %s', 'example price', 'woocommerce-subscriptions' ), esc_attr( wc_format_localized_price( '5.90' ) ) ); ?>" step="any" min="0" value="<?php echo esc_attr( wc_format_localized_price( $chosen_price ) ); ?>" />
```
barryhughes: This is certainly a nit, so feel free to skip, but I wonder if this would be very marginally better (that is, pass the final result through `esc_attr()` *once* instead of passing each component through separately):

```suggestion
				<input type="text" id="_subscription_price" name="_subscription_price" class="wc_input_price wc_input_subscription_price" placeholder="<?php echo esc_attr( sprintf( _x( 'e.g. %s', 'example price', 'woocommerce-subscriptions' ), wc_format_localized_price( '5.90' ) ) ); ?>" step="any" min="0" value="<?php echo esc_attr( wc_format_localized_price( $chosen_price ) ); ?>" />
```

includes/admin/class-wc-subscriptions-admin.php
```diff
@@ -354,7 +355,8 @@ public static function subscription_pricing_fields() {
 					'class'             => 'wc_input_subscription_intial_price wc_input_subscription_initial_price wc_input_price  short',
 					// translators: %s is a currency symbol / code
 					'label'             => sprintf( __( 'Sign-up fee (%s)', 'woocommerce-subscriptions' ), get_woocommerce_currency_symbol() ),
-					'placeholder'       => _x( 'e.g. 9.90', 'example price', 'woocommerce-subscriptions' ),
+					'placeholder'       => // translators: %s the formatted example price value.
```
barryhughes, line 358: Similar to an earlier comment, might it be marginally better to escape last? Here we escape the translatable string, but the output of `wc_format_localized_price()` is injected after that, as part of the `sprintf`. 

Escaping last would prevent any accidents that might let the localized price escape the actual HTML attribute.

includes/admin/class-wc-subscriptions-admin.php
```diff
@@ -354,7 +355,8 @@ public static function subscription_pricing_fields() {
...
+						sprintf( esc_attr_x( 'e.g. %s', 'example price', 'woocommerce-subscriptions' ), wc_format_localized_price( '5.90' ) ),
```
barryhughes: Similar to an earlier comment, might it be marginally better to escape last? Here we escape the translatable string, but the output of `wc_format_localized_price()` is injected after that, as part of the `sprintf`. 

Escaping last would prevent any accidents that might let the localized price escape the actual HTML attribute.

templates/admin/html-variation-price.php
```diff
@@ -15,50 +15,34 @@
...
 
-		<label for="variable_subscription_period_interval[<?php echo esc_attr( $loop ); ?>]" class="wcs_hidden_label"><?php esc_html_e( 'Billing interval:', 'woocommerce-subscriptions' ); ?></label>
-		<select name="variable_subscription_period_interval[<?php echo esc_attr( $loop ); ?>]" class="wc_input_subscription_period_interval wc-enhanced-select">
-		<?php foreach ( wcs_get_subscription_period_interval_strings() as $key => $value ) : ?>
-			<option value="<?php echo esc_attr( $key ); ?>" <?php selected( $key, WC_Subscriptions_Product::get_interval( $variation_product ) ); ?>><?php echo esc_html( $value ); ?></option>
-		<?php endforeach; ?>
-		</select>
+		<span class="wrap">
+			<?php // Translators: %s: formatted example price value. ?>
+			<input type="text" class="wc_input_price wc_input_subscription_price" name="variable_subscription_price[<?php echo esc_attr( $loop ); ?>]" value="<?php echo esc_attr( wc_format_localized_price( WC_Subscriptions_Product::get_regular_price( $variation_product ) ) ); ?>" placeholder="<?php printf( esc_attr_x( 'e.g. %s', 'example price', 'woocommerce-subscriptions' ), esc_attr( wc_format_localized_price( '9.90' ) ) ); ?>">
```
barryhughes: I won't repeat each time, but this is another case (specifically thinking of the `printf`) where could escape once, as the last operation before echoing, instead of escaping each individual component.

barryhughes (approved):
This is looking great. I left some notes, but none are deal-breakers. Let me know if I can clarify anything 👍🏼 

Pre-emptively approving, as any remaining changes are small in nature (though I'm also more than happy to give it a final pass if you'd like).

templates/admin/html-variation-price.php
alefesouza: Fixed on https://github.com/Automattic/woocommerce-subscriptions-core/pull/807/commits/ce3bdaf566b50a9696525c1248b0b4e287b93dc8.

includes/admin/class-wc-subscriptions-admin.php
alefesouza: Fixed on https://github.com/Automattic/woocommerce-subscriptions-core/pull/807/commits/ce3bdaf566b50a9696525c1248b0b4e287b93dc8.

includes/admin/class-wc-subscriptions-admin.php
alefesouza: Fixed on https://github.com/Automattic/woocommerce-subscriptions-core/pull/807/commits/ce3bdaf566b50a9696525c1248b0b4e287b93dc8.

alefesouza:
@barryhughes the `Schedule` section is related to WC Core and happens even with Subscriptions disabled.

PR #810, opened 2025-03-11
Load HTML in product name on shipping rates

alefesouza:
Fixes woocommerce/woocommerce-subscriptions#4128

## Description

> This PR fixes the #762 reverted PR by not using the `woocommerce_cart_item_name` filter, I chose to only remove it because `wp_kses_post` already removes any unsafe HTML tag.

This PR fixes raw HTML being shown to the user on the shipping rates section of the classic cart and checkout page when the product name contains HTML.

| Before | After |
|-|-|
| Screenshot 2025-01-10 at 9 48 11 PM | Screenshot 2025-01-10 at 9 48 28 PM |

## How to test this PR

1. Create a subscription with HTML in the name.
2. Add this subscription to the cart.
3. Add a normal product to the cart.
4. Access the classic cart or checkout page.
5. The HTML in the product name for the shipping rate is not loaded.

## Product impact

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -9,6 +9,7 @@
 * Fix - Updated subscription email item table template to align with WooCommerce 9.7 email improvements.
 * Fix - Prevent PHP warning on cart page shipping method updates by removing unused method: maybe_restore_shipping_methods.
 * Fix - Ensure the order_awaiting_payment session arg is restored when loading a renewal cart from the session to prevent duplicate orders.
+* Fix - Correctly load product names with HTML on the cart and checkout shipping rates.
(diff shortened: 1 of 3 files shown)


mattallan (approved):
Thanks @alefesouza for returning to this one and fixing it up after it was reverted 👍 

| Before | After |
|--------|--------|
| image | image | 

Just making a note that the previous fatal error that resulted in us reverting this change was reproducible using the following code snippet:

```php
add_filter( 'woocommerce_cart_item_name', function( $content, $cart_item, $cart_item_key ) {
	return $content;
}, 9, 3 );
```

Since this change no longer uses a filter, it's not possible for this fatal to occur again :) Approving this!

PR #816, opened 2025-03-18
Removed support for WC version <= 4.4

### LINKED ISSUE #151: Remove files/code that contain pre WC 3.0 support
## Description

The subscriptions core code has a fair bit of code that adds compatibility for versions of WooCommerce that we no longer support in WC Subscriptions and WC Payments.

Minimum WC version requirements in our plugins that use Subscriptions Core:
 - WC Payments: 5.2
 - WC Subscriptions: 4.4

The requirements of this issue is to:
 - [ ] Delete any code that is calling `wcs_is_woocommerce_pre( $version )` where the version is before 4.4.
 - [ ] Delete any uses of `wcs_add_woocommerce_dependent_action()` where `$woocommerce_version` is before 4.4
 - [ ] Remove/deprecate legacy class files that still exist:
   - `class-wc-product-variable-subscription-legacy.php`
   - `class-wc-subscription-legacy.php`
   - `class-wcs-array-property-post-meta-black-magic.php`

## Product impact

- Does this feature affect WooCommerce Subscriptions? no
- Does this feature affect WooCommerce Payments? no

Since none of our plugins using subscriptions core library support pre 4.4, these changes should not impact either product.

leonardola:
Fixes #151 

## Description
This PR removes code that was being used to keep backward compatibility with WC version <= 4.4 as we don't support such old WC versions anymore.

## How to test this PR
As the changes happened everywhere in the code it makes sense to smoke test the whole plugin IMHO

## Product impact


includes/admin/class-wc-subscriptions-admin.php
```diff
@@ -136,7 +136,6 @@ public static function init() {
 
 		// Change text for when order items cannot be edited
 		wcs_add_woocommerce_dependent_action( 'woocommerce_admin_order_totals_after_total', array( __CLASS__, 'maybe_attach_gettext_callback' ), '4.0.0', '>' );
```
barryhughes: If we no longer need to support 4.0.0 or earlier, can we also simplify this line?

```suggestion
		add_action( 'woocommerce_admin_order_totals_after_total', array( __CLASS__, 'maybe_attach_gettext_callback' ) );
```

It feels like we no longer need the added indirection that `wcs_add_woocommerce_dependent_action()` and `WCS_Dependent_Hook_Manager` provide in this case.

includes/admin/class-wc-subscriptions-admin.php
```diff
@@ -917,7 +911,7 @@ public static function enqueue_styles_scripts() {
 
 			$script_params['ajaxLoaderImage'] = WC()->plugin_url() . '/assets/images/ajax-loader.gif';
 			$script_params['ajaxUrl']         = admin_url( 'admin-ajax.php' );
-			$script_params['isWCPre24']       = var_export( wcs_is_woocommerce_pre( '2.4' ), true );
+			$script_params['isWCPre24']       = false;
```
barryhughes: Not essential we address this here, but on seeing this change I wondered if this var is even used (just in case we needed a corresponding change inside `admin/admin.js`). Actually, though, it seems like any uses of `isWCPre24` and `isWCPre23` were removed around 9 years ago:

https://github.com/woocommerce/woocommerce-subscriptions/commit/9a237442798208e9ffad7710c35447afd13e4cff#diff-0dec7a408ff08c7c6200cd2b92cdc10704d7b68c3066c70f9e3c89ebadd6a7eb

Given that, it seems like it would be safe to simply remove it—though there is, I guess, a very small back-compat risk (third party code may still reference it—but that seems extremely unlikely).

That said, I think the same thing may be true of various script params added here, and they are also 'harmless', so we can park this and resolve via some future piece of work.

includes/admin/class-wcs-admin-meta-boxes.php
```diff
@@ -42,10 +42,6 @@ public function __construct() {
 		add_action( 'woocommerce_order_action_wcs_create_pending_renewal', array( __CLASS__, 'create_pending_renewal_action_request' ), 10, 1 );
 		add_action( 'woocommerce_order_action_wcs_create_pending_parent', array( __CLASS__, 'create_pending_parent_action_request' ), 10, 1 );
 
-		if ( wcs_is_woocommerce_pre( '3.2' ) ) {
-			add_filter( 'woocommerce_resend_order_emails_available', array( __CLASS__, 'remove_order_email_actions' ), 0, 1 );
```
barryhughes, line 46: In a sense, this 'orphans' the `remove_order_email_actions()` method (it's now just sitting unused).

For back-compat reasons we may not want to strip it out entirely, but what do you think of supporting this change with some updates to that method?

- We could mark it as `@deprecated`.
- We could add a `_deprecated_function()` call.
- We could indicate it is marked for removal in an upcoming release.

In the long term, this would help us to keep the codebase clean by removing unused methods (while also giving a reasonable notice period).

includes/class-wc-subscription.php
```diff
@@ -470,7 +466,7 @@ public function update_status( $new_status, $note = '', $manual = false ) {
 						$end_date = $this->calculate_date( 'end_of_prepaid_term' );
 
 						// If there is no future payment and no expiration date set, or the end date is before now, the customer has no prepaid term (this shouldn't be possible as only active subscriptions can be set to pending cancellation and an active subscription always has either an end date or next payment), so set the end date and cancellation date to now
-						if ( 0 == $end_date || wcs_date_to_time( $end_date ) < current_time( 'timestamp', true ) ) {
+						if ( 0 === $end_date || wcs_date_to_time( $end_date ) < time() ) {
```
barryhughes: I'm all for strictness, but introducing this now *could* cause breakages in some cases, such as if some third party code does something like this (simplified):

```php
add_filter( 
	'woocommerce_subscription_calculated_end_of_prepaid_term_date',
	function( string $date, WC_Subscription $subscription ) {
		if ( $subscription->hypothetical_concern ) {
			return '0';
		}

		return $date;
	},
	10,
	2
);
```

Arguably what they are doing is wrong, or could just be a misunderstanding, but the reason I raise it here is, should we suddenly start interpreting that differently?

If I'm overthinking it/being over zealous, that's totally fine. Wanted to flag it though, as if we can avoid disruption for our merchants then, imho, we should.

includes/class-wc-subscriptions-core-plugin.php
```diff
@@ -226,11 +226,6 @@ public function init_version_dependant_classes() {
 		if ( class_exists( 'WC_Abstract_Privacy' ) ) {
 			new WCS_Privacy();
 		}
-
-		// Loads Subscriptions support for the WooCommerce Navigation feature. This feature was removed in WC 9.3.
-		if ( wcs_is_woocommerce_pre( '9.3' ) ) {
```
barryhughes: Double-checking that we are intentionally removing this ... 9.3 itself is only 7 months old or so, and don't we still support as far back as [8.7.1](https://github.com/woocommerce/woocommerce-subscriptions/blob/4cb1607f5ff23714dc768851711b975b6fce16ce/woocommerce-subscriptions.php#L11)?

To be clear, I think something like an L-2 policy would be fine for us, it's just that isn't something we've firmly settled on, and it seems a little out-of-sync with the PR goals (which is removing code pertaining to *WC 4.4* and earlier)

includes/class-wc-subscriptions-renewal-order.php
```diff
@@ -96,7 +96,6 @@ public static function maybe_record_subscription_payment( $order_id, $orders_old
 		$order_needed_payment = in_array( $orders_old_status, apply_filters( 'woocommerce_valid_order_statuses_for_payment', array( 'pending', 'on-hold', 'failed' ), $order ) );
 
 		if ( $order_completed && $order_needed_payment ) {
-			// Prior to WC 3.0, we need to update the post date (i.e. the date created) to have a reliable representation of the paid date (both because it was in GMT and because it was always set). That's not needed in WC 3.0, but some plugins and store owners still rely on it being updated, so we want to make it possible to update it with 3.0 also.
```
barryhughes: Minor thing, but isn't it better to keep this explanation (if in the future we or other maintainers are wondering what the utility of the filter hook is, might it be helpful)?

includes/class-wcs-cart-renewal.php
```diff
@@ -156,13 +144,7 @@ public function setup_hooks() {
...
 	 */
 	public function attach_dependant_callbacks() {
-
-		if ( wcs_is_woocommerce_pre( '3.0' ) ) {
-			add_action( 'woocommerce_add_order_item_meta', array( &$this, 'add_order_item_meta' ), 10, 2 );
-			add_action( 'woocommerce_add_subscription_item_meta', array( &$this, 'add_order_item_meta' ), 10, 2 );
-		} else {
-			add_action( 'woocommerce_checkout_create_order_line_item', array( &$this, 'add_order_line_item_meta' ), 10, 3 );
-		}
+		add_action( 'woocommerce_checkout_create_order_line_item', array( &$this, 'add_order_line_item_meta' ), 10, 3 );
```
barryhughes: Minor/optional, I realize this pre-dates your change, but we could drop the unnecessary `&`, since an object is always passed by reference:

```suggestion
		add_action( 'woocommerce_checkout_create_order_line_item', array( $this, 'add_order_line_item_meta' ), 10, 3 );
```

includes/class-wcs-cart-renewal.php
```diff
@@ -798,7 +780,7 @@ public function renewal_coupon_data( $data, $code ) {
 
 				// Tweak the coupon data for renewal coupons
 				if ( $coupon_code == $code ) {
-					$expiry_date_property = wcs_is_woocommerce_pre( '3.0' ) ? 'expiry_date' : 'date_expires';
+					$expiry_date_property = 'date_expires';
```
barryhughes: With this change, `$expiry_date_property` can only ever be `(string) 'date_expires'`. Why not drop the variable completely, and use the string literal as the array key (a few lines down, where we define `$renewal_coupon_overrides`?

changelog.txt
```diff
@@ -27,6 +27,7 @@
 * Fix - Prevent empty strings being saved in related orders cache ID meta when backfilling order data to the WP Posts table.
 * Fix - Correctly load product names with HTML on the cart and checkout shipping rates.
 * Dev - Fix Node version mismatch between package.json and .nvmrc (both are now set to v16.17.1).
+* Dev - Removed support for WC version <= 4.4.
```
barryhughes: I think we'll need to move this up to the top section.

barryhughes (commented):
Nice—this is a lot of work, thank you!

- I left some comments inline, mostly (low risk) back-compat concerns, and perhaps not worth worrying about, but I wanted to flag so we could at least pause and consider them.
- Related to which, the complete removal of the `WC_Product_Variable_Subscription_Legacy` class without a deprecation period feels a little concerning. 
    - If there is third party/custom code out there doing something like `( new WC_Product_Variable_Subscription_Legacy( 123 ) )->get_price();` well, it works just fine, even with our current stable release. Is it appropriate to break that without notice?
    - On the other hand, `WC_Subscription_Legacy` seems fundamentally problematic at this point and will break things if loaded, so removing this one seems fine. *Perhaps* that's an indication that these legacy classes aren't used (though it's hard to be sure ... all this really tells us is that *this* class isn't being used— it doesn't guarantee that `WC_Product_Variable_Subscription_Legacy` is not being used).
- I still need to do more holistic testing per the testing instructions, but wanted to share this feedback early (and will continue testing once you've had a chance to take a look at that).

includes/admin/class-wc-subscriptions-admin.php
leonardola: Fixed with dff1a5df844fb6fc4caf45c1865a86f2c05a946b

includes/admin/class-wc-subscriptions-admin.php
leonardola: Fixed with 4752b3e019b2d65f7cdb3114733d0d6def236485

includes/admin/class-wcs-admin-meta-boxes.php
leonardola: Fixed with 45f3e90b52fd4515dd77c524c63ef5083647f009

includes/class-wc-subscription.php
leonardola: Fixed with 384a5de40ce430b947fdae6bc2c1b491ae47d6e9

includes/class-wc-subscriptions-core-plugin.php
leonardola: Reverted with ee936fbb54350fe60ac4e3fbbd213c6f2befd2e9

includes/class-wc-subscriptions-renewal-order.php
leonardola: Fixed with 81c7c2120ecb419ab71e3447a0242c8d84062565

includes/class-wcs-cart-renewal.php
leonardola: Fixed with 844fcd004ef26af062d5a87fede9fa1e499af368

includes/class-wcs-cart-renewal.php
leonardola: Fixed with f245abb86d4b6c804cc2877eff4505efdccc8a73

changelog.txt
leonardola: Fixed when merging trunk into the branch 766991d8ea675df182b6055a584591bb8b4564f3

leonardola:
@barryhughes This PR is ready for another round of review

includes/admin/class-wc-subscriptions-admin.php
```diff
@@ -2016,10 +2009,9 @@ public static function insert_setting_after( &$settings, $insert_after_setting_i
 	 * @param array $settings The list of settings
 	 */
 	public static function add_guest_checkout_setting_note( $settings ) {
-		$is_wc_pre_3_4_0 = wcs_is_woocommerce_pre( '3.4.0' );
-		$current_filter  = current_filter();
+		$current_filter = current_filter();
```
barryhughes, line 2013: (1/2) *Definitely not essential,* but we could spare ourselves an assignment here...

```suggestion
```

includes/admin/class-wc-subscriptions-admin.php
```diff
@@ -2016,10 +2009,9 @@ public static function insert_setting_after( &$settings, $insert_after_setting_i
...
 
-		if ( ( $is_wc_pre_3_4_0 && 'woocommerce_payment_gateways_settings' !== $current_filter ) || ( ! $is_wc_pre_3_4_0 && 'woocommerce_account_settings' !== $current_filter ) ) {
+		if ( 'woocommerce_account_settings' !== $current_filter ) {
```
barryhughes, line 2014: (2/2) And call the function directly:

```suggestion
		if ( 'woocommerce_account_settings' !== current_filter() ) {
```

Imho, this is just as readable, and shaves off some extra lines of code (but again, not strictly needed, either).

includes/admin/class-wcs-admin-meta-boxes.php
```diff
@@ -353,8 +347,11 @@ public static function create_pending_parent_action_request( $subscription ) {
 	 *
 	 * @param array $available_emails
 	 * @since 1.0.0 - Migrated from WooCommerce Subscriptions v2.0
+	 *
+	 * @deprecated
 	 */
 	public static function remove_order_email_actions( $email_actions ) {
+		wcs_deprecated_function( __METHOD__, '8.3.0', 'WCS_Admin_Meta_Boxes::remove_order_email_actions' );
```
barryhughes, line 354: I'm not sure if we cemented this/recorded it anywhere, but I think we recently decided that we should reference Subscriptions Prime versions for things like this, because that will make the most sense to the majority of our users:

```suggestion
		wcs_deprecated_function( __METHOD__, '7.5.0', 'WCS_Admin_Meta_Boxes::remove_order_email_actions' );
```

includes/legacy/class-wc-product-variable-subscription-legacy.php
```diff
@@ -1,441 +0,0 @@
...
- * @package WooCommerce Subscriptions
- * @category Class
- * @since 1.0.0 - Migrated from WooCommerce Subscriptions v2.2.0
- *
- */
-if ( ! defined( 'ABSPATH' ) ) {
-	exit; // Exit if accessed directly
-}
-
-class WC_Product_Variable_Subscription_Legacy extends WC_Product_Variable_Subscription {
```
barryhughes, line 17: Just double-checking you saw my note from [back here](https://github.com/Automattic/woocommerce-subscriptions-core/pull/816#pullrequestreview-2773056924).

We're removing this class completely but it has not officially been deprecated, and it's possible it could be referenced from third party code. Is removing it without a warning/deprecation period wise?

It may be fine, especially if we are confident nobody is using it, but as before I wanted to pause and make sure this is indeed something we're happy to do.

includes/legacy/class-wcs-product-legacy.php
```diff
@@ -30,11 +30,6 @@ public static function init() {
 	 * @since 1.0.0 - Migrated from WooCommerce Subscriptions v2.2.0
 	 */
 	public static function set_product_class( $classname, $product_type, $post_type, $product_id ) {
-
-		if ( wcs_is_woocommerce_pre( '3.0' ) && in_array( $classname, array( 'WC_Product_Subscription', 'WC_Product_Variable_Subscription', 'WC_Product_Subscription_Variation' ) ) ) {
-			$classname .= '_Legacy';
-		}
-
 		return $classname;
```
barryhughes, line 33: This now seems a bit redundant, because we filter `$classname` to only ever return `$classname`. Some options we could think about:

- Removing the corresponding `add_filter()` call from the `init()` method.
- Removing this method entirely.

Or, if that's too risky, we could alternatively:

- Remove the corresponding `add_filter()` call from the `init()` method.
- Deprecate this method and mark for removal in an upcoming release.

barryhughes (commented):
Looks good! Leaving some additional notes, and will continue with smoke-testing.

barryhughes (approved):
Went through a decent % of critical flows, this is feeling pretty solid to me—thanks for working on it!

Hitting approve so you can merge once ready, though I left some comments earlier (mostly concerning optional tweaks). In particular, I think it's worth reconsidering the complete removal of `class-wc-subscription-variable-subscription-legacy.php`, but happy to leave that to your discretion.

barryhughes:

PR #823, opened 2025-03-24
Pass proper parameters to the wcs_get_edit_post_link function, hence …

### LINKED ISSUE #822: [Compatibility] Edit links filter for listed orders by subscription use the wrong type
## Describe the bug
When listing the orders by a subscription in the backend, the logic used to display the edit link for each order is passing a wrong variable type to a WordPress native filter hook. This can cause third parties expecting documented types to fail, even with a fatal error (depending on PHP versions).

We got this reported as a bug for WPML: https://wpml.org/forums/topic/error-when-viewing-wc-subscription-related-orders/

## To Reproduce
1. Install wooCommerce Subscriptions and create a subscription product and have some client buy it.
2. Head to the backend subscriptins list and click on the number of related orders:

Image
3. This will get you to the orders list, filtered by the related subscription, with at least one item.

By now, there will be no error. Now, create a callback for the `get_edit_post_link` filter with heavy dependency on the second parameter being an integer:
```
add_filter( 'get_edit_post_link', function( $url, $post_id ) {
  global $wpdb;
  $item = $wpdb->get_row(
    $wpdb->prepare( 'SELECT * from wp_posts WHERE ID = %d, $post_id" )
  );
  // ...
  return $url;
}, 10, 2 );
```
4. See the error `PHP Notice:  Function wpdb::prepare was called <strong>incorrectly</strong>. Unsupported value type (object).` in your logs.

### Expected behavior
Generating custom edit links should play nicely with WordPress native hooks expected types.

### Actual behavior
Arbitrary PHP notices and errors depending on legitimate callbacks using a value that should be an integer.

## Product impact
- [ ] Does this issue affect WooCommerce Payments?

Additionally, it affects WPML customers: https://wpml.org/forums/topic/error-when-viewing-wc-subscription-related-orders/

decodekult:
…to the get_edit_post_link filter

Fixes #822 

## Description

Make sure that `wcs_get_edit_post_link ` is called with a proper ID instead of an object. This means that the native filter `get_edit_post_link ` will also get a proper integer as second parameter, as expected.

## How to test this PR

1. Create a callback for the `get_edit_post_link` using the `post_id` second parameter ina  way that will produce an error if the parameter is not an integer.
2. Install WooCommer Subscriptions, generate a subscription product and get a customer purchase it.
3. Head to the backend subscriptins list and click on the number of related orders.
4. This will get you to the orders list, filtered by the related subscription, with at least one item. The error expected in 1 should be produced.

## Product impact

- [ ] Will this PR affect WooCommerce Payments?

Additionally, it affects WPML customers: https://wpml.org/forums/topic/error-when-viewing-wc-subscription-related-orders/

### DIFF
--- a/includes/admin/class-wc-subscriptions-admin.php
+++ b/includes/admin/class-wc-subscriptions-admin.php
@@ -1592,7 +1592,7 @@ class WC_Subscriptions_Admin {
 
 			echo '<div class="updated dismiss-subscriptions-search"><p>';
 			// translators: placeholders are opening link tag, ID of sub, and closing link tag
-			printf( esc_html__( 'Showing orders for %1$sSubscription %2$s%3$s', 'woocommerce-subscriptions' ), '<a href="' . esc_url( wcs_get_edit_post_link( $subscription ) ) . '">', esc_html( $subscription->get_order_number() ), '</a>' );
+			printf( esc_html__( 'Showing orders for %1$sSubscription %2$s%3$s', 'woocommerce-subscriptions' ), '<a href="' . esc_url( wcs_get_edit_post_link( $subscription_id ) ) . '">', esc_html( $subscription->get_order_number() ), '</a>' );
 			echo '</p>';
 			printf(
 				'<a href="%1$s" class="notice-dismiss"><span class="screen-reader-text">Dismiss this notice.</span></a>',


decodekult:
Covered by https://github.com/Automattic/woocommerce-subscriptions-core/pull/826

PR #825, opened 2025-03-26
Exploring a free trial limitation

leogermani:
This is an initial exploration to limit Free trials to one per user per product

### The problem

This has been previously descibed [here](https://woocommerce.com/feature-request/limit-free-trials-to-1-per-customer/).

People can abuse Free trials by cancelling and purchasing the same subscription again multiple times. The only way to prevent that is to set the limit to "1 subscription of any status" per user. But that's not a great solution, for the reasons described in the feature request above, but also because it creates additional friction for legit users who want to resubscribe to a subscription that they had in the past but that is currently active.

In these cases, they need to figure out that they have to go to My Account, find that cancelled subscription and click "Resubscribe" from there.

It would be much more straight forward if we limited the free trial to only one per product per user.

### The exploration

This is just an initial exploration, with some amount of AI generated code to help me bootstrap the idea.

This will add a global setting to define whether this limitation is enabled and then:

* Create a record of all the products the user has already used a free trial for and save it as user meta
* Upon checkout, check if the user has already used this Free trial
* If they have, show a notice at the top of the cart or checkout page saying that they need to pay upfront and set the trial period to zero in the cart

### Exploration limitations

I'm pretty sure this is not the best way to do things, but it demonstrates the idea.

I'm not super familiar with all the Woo flows, and I don't believe doing the cart manipulation on the fly in the checkout page is the way to go... would love to have some guidance.

Also, the way I'm manipulating the cart is totally wrong. I'm changing the trial period of the product itself, and not the line item in the cart... but is serves the purposes of testing how the flow looks.

Note: I've tried using `wcs_set_objects_property` with the option to not save, but then the Stripe checkout plugin would do a request to get the cart details and would end up overriding the changes...

Note 2: I also trying doing the logic on the `maybe_set_free_trial` method in `WC_Subscriptions_Synchroniser` class but that didn't work...

## How to test this PR

1. Go to Woocommerce > Settings > Subscriptions
2. Enable "Limit customers to using a free trial only once per product"
3. Set up a Subscription product with a Free trial
4. As a user, buy this subscription
5. Start in a fresh session and try to buy the subscription again with the same user
6. When you get to the cart page, confirm you see a notice at the top saying you've already used the free trial
7. Confirm the cart includes the full price
8. (if you want to test again, you need to update the product trial length because this PR is wrong and is modifying the product)

### DIFF
--- a/includes/admin/class-wc-subscriptions-admin.php
+++ b/includes/admin/class-wc-subscriptions-admin.php
@@ -1248,6 +1248,27 @@ class WC_Subscriptions_Admin {
 					'type' => 'sectionend',
 					'id'   => self::$option_prefix . '_miscellaneous',
 				),
+
+				array(
+					'name'     => __( 'Trial Restrictions', 'woocommerce-subscriptions' ),
+					'type'     => 'title',
+					'desc'     => __( 'Control how free trials can be used by customers.', 'woocommerce-subscriptions' ),
+					'id'       => 'subscription_trial_restrictions',
+				),
+
+				array(
+					'name'     => __( 'One-Time Trials', 'woocommerce-subscriptions' ),
+					'desc'     => __( 'Limit customers to using a free trial only once per product', 'woocommerce-subscriptions' ),
+					'id'       => 'woocommerce_subscriptions_one_time_trials',
+					'type'     => 'checkbox',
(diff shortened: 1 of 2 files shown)


barryhughes:

PR #827, opened 2025-03-30
Display the parent order icon by default on the WooCommerce Orders list table

mattallan:
In Subscriptions 7.3, we introduced a change that no longer displayed the parent order icon in the Order relation column on the WooCommerce > Orders list table.

This change was to increase performance of the WooCommerce > Orders list table, but it has impacted merchants ability to manage their store which we did not anticipate.

This PR returns the parent order icon by default, while still allowing stores to turn it off via a filter to improve the performance when loading the Orders table.

### Testing instructions
1. Purchase a subscription product on your store
2. Navigate to the WooCommerce > Orders table
3. In the Subscriptions Relation Column confirm you see the parent order icon
image

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,8 @@
 *** WooCommerce Subscriptions Core Changelog ***
 
+= 8.1.1 - 2025-xx-xx =
+* Update - Display the subscriptions parent order icon in the WooCommerce Orders list table by default.
+
 = 8.1.0 - 2025-03-24 =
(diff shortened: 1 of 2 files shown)


james-allan (approved):
Changes look good. I confirmed the icon is returned for parent orders

| `trunk` | This branch |
|--------|--------|
| Screenshot 2025-03-31 at 9 37 49 am | Screenshot 2025-03-31 at 9 38 51 am |

PR #833, opened 2025-04-10
Pagination Controls for Frontend Subscriptions View

barryhughes:
Subscribers can visit **My Account ‣ Subscriptions ‣ View Subscription** and, amongst other things, they will be presented with a list of related orders. However, there is currently no limit on how many orders are listed. Imagine a subscription that renews daily: if this runs for a few years, we'll have hundreds (and eventually *thousands)* of related orders. 

- This presents a usability issue:
    - A massive table of hundreds (/thousands) of rows is not easy to consume.
    - It pushes other information out of sight (by default, billing and shipping addresses).
- It presents a performance issue:
    - Our default caching strategy (see [WCS_Related_Order_Store_Cached_CPT](https://github.com/Automattic/woocommerce-subscriptions-core/blob/8.1.1/includes/data-stores/class-wcs-related-order-store-cached-cpt.php)) seems to scale really quite well, by storing several lists of related order IDs for each subscription, and I have not touched this. *If we **do** feel a need to adapt this, or supplement the current caching strategy with something else, it should probably be a project of its own.*
    - The cost of fetching the raw IDs, then, isn't part of what we address here. Hydration of the order objects is.

With that context in mind, in this PR we introduce a new method to the `WC_Subscription` class, making it possible to fetch individual pages of related orders, and we add pagination controls to the frontend related orders list.

<div align="center"></div>

Fixes WOOSUBS-162 | https://github.com/woocommerce/woocommerce-subscriptions/issues/4751

## Further notes

From a local test using a subscription with 339 related orders, I observed the following results (measured using Query Monitor):

| Branch | Object cache | Peak memory (MB) | Total DB queries |
| --- | --- | --- | --- |
| Trunk | off | 198 | 5381 |
| Trunk | on (+primed) | 147 | 3241 |
| PR | off | 131 | 3124 |
| PR | on (+primed) | 85 | 1159 |

💡 Be aware that if you test with enough orders, and with Query Monitor activated, you may run out of memory. So be ready to adjust that `php.ini` file and bump `memory_limit` accordingly (or else reduce the size of the test data set).

**How can this code break?**

Before this change, we passed an array of *all* related order IDs into the `related-orders.php` template whereas, with this change, a maximum of 10 IDs will be passed in. It's possible that, in some cases, our users will have implemented their own pagination (or have some other custom feature that depends on having access to the full range of IDs for some other reason).

I suspect this would be the exception rather than the rule, and if it surfaces we could initially guide them toward restoring the previous behaviour: for example, they could unhook `WC_Subscriptions_Order::get_related_orders_template()` from the `woocommerce_subscription_details_after_subscription_table` action, and replace it with a custom callback that follows the earlier strategy.

Additionally, it was previously the case that the [`woocommerce_subscription_related_orders`](https://github.com/Automattic/woocommerce-subscriptions-core/blob/8.1.1/includes/class-wc-subscription.php#L2116) filter hook would be invoked one or more times (once for each related order type, ie 'renewal', 'switch', etc) whereas with this change, that will no longer happen. Once again, this *could* cause issues in some cases, and the mitigation would be similar to what I described earlier.

## How to test this PR

If you want to look at performance characteristics, I'd recommend using Query Monitor. However, if, like me, you test using a 'real' customer account instead of the main admin account, Query Monitor output won't be available. You can fix this with a small snippet like this:

```php
/**
 * Make Query Monitor available to all users.
 * 
 * ⚠️ Remove after testing!
 */
add_action( 'init', function() {
	$user = wp_get_current_user();

	if ( is_object( $user ) ) {
		$user->add_cap( 'view_query_monitor' );
	}
} );
```

To functionally test pagination, you may also wish to generate a large-ish number of renewal orders. You can do this manually, via the subscription editor, by repeatedly running the *create pending renewal* order action. That could get a little tiresome, though, depending on how many orders you want to generate and so, though not perfect, you could also consider leveraging a bit of code like this:

```php
/**
 * This script can be used to quickly add a stack of renewals
 * to an existing subscription. Can be useful when testing
 * scalability concerns for WooCommerce Subscriptions.
 */
function generate_renewal_order_for_subscription( $subscription_id ) {
	$subscription = new WC_Subscription( $subscription_id );

	// We can't successfully trigger creation of a renewal unless the subscription
	// is active, which may not initially be the case.
	if ( 'active' !== $subscription->get_status() ) {
		$subscription->update_status( 'active' );
	}

	// Try triggering a renewal. This also often sets the subscription to on-hold,
	// so we again change the status.
	WC_Subscriptions_Manager::prepare_renewal( $subscription_id );
	$subscription->update_status( 'active' );
}

// Generate 100 renewals for the subscription (adjust the sub ID as needed).
for ( $i = 1; $i <= 100; $i++ ) {
	generate_renewal_order_for_subscription( 12345 ); // ⚠️ Supply actual subscription ID.
	print '.' . ( ( $i % 10 ) ? '' : PHP_EOL );       // Optional CLI progress dots :-)
}
```

For the testing itself, navigate to a suitable subscription on the frontend: **My Account  Subscriptions ‣ View Subscription**:

1. A maximum of 10 related orders will be rendered.
2. "First", "Previous", "Next" and "Last" pagination links will appear below the table.
3. If they are not useful (for instance, "First" or "Previous" are not useful if we are looking at page 1 of the results) then they still appear, and are still functional, but are styled a little differently by applying the `disabled` class.
4. No ajax magic here. However, the links take advantage of an anchor/ID to keep the customer in broadly the same part of the screen whenever they use the pagination controls.

## Product impact

- [x] Will this PR affect WooCommerce Subscriptions? **Yes → WOOSUBS-162**
- [x] Will this PR affect WooCommerce Payments? **No**
- [x]  Added deprecated functions, hooks or classes to the [spreadsheet](https://docs.google.com/spreadsheets/d/1xw9xszcPMnWsp4C8OKZMsLzZob7tOmWT7qMqmEIq314/edit#gid=0) **None*

### DIFF
--- a/assets/css/view-subscription.css
+++ b/assets/css/view-subscription.css
@@ -103,8 +103,9 @@
 	display: inline-block;
 }
 
-.woocommerce-subscriptions-related-orders-pagination-links .pagination-links a.disabled {
-	visibility: hidden;
+.woocommerce-subscriptions-related-orders-pagination-links .pagination-links a.disabled:hover {
+	cursor: default;
+	text-decoration: none;
 }
 
 .woocommerce-subscriptions-related-orders-pagination-links .pagination-links a .symbol {
@@ -119,7 +120,7 @@
 	direction: rtl;
 }
 
-@media (max-width: 30em) {
+@media ( max-width: 30em ) {
 	.woocommerce-subscriptions-related-orders-pagination-links .pagination-links a {
 		padding: 0.5em 1em;
 	}
--- a/assets/css/view-subscription.css
+++ b/assets/css/view-subscription.css
@@ -90,3 +90,22 @@
 	font-size: 1.4em;
 	text-align: center;
 }
+
+.woocommerce-subscriptions-related-orders-pagination-links {
+	display: flex;
+	flex-direction: row;
+	justify-content: space-between;
+	margin-bottom: 1em;
+}
+
+.woocommerce-subscriptions-related-orders-pagination-links .pagination-links a {
+	display: inline-block;
+}
+
+.rtl .woocommerce-subscriptions-related-orders-pagination-links {
+	flex-direction: row-reverse;
+}
+
+.rtl .woocommerce-subscriptions-related-orders-pagination-links .pagination-links {
+	direction: rtl;
+}
--- a/assets/css/view-subscription.css
+++ b/assets/css/view-subscription.css
@@ -91,7 +91,7 @@
 	text-align: center;
 }
 
-.woocommerce-subscriptions-related-orders-pagination-links {
+.woocommerce-subscriptions-related-orders-pagination-links.woocommerce-pagination {
 	display: flex;
 	flex-direction: row;
 	justify-content: space-between;
--- a/assets/css/view-subscription.css
+++ b/assets/css/view-subscription.css
@@ -96,10 +96,19 @@
 	flex-direction: row;
 	justify-content: space-between;
 	margin-bottom: 1em;
+	width: 100%;
 }
 
 .woocommerce-subscriptions-related-orders-pagination-links .pagination-links a {
 	display: inline-block;
+}
+
+.woocommerce-subscriptions-related-orders-pagination-links .pagination-links a.disabled {
+	visibility: hidden;
+}
+
+.woocommerce-subscriptions-related-orders-pagination-links .pagination-links a .symbol {
+	display: none;
 }
 
 .rtl .woocommerce-subscriptions-related-orders-pagination-links {
@@ -108,4 +117,18 @@
 
 .rtl .woocommerce-subscriptions-related-orders-pagination-links .pagination-links {
 	direction: rtl;
+}
+
+@media (max-width: 30em) {
+	.woocommerce-subscriptions-related-orders-pagination-links .pagination-links a {
+		padding: 0.5em 1em;
+	}
+
+	.woocommerce-subscriptions-related-orders-pagination-links .pagination-links a .label {
+		display: none;
+	}
+
+	.woocommerce-subscriptions-related-orders-pagination-links .pagination-links a .symbol {
+		display: inherit;
+	}
 }
--- a/includes/class-wc-subscription.php
+++ b/includes/class-wc-subscription.php
@@ -2151,13 +2151,13 @@ class WC_Subscription extends WC_Order {
 	 * @param int          $page          Optional. Can be used to specify which page of results is desired.
 	 * @param int          $limit         Optional. Can be used to specify how many results are desired per page. Defaults to -1, which is treated as meaning 'unlimited'.
 	 *
-	 * @return array {
+	 * @return object {
 	 *     array: orders,
 	 *     int:   total,
 	 *     int:   max_num_pages
 	 * }
 	 */
-	public function get_paginated_related_orders( string $return_fields = 'ids', $order_types = array( 'parent', 'renewal', 'switch' ), int $page = 1, int $limit = 10 ): array {
+	public function get_paginated_related_orders( string $return_fields = 'ids', $order_types = array( 'parent', 'renewal', 'switch' ), int $page = 1, int $limit = 10 ): object {
 		$order_types    = ! is_array( $order_types ) ? (array) $order_types : $order_types;
 		$related_orders = [];
 
@@ -2178,7 +2178,7 @@ class WC_Subscription extends WC_Order {
 
 		$order_ids     = array_unique( $this->get_related_order_ids( $order_types, 'flat' ) );
 		$total         = count( $order_ids );
-		$max_num_pages = (int) ceil( $total / $limit );
+		$max_num_pages = 0 === $limit ? 0 : (int) ceil( $total / $limit );
 		rsort( $order_ids );
 
 		if ( $limit >= 0 && $page > 0 ) {
@@ -2190,7 +2190,7 @@ class WC_Subscription extends WC_Order {
 			$related_orders[ $id ] = 'all' === $return_fields ? wc_get_order( $id ) : $id;
 		}
 
-		return array(
+		return (object) array(
 			'orders'        => $related_orders,
 			'total'         => $total,
 			'max_num_pages' => $max_num_pages,
--- a/includes/class-wc-subscription.php
+++ b/includes/class-wc-subscription.php
@@ -2144,7 +2144,7 @@ class WC_Subscription extends WC_Order {
 	/**
 	 * Offers a means of fetching paginated sets of related orders.
 	 *
-	 * @since 7.4.0
+	 * @since 7.5.0
 	 *
 	 * @param string       $return_fields The columns to return, either 'all' or 'ids'
 	 * @param array|string $order_types   Can include 'any', 'parent', 'renewal', 'resubscribe' and/or 'switch'. Custom types possible via the 'woocommerce_subscription_related_orders' filter. Defaults to array( 'parent', 'renewal', 'switch' ).
@@ -2162,17 +2162,17 @@ class WC_Subscription extends WC_Order {
 		$related_orders = [];
 
 		if ( ! in_array( $return_fields, array( 'all', 'ids' ), true ) ) {
-			wc_doing_it_wrong( __METHOD__, 'The $return_fields parameter must be either "all" or "ids".', '7.4.0' );
+			wc_doing_it_wrong( __METHOD__, 'The $return_fields parameter must be either "all" or "ids".', '7.5.0' );
 			$return_fields = 'ids';
 		}
 
 		if ( $page < 1 ) {
-			wc_doing_it_wrong( __METHOD__, 'The $page parameter must be a positive, non-zero integer.', '7.4.0' );
+			wc_doing_it_wrong( __METHOD__, 'The $page parameter must be a positive, non-zero integer.', '7.5.0' );
 			$page = 1;
 		}
 
 		if ( $limit < -1 ) {
-			wc_doing_it_wrong( __METHOD__, 'The $limit parameter must be an integer, and must have a value of -1 or higher.', '7.4.0' );
+			wc_doing_it_wrong( __METHOD__, 'The $limit parameter must be an integer, and must have a value of -1 or higher.', '7.5.0' );
 			$limit = 10;
 		}
 
--- a/includes/class-wc-subscription.php
+++ b/includes/class-wc-subscription.php
@@ -2142,6 +2142,62 @@ class WC_Subscription extends WC_Order {
 	}
 
 	/**
+	 * Offers a means of fetching paginated sets of related orders.
+	 *
+	 * @since 7.4.0
+	 *
+	 * @param string       $return_fields The columns to return, either 'all' or 'ids'
+	 * @param array|string $order_types   Can include 'any', 'parent', 'renewal', 'resubscribe' and/or 'switch'. Custom types possible via the 'woocommerce_subscription_related_orders' filter. Defaults to array( 'parent', 'renewal', 'switch' ).
+	 * @param int          $page          Optional. Can be used to specify which page of results is desired.
+	 * @param int          $limit         Optional. Can be used to specify how many results are desired per page. Defaults to -1, which is treated as meaning 'unlimited'.
+	 *
+	 * @return array {
+	 *     array: orders,
+	 *     int:   total,
+	 *     int:   max_num_pages
+	 * }
+	 */
+	public function get_paginated_related_orders( string $return_fields = 'ids', $order_types = array( 'parent', 'renewal', 'switch' ), int $page = 1, int $limit = 10 ): array {
+		$order_types    = ! is_array( $order_types ) ? (array) $order_types : $order_types;
+		$related_orders = [];
+
+		if ( ! in_array( $return_fields, array( 'all', 'ids' ), true ) ) {
+			wc_doing_it_wrong( __METHOD__, 'The $return_fields parameter must be either "all" or "ids".', '7.4.0' );
+			$return_fields = 'ids';
+		}
+
+		if ( $page < 1 ) {
+			wc_doing_it_wrong( __METHOD__, 'The $page parameter must be a positive, non-zero integer.', '7.4.0' );
+			$page = 1;
+		}
+
+		if ( $limit < -1 ) {
+			wc_doing_it_wrong( __METHOD__, 'The $limit parameter must be an integer, and must have a value of -1 or higher.', '7.4.0' );
+			$limit = 10;
+		}
+
+		$order_ids     = array_unique( $this->get_related_order_ids( $order_types, 'flat' ) );
+		$total         = count( $order_ids );
+		$max_num_pages = (int) ceil( $total / $limit );
+		rsort( $order_ids );
+
+		if ( $limit >= 0 && $page > 0 ) {
+			$offset    = ( $page - 1 ) * $limit;
+			$order_ids = array_slice( $order_ids, $offset, $limit );
+		}
+
+		foreach ( $order_ids as $id ) {
+			$related_orders[ $id ] = 'all' === $return_fields ? wc_get_order( $id ) : $id;
+		}
+
+		return array(
+			'orders'        => $related_orders,
+			'total'         => $total,
+			'max_num_pages' => $max_num_pages,
+		);
+	}
+
+	/**
 	 * Get the related order IDs for a subscription based on an order type.
 	 *
 	 * @since 1.0.0 - Migrated from WooCommerce Subscriptions v2.3.0
--- a/includes/class-wc-subscriptions-order.php
+++ b/includes/class-wc-subscriptions-order.php
@@ -968,8 +968,8 @@ class WC_Subscriptions_Order {
 	 * Introduced to support pagination of the related orders list (within the My Account > View Subscription
 	 * screen).
 	 *
-     * @since 7.5.0 Updated to support pagination of the related orders list.
-     *
+	 * @since 7.5.0 Updated to support pagination of the related orders list.
+	 *
 	 * @param WC_Subscription $subscription        The subscription whose related orders we are interested in.
 	 * @param int[]|null      $subscription_orders IDs of the related orders.
 	 * @param int|null        $page                The current page number.
--- a/includes/class-wc-subscriptions-order.php
+++ b/includes/class-wc-subscriptions-order.php
@@ -933,7 +933,7 @@ class WC_Subscriptions_Order {
 	 * Loads the related orders table on the view subscription page
 	 *
 	 * @since 1.0.0 Migrated from WooCommerce Subscriptions v2.0.
-	 * @since 7.4.0 Updated to support pagination of the related orders list.
+	 * @since 7.5.0 Updated to support pagination of the related orders list.
 	 *
 	 * @param WC_Subscription $subscription The subscription whose related orders we are interested in.
 	 */
--- a/includes/class-wc-subscriptions-order.php
+++ b/includes/class-wc-subscriptions-order.php
@@ -61,7 +61,8 @@ class WC_Subscriptions_Order {
 
 		add_action( 'woocommerce_order_details_after_order_table', __CLASS__ . '::add_subscriptions_to_view_order_templates', 10, 1 );
 
-		add_action( 'woocommerce_subscription_details_after_subscription_table', __CLASS__ . '::get_related_orders_template', 10, 1 );
+		add_action( 'woocommerce_subscription_details_after_subscription_table', __CLASS__ . '::get_related_orders_template' );
+		add_action( 'woocommerce_subscription_details_after_subscription_related_orders_table', __CLASS__ . '::get_related_orders_pagination_template', 5, 4 );
 
 		add_filter( 'woocommerce_my_account_my_orders_actions', __CLASS__ . '::maybe_remove_pay_action', 10, 2 );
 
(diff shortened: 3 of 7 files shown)


barryhughes:
✍🏼 I referenced `7.4.0` a few times (for `@since` tags, etc). I am happy to update those as needed, as the timing of this PR means it may not land in that release.

barryhughes:
Lint fails flagged in CI aren't actually related to lines touched in this PR. Happy to address them if preferred (my usual default is *not* to do that, but, I'm flexible!).
```
FILE: /home/runner/work/woocommerce-subscriptions-core/woocommerce-subscriptions-core/includes/class-wc-subscriptions-order.php
+------+--------------------------------------------------+
| Line | Description                                      |
+------+--------------------------------------------------+
| 2432 | Processing form data without nonce verification. |
| 2432 | Processing form data without nonce verification. |
+------+--------------------------------------------------+
```

barryhughes:
✍🏼 One other note: the pagination controls seem (to my easily pleased eye) pretty good when using the usual Twenty Twenty-something themes. However, they don't look *quite* as pleasing when using Storefront ... but since Storefront is also less of an active concern these days, I'm not sure if we ought to add special CSS rules for it or not (I assume not, but happy to revise as needed).

includes/class-wc-subscriptions-order.php
```diff
@@ -916,18 +916,32 @@ public static function add_subscriptions_to_view_order_templates( $order_id ) {
...
+	 * @param WC_Subscription $subscription The subscription whose related orders we are interested in.
 	 */
 	public static function get_related_orders_template( $subscription ) {
-
-		$subscription_orders = $subscription->get_related_orders();
+		// phpcs:disable WordPress.Security.NonceVerification.Recommended
+		// phpcs:disable WordPress.Security.ValidatedSanitizedInput.MissingUnslash
+		// phpcs:disable WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
+		$page  = (int) ( $_GET['related_orders_page'] ?? 1 );
+		$limit = (int) ( $_GET['related_orders_limit'] ?? 10 );
```
barryhughes: I left this in place (`$limit = ...`), but happy to remove it. It provides some flexibility for themers/developers, and the attendant risk of a user supplying `-1` *probably* isn't that bad when you consider that is effectively the same as the extant behaviour in our stable release.

An alternative could be introducing a filter hook but, rather than inflate the public API surface pre-emptively, it may be better to wait and see if there is demand for such a thing.

templates/myaccount/related-orders.php
```diff
@@ -99,4 +103,31 @@
...
 
+<div class="woocommerce-subscriptions-related-orders-pagination-links woocommerce-pagination">
+	<?php
+	$prev_page = max( 1, $page - 1 );
+	$next_page = min( $max_num_pages, $page + 1 );
+
+	$first = '?related_orders_page=1#woocommerce-subscriptions-related-orders-table';
+	$prev  = '?related_orders_page=' . $prev_page . '#woocommerce-subscriptions-related-orders-table';
+	$next  = '?related_orders_page=' . $next_page . '#woocommerce-subscriptions-related-orders-table';
+	$last  = '?related_orders_page=' . $max_num_pages . '#woocommerce-subscriptions-related-orders-table';
```
vbelolapotkov: I think there is an opportunity to extract helper method for composing query params

includes/class-wc-subscription.php
```diff
@@ -2121,6 +2121,62 @@ public function get_related_orders( $return_fields = 'ids', $order_types = array
...
+	public function get_paginated_related_orders( string $return_fields = 'ids', $order_types = array( 'parent', 'renewal', 'switch' ), int $page = 1, int $limit = 10 ): object {
+		$order_types    = ! is_array( $order_types ) ? (array) $order_types : $order_types;
+		$related_orders = [];
+
+		if ( ! in_array( $return_fields, array( 'all', 'ids' ), true ) ) {
+			wc_doing_it_wrong( __METHOD__, 'The $return_fields parameter must be either "all" or "ids".', '7.5.0' );
+			$return_fields = 'ids';
+		}
+
+		if ( $page < 1 ) {
```
vbelolapotkov, line 2169: Probably a minor thing, but looks like floats will pass that check. E.g. 1.6 < 1 would return false and proceed. IDK if it'll cause any issues further down the line, but maybe worth adding is_int check?

vbelolapotkov (approved):
I've tested the code and everything works as expected, great enhancement @barryhughes! 🙌 

I'm surprised the number of DB queries is still high, but I suppose it's because we still need to fetch meta keys for order to select only related ones.

Anyway, DB query time is reduced ~30% on my local site, so I think it's a good enhancement.

I've made some minor comments, but those are not critical IMO, so approving.

Finally, you've raised some good points in description, but I don't have enough knowledge at the moment to comment on them, so worth asking for additional opinion.

vbelolapotkov:
> Lint fails flagged in CI aren't actually related to lines touched in this PR. Happy to address them if preferred (my usual default is _not_ to do that, but, I'm flexible!).

I agree here.

includes/class-wc-subscription.php
barryhughes: We don't do anything to `$page` before this point, so I feel we can be confident it is indeed an integer and not some other type (if for instance a float was supplied by the caller, it would still be coerced into an int before the method body runs).

barryhughes: Or am I misunderstanding—do you mean we should remove the type hint on the parameter, and manually check the type here?

vbelolapotkov: I'd check the type explicitly for safety. If it's not, somebody is probably tampering the url/request, and that's ok IMO to treat value as bad argument.

barryhughes: Well, someone can tamper with URL (or even the `$_GET` superglobal) all they want, but when we reach this point the type of `$page` can only be an integer (PHP guarantees it), *unless* we remove the [type declaration](https://www.php.net/manual/en/language.types.declarations.php) from the function argument. 

I don't feel that we should remove that. And, unless we remove it, there's no benefit to a manual type check here, right, because `$page` can't be any other type.

barryhughes: Removing the type declaration also increases the chance of subtle errors going undetected in development.

(...though, I still have a feeling I am somewhat misunderstanding you here 😅)

templates/myaccount/related-orders.php
barryhughes: WordPress provides just such a helper, in the form of `add_query_arg()`, though I'm not sure it would make the code any more concise in this case. That said, I agree what we currently have doesn't feel *great* (though, it's simple).

Are you thinking along these sorts of lines?

```php
$first = paginated_related_orders_url( 1 );
$prev  = paginated_related_orders_url( $prev_page );
$next  = paginated_related_orders_url( $next_page );
/* etc */
```

If so, though a little unconventional, what do you think of defining this as an anonymous function right here, within the template (ie `$paginated_url = function () {}`)? This would avoid adding such a specific helper function to the public API surface.

Further alternative: we could move this entire chunk of pagination code into a child template.

includes/class-wc-subscription.php
vbelolapotkov: > unless we remove the [type declaration](https://www.php.net/manual/en/language.types.declarations.php) from the function argument.

I missed the declaration, sorry 🤦 TIL that PHP does coercion in that case if it can, unless strict mode is enabled. So it's right, we can keep the check as is and rely on argument type declaration.

templates/myaccount/related-orders.php
vbelolapotkov: I like anonymous function option, but do not insist on it :)

barryhughes:
Interesting observation/thing that tripped me up: if WooCommerce Subscriptions Gifting is also active, then it overrides the `myaccount/related-orders.php` template with its own (so, pagination will not be present).

templates/myaccount/related-orders.php
barryhughes: Added in https://github.com/Automattic/woocommerce-subscriptions-core/pull/833/commits/91959bdc5d969d280e1b9de9083051957c6f7939.

barryhughes:
@vbelolapotkov could I ask you to take one further look, particularly at these new commits?

https://github.com/Automattic/woocommerce-subscriptions-core/pull/833/commits/91959bdc5d969d280e1b9de9083051957c6f7939
https://github.com/Automattic/woocommerce-subscriptions-core/pull/833/commits/2c4d394c44f53ae4ca7e5469901cfa52a6cee5cb

templates/myaccount/related-orders.php
```diff
@@ -99,4 +103,31 @@
...
+		<?php echo $page_link( 1, 'first', _x( 'First', 'Related orders pagination', 'woocommerce-subscriptions' ), 1 !== $page ); ?>
+		<?php echo $page_link( $prev_page, 'prev', _x( 'Previous', 'Related orders pagination', 'woocommerce-subscriptions' ), 1 !== $page ); ?>
+	</div>
+
+	<div class="pagination-links">
+		<?php echo $page_link( $next_page, 'next', _x( 'Next', 'Related orders pagination', 'woocommerce-subscriptions' ), $page < $max_num_pages ); ?>
+		<?php echo $page_link( $max_num_pages, 'last', _x( 'Last', 'Related orders pagination', 'woocommerce-subscriptions' ), $page < $max_num_pages ); ?>
+	</div>
+	<?php // phpcs:enable ?>
+</div>
```
vbelolapotkov: I like this implementation a lot, well done! I've made another round of testing, and noticed that layout looks terrible on mobile devices: 

CleanShot 2025-04-28 at 11 24 30@2x

Let's update markup and styling for that:
- Replace text on buttons with "<<" for first and "<" for pervious.
- Fix buttons alignment on the right edge.

includes/class-wc-subscriptions-order.php
```diff
@@ -916,18 +916,31 @@ public static function add_subscriptions_to_view_order_templates( $order_id ) {
...
+		// phpcs:disable WordPress.Security.NonceVerification.Recommended
+		// phpcs:disable WordPress.Security.ValidatedSanitizedInput.MissingUnslash
+		// phpcs:disable WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
+		$page = (int) ( $_GET['related_orders_page'] ?? 1 );
+		// phpcs:enable WordPress.Security.NonceVerification.Recommended
+		// phpcs:enable WordPress.Security.ValidatedSanitizedInput.MissingUnslash
+		// phpcs:enable WordPress.Security.ValidatedSanitizedInput.InputNotSanitized
 
-		$subscription_orders = $subscription->get_related_orders();
+		$related_orders      = $subscription->get_paginated_related_orders( 'ids', array( 'parent', 'renewal', 'switch' ), $page, 10 );
```
vbelolapotkov, line 949: @barryhughes, I noticed you switched back to hardcoded limit of 10 items per page. Could you please clarify your reasoning here?

vbelolapotkov (changes requested):
@barryhughes I've tested everything again, and it works as expected in case of many orders.

When tested with little number of related orders I noticed that pagination buttons are still rendered and they are clickable, which is not ideal IMO. The most annoying behavior is that on the first click the page is reloaded.

CleanShot 2025-04-28 at 11 50 13@2x

What do you think about hiding them completely (that would be my preference) when there is only one page or at least make them non-clickable?

> Interesting observation/thing that tripped me up: if WooCommerce Subscriptions Gifting is also active, then it overrides the myaccount/related-orders.php template with its own (so, pagination will not be present).

Tested this as well, and you're right. I guess to fix this, we'll have to ship a template update for subs gifting. Do you plan doing this? Also are you aware of other our extensions which customize this template? I suppose we'd need to communicate this change with devs during or before release.

Side note, subs gifting list looks cleaner IMO because of the links instead of buttons. Maybe we should do the same? Let's ask designers, cc @poligilad-auto on this or any other design recommendations here.

CleanShot 2025-04-28 at 11 47 00@2x

includes/class-wc-subscriptions-order.php
barryhughes: Sure! I outlined some of this in the full commit message for https://github.com/Automattic/woocommerce-subscriptions-core/pull/833/commits/2c4d394c44f53ae4ca7e5469901cfa52a6cee5cb, but to expand on that a little:

- Previously, *all* related orders were loaded. So, in some respects, there is little harm in allowing this (the worst thing someone might do is take advantage of this to bring back the earlier behavior, which worked, mostly successfully, for a long time).
- However, after pausing to think about it, I also don't think we should let end-users have this level of control. Suppose in some cases the previous behavior does indeed have an adverse affect on performance, well, we shouldn't let customers bring that back since it could impact the rest of the site. 
- The obvious compromise is to introduce a filter hook. This would allow merchants and extension developers to safely change the limit if they need to. I didn't want to add this initially, though, because every time we add a hook (or indeed add anything to the public API surface) we are making a promise that potentially makes it harder to refactor in the future. So, my thinking was that we'd wait and see if there is even demand for such a hook before adding it.

vbelolapotkov: Good points, let's keep it hardcoded for now, and change later if asked. I'm also in support of having less hooks. Even later, I'd move it to options and/or settings for configuration instead of introducing a new hook.

barryhughes:
🎅🏼 I come bearing updates.

---

> Replace text on buttons with "<<" for first and "<" for previous.

<div align="center"></div>

This has been implemented for small viewports.

> Fix buttons alignment on the right edge.

Also addressed, and a small tweak made to improve things when using Storefront.

> What do you think about hiding them completely (that would be my preference) when there is only one page or at least make them non-clickable?

Done, and I also hide them if they are not needed (ie, the next and last buttons are not needed if we are on the final page of the result set).

> I guess to fix this, we'll have to ship a template update for subs gifting. Do you plan doing this?

I will create an issue so we don't forget (or perhaps will go ahead and create a PR since it's a small change). Inspired by this problem, though, I moved the pagination controls into their own template (and I integrate them into the *related-orders* template using an existing hook).

For Gifting, we'll still need to make an update, though, so that it triggers the same hook and passes along the same set of parameters.

assets/css/view-subscription.css
```diff
@@ -90,3 +90,45 @@
...
+
+.rtl .woocommerce-subscriptions-related-orders-pagination-links {
+	flex-direction: row-reverse;
+}
+
+.rtl .woocommerce-subscriptions-related-orders-pagination-links .pagination-links {
+	direction: rtl;
+}
+
+@media (max-width: 30em) {
```
vbelolapotkov: @barryhughes I'd like to understand where the value comes from. The rest of the codebase uses px values (different and weird at times), so I'm wondering what's your source for this choice?

includes/class-wc-subscriptions-order.php
```diff
@@ -932,25 +933,66 @@ public static function add_subscriptions_to_view_order_templates( $order_id ) {
...
 			);
 		}
 	}
 
+	/**
+	 * Introduced to support pagination of the related orders list (within the My Account > View Subscription
+	 * screen).
+	 *
+     * @since 7.5.0 Updated to support pagination of the related orders list.
+     *
```
vbelolapotkov: FYI, there seem to be some formatting discrepancy. Looks like spaces applied instead of tabs.

includes/class-wc-subscriptions-order.php
```diff
@@ -932,25 +933,66 @@ public static function add_subscriptions_to_view_order_templates( $order_id ) {
...
+	 * @param WC_Subscription $subscription        The subscription whose related orders we are interested in.
+	 * @param int[]|null      $subscription_orders IDs of the related orders.
+	 * @param int|null        $page                The current page number.
+	 * @param int|null        $max_num_pages       The maximum number of pages in the set.
+	 *
+	 * @return void
+	 */
+	public static function get_related_orders_pagination_template( WC_Subscription $subscription, ?array $subscription_orders = null, ?int $page = null, ?int $max_num_pages = null ) {
```
vbelolapotkov, line 980: Double-checking my understanding here. I see that $subscription and $subscription_orders are not used in pagination template. Is it here for other extensions in case they want to have more advanced pagination? If that's the case, sure let's keep it, otherwise, we can limit pagination template to only two params used page and max_num_pages.

vbelolapotkov (approved):
Great improvements @barryhughes, thank you! I've tested it and everything works well. I have some minor non-blocking comments, but don't want to hold this PR anymore, so approving. Feel free to merge once ready.

Styling and layout look much better now. The small thing would be nice to fix before shipping is the space between buttons and customer billing info. Right now there is very little space, especially in Storefront.

CleanShot 2025-05-01 at 14 00 21@2x

> What do you think about hiding them completely (that would be my preference) when there is only one page or at least make them non-clickable?
> Done, and I also hide them if they are not needed (ie, the next and last buttons are not needed if we are on the final page of the result set).

I played with it, and found UX a bit odd, especially when you have 11 orders. Those buttons just jump on the left/right. I leave the final decision up to you, but I'd show/hide all of them together.

>  Inspired by this problem, though, I moved the pagination controls into their own template (and I integrate them into the related-orders template using an existing hook).

I love that, thanks!

assets/css/view-subscription.css
barryhughes: It looked like both units of measure are in use in this file (for instance, look at the two blocks which precede the pagination rules—both use `em`) though you're right that `px` is more common. It was really just personal choice: I feel `em` makes more sense when we're shaping how text is laid out. I can certainly switch to `px`. As to how I got to this specific value, just through experimentation with different themes since there isn't a single breakpoint that works perfectly for all.

includes/class-wc-subscriptions-order.php
barryhughes: Apologies ... not sure how that slipped through. Should be addressed in https://github.com/Automattic/woocommerce-subscriptions-core/pull/833/commits/d7d92e9ff100566e11735d4fae620514c3846ba2.

includes/class-wc-subscriptions-order.php
barryhughes: The template doc block was incorrect, and implied that `$subscription` and `$subscription_orders` were in scope. That was wrong, so I've updated it:

https://github.com/Automattic/woocommerce-subscriptions-core/pull/833/commits/266e9b34d1d623a0fcff10029a6b3f929f953f99

Does that clarify things?

barryhughes:
> I played with it, and found UX a bit odd, especially when you have 11 orders. Those buttons just jump on the left/right. I leave the final decision up to you, but I'd show/hide all of them together.

Yep, I agree. Updated so that they remain on the page, but they are more obviously disabled (ie, clicking on them doesn't do anything): https://github.com/Automattic/woocommerce-subscriptions-core/pull/833/commits/1ed30d2f08901f604143d504e71188c930a40e96.

barryhughes:
> The small thing would be nice to fix before shipping is the space between buttons and customer billing info. Right now there is very little space, especially in Storefront.

In Storefront specifically, a rule from the theme was taking precedence so our margins weren't being respected. I've made our own rule a little more specific (see https://github.com/Automattic/woocommerce-subscriptions-core/pull/833/commits/ea8e383534d9a00c0a2b2caeffaeb84f93a81594). This now looks fairly balanced to me, on Storefront as well as Twenty Twenty-Two/Three/Four, but let me know if you want them to be tweaked a little more.

barryhughes:
> I have some minor non-blocking comments, but don't want to hold this PR anymore, so approving. Feel free to merge once ready.

Thanks, though definitely feel free to take another look if you have time and bandwidth. 

At this point, I don't want to merge immediately (there are some structural changes unrelated to this issue that it probably makes sense to complete first), so we have time and space to keep finessing, bandwidth-permitting. Thank you, too, for all the feedback—it should result in a much more polished change 😎

assets/css/view-subscription.css
vbelolapotkov: Thanks for clarifying. I'd say go with whatever makes the most sense for you at this point, but I'd want to review our design guidelines later, and hopefully come to some more consistent approach.

includes/class-wc-subscriptions-order.php
vbelolapotkov: No worries, it's hard/impossible to spot. I noticed in GitHub UI, but it looked fine in IDE until I highlighted everything :D

poligilad-auto:
Hey @vbelolapotkov thanks for the ping!

> Side note, subs gifting list looks cleaner IMO because of the links instead of buttons. Maybe we should do the same? Let's ask designers, cc @poligilad-auto on this or any other design recommendations here.

Regarding the buttons, I think it makes sense to use buttons as "View" order / subscription are the main actions users want to take on the page. Also, it's best to keep the account area unified and that is what we are using in the "Orders" page.

I took a look and I see we are using the pagination style that is also used in the `Orders` page.

image

Ideally we'd used the pagination component and style that's coming from the site 👇

image

I'll check how we can raise this topic holistically for the account area.

cc @barryhughes

barryhughes:
> I took a look and I see we are using the pagination style that is also used in the Orders page.

Thanks for taking a peek, @poligilad-auto! Yes, that was the basis for this.

> Ideally we'd used the pagination component and style that's coming from the site 👇

OK. I'm not sure what's involved there or what sort of variation we might see in pagination implementations across themes, but can certainly take a look.

Would this be worth holding up the change for? I'm happy with that, to be clear, but just trying to gauge if it's something where we can ship now (there are performance and usability benefits to this work, for stores with a large number of renewal or switch orders) and iterate later on the pagination aesthetics?

poligilad-auto:
> Would this be worth holding up the change for? I'm happy with that, to be clear, but just trying to gauge if it's something where we can ship now (there are performance and usability benefits to this work, for stores with a large number of renewal or switch orders) and iterate later on the pagination aesthetics?

@barryhughes I think it’s definitely good to ship for now! We can explore that as a separate step. I’ll post this in a quick P2, I’m also not sure how it was originally implemented.

barryhughes:

PR #837, opened 2025-04-15
Fixed manual discounts on pay for order flow

leonardola:
Fixes https://github.com/woocommerce/woocommerce-subscriptions/issues/4233

## Description
This PR fixes a bug that when manually editing cost of one sub product ended up applying the discount to all sub products on the order

## How to test this PR
**Setup**
- Checkout to this branch
- Create 2 subscription products with the same cadence
- Create a product discount coupon code (non-recurring) that only applies to one of the products you created
- Enable Check payment method
- Enable /wp-admin > WooCommerce > Settings > Subscriptions > Accept Manual Renewals

**Single use coupon**
- As a customer create an order with two subscription products with the same cadence
- Pay with Check
- As an admin edit the parent order created
- Add the discount coupon
- Give a manual discount to the other product
- Set the order status to `Pending Payment`
- Save the changes
- Copy the `Customer Payment Page` link
- As a customer open the link
- Make sure both the manual discount and the coupon discount showed up and are correct

**Recurring coupon**
- As a customer create an order with two subscription products with the same cadence
- Pay with Check
- As an admin edit the subscription
- Give a manual discount to one of the products
- Add the discount coupon
- Set the order status to `Pending Payment`
- Save the changes
- Delete the coupon
- Create a renewal order
- Copy the `Customer Payment Page` link from the renewal order
- As a customer open the link
- Make sure both the manual discount and the coupon discount showed up and are correct

## Product impact

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,5 +1,9 @@
 *** WooCommerce Subscriptions Core Changelog ***
 
+= 8.3.0 - 2025-xx-xx =
+* Fix - Manual discounts on pay for order flow
+
+
 = 8.2.0 - 2025-04-14 =
 * Update - Increase the number of args accepted by wcs_get_subscriptions(), to bring about parity with wc_get_orders().
 * Dev - Update wcs_maybe_prefix_key() and wcs_maybe_unprefix_key() to support an array of keys.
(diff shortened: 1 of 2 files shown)


alefesouza:
Tested on both Classic and Blocks checkout on the latest WooCommerce version (9.8.2) and it's not working, am I doing something wrong? Also I remember the pay for order page had a different design and now it's just a redirect, maybe it's related.

| | | |
|-|-|-|
| Screenshot 2025-04-28 at 6 37 54 PM | Screenshot 2025-04-28 at 6 38 57 PM | Screenshot 2025-04-28 at 6 30 45 PM |

alefesouza (approved):
I got a problem before but it looks like the manual discount is removed when we apply a coupon from the admin order page, then we need to add the manual discount again.

Now it works as expected.

✅ The manual discount and coupon discount are applied as expected on the pay for order flow.

PR #839, opened 2025-04-24
Generate docs for developers

vbelolapotkov:
Fixes #

## How to test this PR

1. Go to …
2. Click on …

## Product impact

### DIFF
--- a/docs/init.md
+++ b/docs/init.md
@@ -1,6 +1,6 @@
 # WooCommerce Subscriptions Core Initialization Sequence
 
-This document explains how WooCommerce Subscriptions Core is loaded and initialized under different contexts including admin pages, customer-facing pages, Action Scheduler events, REST API requests, and AJAX calls.
+This document explains how WooCommerce Subscriptions Core is loaded and initialized.
 
 ## Core Initialization Flow
 
@@ -15,7 +15,7 @@ Regardless of the context, WooCommerce Subscriptions Core follows this base init
 
 ### Key Components Initialized
 
-During the `init()` method, the following core components are initialized:
+During the `WC_Subscriptions_Core_Plugin::init()` method all core components are initialized:
 
 ```php
 // Key classes initialized immediately
@@ -46,6 +46,8 @@ add_action('after_setup_theme', array('WC_Subscriptions_Upgrader', 'init'), 11);
 add_action('init', array('WC_PayPal_Standard_Subscriptions', 'init'), 11);
 ```
 
+Some classess need to be initialized after all other plugins are loaded, this is handled by hooking `WC_Subscriptions_Core_Plugin::init_version_dependant_classes()` to `plugins_loaded` action.
+
 ### Hooks Registration
 
 The `init_hooks()` method sets up the following key hooks:
@@ -59,125 +61,16 @@ The `init_hooks()` method sets up the following key hooks:
 
 ## Context-Specific Initialization
 
-### Admin Pages
+Core plugin doesn't have custom logic for it and relies on WordPress and WooCommerce logic here.
+So some hooks might be skipped by WordPress or WooCommerce in various context (Admin, Store front end, REST_API or WP_Ajax).
 
-When WooCommerce Subscriptions Core is loaded in the WordPress admin:
+However, WooCommerce Subscriptions Core provides global helper functions which allow to execute code within certain context only:
 
-1. The core initialization flow runs first
-2. `WC_Subscriptions_Admin::init()` is called, which:
-   - Loads admin-specific scripts and styles
-   - Registers admin menus and settings
-   - Sets up meta boxes for subscription editing
-   - Configures admin notices
-
-3. Version-dependent admin classes are initialized via `init_version_dependant_classes()`:
-   ```php
-   new WCS_Admin_Post_Types();
-   new WCS_Admin_Meta_Boxes();
-   // ... other admin-specific components
-   ```
-
-4. Admin-specific hooks are registered for:
-   - Order management screens
-   - Product management screens
-   - Payment gateway configuration
-   - Reporting and analytics
-
-### Customer-Facing Pages
-
-For front-end pages visible to customers:
-
-1. The core initialization flow runs
-2. Front-end specific components are initialized:
-   - `WC_Subscriptions_Cart` - manages subscription products in the cart
-   - `WC_Subscriptions_Checkout` - handles checkout flow for subscriptions
-   - `WC_Subscriptions_Frontend_Scripts` - loads necessary JS/CSS
-
-3. No admin-specific components are loaded, which keeps front-end pages lean
-4. `WCS_Template_Loader` manages template overrides and customizations
-
-### Action Scheduler Events
-
-When an Action Scheduler event runs for subscriptions:
-
-1. WordPress core loads and the request is identified as a `cron` request (Action Scheduler uses WP Cron)
-2. The core initialization flow runs, but admin UI components are not loaded
-3. `WCS_Action_Scheduler` (initialized during core initialization) processes the scheduled event:
-   - Retrieves the correct subscription by ID from the action arguments
-   - Performs the scheduled action (payment processing, trial end, etc.)
-   - Updates subscription dates and statuses as appropriate
-
-4. The subscription status may change based on the event, which then triggers:
-   - Email notifications
-   - Payment processing
-   - Status changes and associated actions
-
-### REST API Requests
-
-For REST API requests related to subscriptions:
-
-1. WordPress identifies the request as a REST API request via `WC()->is_rest_api_request()`
-2. The core initialization flow runs
-3. WC Subscriptions checks if it's a REST API request using `wcs_is_rest_api_request()`
-4. Admin UI components are not loaded to keep the request lightweight
-5. Data handling remains consistent with standard requests, but presentation layers are skipped
-
-### AJAX Requests
-
-For AJAX requests related to subscriptions:
-
-1. WordPress identifies the request as an AJAX request via `wp_doing_ajax()`
-2. The core initialization flow runs
-3. WC Subscriptions checks if it's an AJAX request using `wcs_doing_ajax()`
-4. Depending on the specific AJAX action:
-   - Admin-specific components may be loaded for admin AJAX requests
-   - Front-end components for customer AJAX requests
-   - The response is typically JSON formatted data rather than rendered HTML
-
-## Key Differences Between Contexts
-
-| Context | Admin UI Components | Frontend Components | Template Loading | Performance Considerations |
-|---------|--------------------|--------------------|-----------------|---------------------------|
-| Admin Pages | ✅ All loaded | ❌ Not loaded | ✅ Admin templates only | Loads more components for full functionality |
-| Customer Pages | ❌ Not loaded | ✅ All loaded | ✅ Frontend templates | Focuses on cart, checkout, and my-account functionality |
-| Action Scheduler | ❌ Not loaded | ❌ Not loaded | ❌ Not needed | Lightweight, focuses on data processing only |
-| REST API | ❌ Not loaded | ⚠️ Partial loading | ❌ Not needed | Data-focused, returns structured data |
-| AJAX | ⚠️ Context-dependent | ⚠️ Context-dependent | ❌ Not needed | Minimal loading for fast responses |
-
-## Initialization Detection Functions
-
-WC Subscriptions provides helper functions to determine the current request context:
-
-```php
-// Check if current request is AJAX
-wcs_doing_ajax()
-
-// Check if current request is wp-cron (includes Action Scheduler)
-wcs_doing_cron()
-
-// Check if current request is REST API
-wcs_is_rest_api_request()
-
-// Check if current request is frontend (not admin, not AJAX, not cron, not REST)
-wcs_is_frontend_request()
-```
-
-## Performance Considerations
-
-- Admin pages load the most components and have the highest overhead
-- Action Scheduler events are optimized for performance with minimal component loading
-- REST API and AJAX requests are designed to be lightweight for responsiveness
-- Frontend pages load only what's needed for customer interactions
-
-## Common Issues and Troubleshooting
-
-1. **Action Scheduler Events Not Running**: Check if cron is working correctly on your server and that the action is properly scheduled in the Action Scheduler tables.
-
-2. **Admin vs Frontend Loading Conflicts**: If you're building a custom integration, be aware that some components aren't available in all contexts.
-
-3. **REST API Authentication**: REST API requests require proper authentication; errors often occur due to permission issues rather than initialization problems.
-
-4. **AJAX Nonce Verification**: AJAX requests include nonce verification that can fail if the user session expires.
+- `wcs_is_rest_api_request()`
+- `wcs_is_checkout_blocks_api_request()`
+- `wcs_doing_cron()`
+- `wcs_doing_ajax()`
+- `wcs_is_frontend_request()`
 
 ## Extending the Initialization Process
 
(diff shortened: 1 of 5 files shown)


docs/init.md
```diff
@@ -0,0 +1,201 @@
...
+   - `WC_Subscriptions_Frontend_Scripts` - loads necessary JS/CSS
+
+3. No admin-specific components are loaded, which keeps front-end pages lean
+4. `WCS_Template_Loader` manages template overrides and customizations
+
+### Action Scheduler Events
+
+When an Action Scheduler event runs for subscriptions:
+
+1. WordPress core loads and the request is identified as a `cron` request (Action Scheduler uses WP Cron)
```
barryhughes: This doesn't seem quite right. Action Scheduler *does* use WP Cron, but not exclusively. Sometimes (perhaps even *oftentimes),* scheduled actions are processed outside of this context.

docs/init.md
```diff
@@ -0,0 +1,201 @@
...
+2. The core initialization flow runs, but admin UI components are not loaded
```
barryhughes: Admin UI components not being loaded in this context sounds desirable, but it's a bit of a half-truth. For example, we still register callbacks to enqueue JS and CSS (even if those hooks typically will not be triggered in this context).

docs/init.md
```diff
@@ -0,0 +1,201 @@
...
+3. `WCS_Action_Scheduler` (initialized during core initialization) processes the scheduled event:
+   - Retrieves the correct subscription by ID from the action arguments
+   - Performs the scheduled action (payment processing, trial end, etc.)
+   - Updates subscription dates and statuses as appropriate
```
barryhughes: This reads like `WCS_Action_Scheduler` is a sort of single processing hub for scheduled actions, but scheduled actions are enqueued and processed across various parts of the codebase, not just here. Example: `WCS_Background_Repairer::update_item()`.

Also, some of the scheduled actions `WCS_Action_Scheduler` may create (such as for subscription date changes) are processed elsewhere.

docs/init.md
```diff
@@ -0,0 +1,201 @@
...
+4. The subscription status may change based on the event, which then triggers:
+   - Email notifications
+   - Payment processing
+   - Status changes and associated actions
+
+### REST API Requests
+
+For REST API requests related to subscriptions:
+
+1. WordPress identifies the request as a REST API request via `WC()->is_rest_api_request()`
```
barryhughes: This isn't quite right. *WordPress* definitely does not rely on WooCommerce to identify the request as being related to the REST API 🙃

docs/init.md
```diff
@@ -0,0 +1,201 @@
...
+   - Updates subscription dates and statuses as appropriate
+
+4. The subscription status may change based on the event, which then triggers:
+   - Email notifications
+   - Payment processing
+   - Status changes and associated actions
+
+### REST API Requests
+
+For REST API requests related to subscriptions:
```
barryhughes: This section probably makes more sense if we view these docs as relating to the WooCommerce Subscriptions product as a whole, rather than just for Subscriptions Core (I'm not actually sure what the intention is, though).

Anyway, most of our REST API logic sits outside of Core.

docs/init.md
```diff
@@ -0,0 +1,201 @@
...
+
+1. WordPress identifies the request as a REST API request via `WC()->is_rest_api_request()`
+2. The core initialization flow runs
+3. WC Subscriptions checks if it's a REST API request using `wcs_is_rest_api_request()`
+4. Admin UI components are not loaded to keep the request lightweight
```
barryhughes: Similar to an earlier comment re handling of scheduled events, this sounds like a desirable separation of concerns but I'm not sure it fully reflects the reality (of course, we generally won't render anything destined for the admin UI in this context but that doesn't mean we aren't hooking things up ready to do so).

docs/init.md
```diff
@@ -0,0 +1,201 @@
...
+5. Data handling remains consistent with standard requests, but presentation layers are skipped
+
+### AJAX Requests
+
+For AJAX requests related to subscriptions:
+
+1. WordPress identifies the request as an AJAX request via `wp_doing_ajax()`
```
barryhughes: This doesn't sound right to me. `wp_doing_ajax()` can be used to check if WordPress has identified the request as being an ajax one, but is not itself used to make the determination.

docs/init.md
```diff
@@ -0,0 +1,201 @@
...
+2. The core initialization flow runs
+3. WC Subscriptions checks if it's an AJAX request using `wcs_doing_ajax()`
```
barryhughes: I think this is essentially back-to-front.

In the context of this list, it makes it seem like this function is called before Subscriptions Core initializes any ajax-related logic, but that doesn't really match what happens. Instead, callbacks are registered (using the standard `wp_ajax_*` prefix) and WordPress will invoke them or not.

This function is, afaik, used primarily to _avoid_ setting things up within an ajax context.

docs/init.md
```diff
@@ -0,0 +1,201 @@
...
+4. Depending on the specific AJAX action:
+   - Admin-specific components may be loaded for admin AJAX requests
+   - Front-end components for customer AJAX requests
```
barryhughes: I'm not sure what this means, I guess, but it sounds misleading. Generally an ajax callback will not initialize a new component, it will just perform an action and indicate success, or else return a packet of data.

docs/init.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Key Differences Between Contexts
+
+| Context | Admin UI Components | Frontend Components | Template Loading | Performance Considerations |
+|---------|--------------------|--------------------|-----------------|---------------------------|
+| Admin Pages | ✅ All loaded | ❌ Not loaded | ✅ Admin templates only | Loads more components for full functionality |
+| Customer Pages | ❌ Not loaded | ✅ All loaded | ✅ Frontend templates | Focuses on cart, checkout, and my-account functionality |
+| Action Scheduler | ❌ Not loaded | ❌ Not loaded | ❌ Not needed | Lightweight, focuses on data processing only |
+| REST API | ❌ Not loaded | ⚠️ Partial loading | ❌ Not needed | Data-focused, returns structured data |
+| AJAX | ⚠️ Context-dependent | ⚠️ Context-dependent | ❌ Not needed | Minimal loading for fast responses |
```
barryhughes: Many of the concerns I already listed are relevant here, too. I think this paints a misleading picture.

docs/init.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Initialization Detection Functions
+
+WC Subscriptions provides helper functions to determine the current request context:
+
+```php
+// Check if current request is AJAX
+wcs_doing_ajax()
+
+// Check if current request is wp-cron (includes Action Scheduler)
```
barryhughes: Similar to an earlier comment, this is misleading. All `wcs_doing_cron()` tells us, is that we are (or are not) inside the context of a cron request. I'd drop the Action Scheduler element.

docs/init.md
```diff
@@ -0,0 +1,201 @@
...
+// Check if current request is frontend (not admin, not AJAX, not cron, not REST)
+wcs_is_frontend_request()
+```
+
+## Performance Considerations
+
+- Admin pages load the most components and have the highest overhead
+- Action Scheduler events are optimized for performance with minimal component loading
+- REST API and AJAX requests are designed to be lightweight for responsiveness
+- Frontend pages load only what's needed for customer interactions
```
barryhughes: This reads like a (slightly aspirational) statement of how things are, rather than what practices to follow in each context, or when to move work from one context to another.

docs/init.md
```diff
@@ -0,0 +1,201 @@
...
+
+## Common Issues and Troubleshooting
+
+1. **Action Scheduler Events Not Running**: Check if cron is working correctly on your server and that the action is properly scheduled in the Action Scheduler tables.
```
barryhughes: This is a pretty common question and so it is worth answering, but it also surfaces across many products that use Action Scheduler. Perhaps it would be worth creating and then linking to a more complete troubleshooting resource (that could be shared by other plugins) 🤔

docs/init.md
```diff
@@ -0,0 +1,201 @@
...
+
+1. Hook into the appropriate WordPress or WooCommerce action hooks
+2. Use the plugin-specific filters provided by WC Subscriptions
+3. Implement your initialization logic in the correct context (admin, frontend, etc.)
+
+Example:
+
+```php
+// Add custom admin initialization
+add_action('woocommerce_subscriptions_initialized', function() {
```
barryhughes: Is this a real hook?

docs/overview.md
```diff
@@ -0,0 +1,140 @@
...
+
+3. **Functional Organization**: Most functionality is organized into classes by feature with supporting functions in dedicated files.
+
+## Key Components
+
+### 1. Subscription Data Model
+
+- `WC_Subscription` - Core class that represents a subscription
+- `WC_Subscriptions_Order` - Manages the relationship between orders and subscriptions
+- Data is stored in the `wp_posts` table as a custom `shop_subscription` post type with meta data in `wp_postmeta`
```
barryhughes: Or is stored within the dedicated orders tables, which is the current default.

docs/overview.md
```diff
@@ -0,0 +1,140 @@
...
+- pending-cancel: Will be cancelled at the end of the billing period
+
+Status changes trigger actions that update dates, send emails, etc.
+
+### 4. Customer-Initiated Actions
+
+Customers can:
+- Cancel subscriptions
+- Change payment methods
+- Update subscription items (if allowed)
```
barryhughes: Could specifically calling out the ability to switch between tiers add some extra clarity here?

docs/overview.md
vbelolapotkov: Good catch! I replaced it with the link to HPOS.

docs/overview.md
vbelolapotkov: Fair point, I've changed the wording to more generic. I haven't explored possible customer options to modify subscription, so probably it'll make sense to create a separate page for that or link to product page.

docs/init.md
vbelolapotkov: Thanks! I've check the docs for AS, and indeed it looks like WP_Cron is not always used. More on that, looks like we allow overriding the scheduler with `woocommerce_subscriptions_scheduler` hook. So we can't even be sure that AS is used. Lastly I haven't found any checks which would tell us for sure that we're running within scheduled action context. So it looks like everything is loaded when AS event is triggered.

Unless I'm wrong we also load both customer facing and admin stuff in all context. At least I don't see any checks before calling WC_Subscriptions_Admin::init(); or WC_Subscriptions_Cart::init();

docs/init.md
vbelolapotkov: Good point. I think a dedicated page on how actions are scheduled and how AS is used would be great

docs/init.md
vbelolapotkov: Fair enough, after exploring our init logic, I've decided to remove the section completely.

docs/init.md
vbelolapotkov: Good point, removing

docs/init.md
vbelolapotkov: Nope, cleaning up

docs/init.md
vbelolapotkov: You're right, in fact we load everything every time 😬

docs/init.md
vbelolapotkov: Weird, but there is a helper function `wcs_is_rest_api_request` which does this check if available. Anyway, that's not enough to be confident, so I replaced with reference to helper function only.

docs/init.md
vbelolapotkov: Right, IDK if it's worth highlighting that core doesn't have anything, but we don't have any checks either, so I just mentioned helper functions we have in place which could be used for identifying current context.

docs/init.md
vbelolapotkov: > This function is, afaik, used primarily to avoid setting things up within an ajax context.

And we don't use it :D At least I haven't found its usage.

docs/init.md
vbelolapotkov: Yep, edited.

docs/init.md
vbelolapotkov: Fair, removed that section completely, because we don't have such logic at all.

docs/init.md
vbelolapotkov: Yep, dropped AS element. However, it kind of sucks that we can't detect that and avoid loading bunch of unnecessary code like admin or store front stuff 😬

docs/init.md
barryhughes: > More on that, looks like we allow overriding the scheduler with woocommerce_subscriptions_scheduler hook.

Interesting ... I wasn't aware of that. TIL!

Actually, there's a parallel to WooCommerce here because, in both plugins, we have a mix of usages (in the sense that both sometimes interact *directly* with Action Scheduler, but on other occasions use an abstraction—`WCS_Action_Scheduler` in our case or `WC_Action_Queue` in WC Core's case).

> Unless I'm wrong we also load both customer facing and admin stuff in all context.

Yep, that's my read. Though, *some* amount of this work comes down to registering callbacks that will never be invoked, which isn't an uncommon pattern, and by itself doesn't have too much performance impact.

> Lastly I haven't found any checks which would tell us for sure that we're running within scheduled action context.

I seem to remember this being requested (as an enhancement for Action Scheduler itself), but I don't think it has been implemented yet. That said, though it wouldn't be very elegant, one *could* use actions like these to determine if we've entered or exited a queue processing context:

- `action_scheduler_before_process_queue`
- `action_scheduler_after_process_queue`

That said, I'm not actually sure how useful this would be, and the Action Scheduler model as I understand it means that a scheduled action is still just an action (albeit one that has been deferred to some future point in time).

docs/init.md
```diff
@@ -0,0 +1,81 @@
...
+4. Activation/deactivation procedures
+5. Action Scheduler batch size configuration
+6. Initialization of notification batch processor
+
+## Context-Specific Initialization
+
+Core plugin doesn't have custom logic for it and relies on WordPress and WooCommerce logic here.
+So some hooks might be skipped by WordPress or WooCommerce in various context (Admin, Store front end, REST_API or WP_Ajax).
+
+However, WooCommerce Subscriptions Core provides global helper functions which allow to execute code within certain context only:
```
barryhughes, line 67: Not sure if this is worth calling out—perhaps it's obvious—but these are also used to check that we are *not* inside a given context. For instance, we might check if we are inside an ajax request in order to bail out of an operation or to suppress output (to avoid 'contaminating' the response).

docs/subscriptions_orders.md
```diff
@@ -0,0 +1,401 @@
...
+**Characteristics**:
+- Standard `shop_order` type
+- Contains subscription products that triggered subscription creation
+- May contain both subscription and non-subscription products
+- Processed like a normal WooCommerce order
+
+**How to identify**:
+```php
+// Check if an order is a parent order
+$is_parent = wcs_order_contains_subscription($order);
```
barryhughes, line 136: Could be a little misleading. Might this be better?

```suggestion
$is_parent = wcs_order_contains_subscription( $order, 'parent' );
```

Otherwise `$is_parent` will be `true` even if the `$order` is a switch order.

docs/subscriptions_orders.md
```diff
@@ -0,0 +1,401 @@
...
+## Order Status Flow
+
+Orders and subscriptions have different status flows:
+
+### Subscription Status Flow
+
+- `pending` → `active` → `cancelled`/`expired`/`on-hold`
+- `on-hold` → `active` or `cancelled`
+- `cancelled` (terminal state)
+- `expired` (terminal state)
```
barryhughes, line 227: Should `pending-cancel` be factored in here?

docs/subscriptions_orders.md
```diff
@@ -0,0 +1,401 @@
...
+- `failed` → `processing` (after successful payment retry)
+- `processing` → `completed`
+- `completed` (terminal state)
+- `cancelled` (terminal state)
+
+## Order Generation Process
+
+### Renewal Order Creation
+
+1. Scheduled via Action Scheduler based on the `next_payment` date
```
barryhughes, line 241: *Might* be worth flagging that admins/merchants can also manually trigger this via the subscription editor?

docs/init.md
vbelolapotkov: Thanks for adding more context, it's really helpful as I don't much about how AS works. Looks like another somewhat messy/complicated area, so I bookmark it for now and will move on to other areas. We can always enhance that integration and docs later :)

vbelolapotkov:
Closing as not relevant anymore

